PR #2, opened 2021-06-01
Switch to lts-17.14

### LINKED ISSUE #3: Use newer version of GHC / stack lts
Not a very high priority task, but should attempt to move to a newer stackage resolver just to keep up...

plajjan:
Fixes #3


plajjan:
Spoke offline with @nordlander who blessed this. In essence, we trust Haskell enough that if something compiles, which it does, then it will also work. Thus merging.

PR #14, opened 2021-07-19
Fix gcc linking

plajjan:
The order in which arguments are provided matters. gcc/ld will look for
things to link based on the compiled source files, thus the sources
should come before the linking args. Putting the libs to link first
becomes like a NOP since we have no unresolved references at that time.

I'm not sure why, guessing it has to do with different implementations
of gcc/ld, but this was working on OS X already before but trying to use
actonc under Linux would result in a lot of "undefined reference to X",
which are now all fixed.

Since we need -lkqueue on Linux, but there is no such library on OS X,
we use preprocessing to make compilation of actonc conditional based on
the host OS.

Note how this is not the only thing that needs fixing for everything to work correctly on Linux.

### DIFF
--- a/compiler/ActonCompiler.hs
+++ b/compiler/ActonCompiler.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 -- Copyright (C) 2019-2021 Data Ductus AB
 --
 -- Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


plajjan:
@nordlander PTAL, in particular on this preprocessing stuff, which I'm not sure is a good idea or not.

nordlander:
Good work, it seems like clang has allowed us to be a bit sloppy regarding argument order. Using CPP is ok since it's practically a language extension rather than a preprocessor.

PR #15, opened 2021-07-19
Make Acton work on Linux

plajjan:
This avoids some multiple definition errors we see on Linux (Debian buster, gcc 10.2) and with it, the test suit passes.

This builds upon the fixes in #14. Using two PRs to let the work be reviewed independently.


nordlander:
Exactly right!

PR #27, opened 2021-07-22
actonc: Add --version

plajjan:
This adds a --version argument to actonc that prints the version
together with information about how it was compiled, like this:

    acton 0.3.0.20210722.231335

The version is set to the version in package.yaml and like above, the
date and time is appended when it is compiled. This is to denote that it
is a "local build". A release will have the simple version number, for
example:

    acton 0.3.0

We can tell the build system that we are building a release by defining
the environment variable BUILD_RELEASE.

I originally tried appending git information, for example:

    git describe --all --long --dirty
    heads/perf-0-gc254379-dirty

However, Cabal isn't very happy about all that extra stuff. Haskell
versions consists of dot-separated numbers. Whatever is not a number
is thought of as a "tag". There is some support for "tags", like the
Data.Version.showVersion function supports the concept. Cabal does not,
so in the end I had to stick to plain numbers and opted to just take the
build date and time rather than something related to git. I suppose it
would be possible to encode the git info as integers... but it feels
rather convoluted :P


plajjan:
This is already merged through #28 but this isn't marked as merged... hmm, I must have done some slight rewrite to some commit. Closing this. The code really is merged.

PR #28, opened 2021-07-22
actonc: detect C compilation errors

### LINKED ISSUE #19: Detect errors / return code in gcc compilation
actonc uses C as an intermediate representation. That means after we've done compiling an .act program to C we run gcc to compile C to a binary and then include than in a library file that we can finally link together to form a program.

Some Acton programs do not currently compile correctly. They are valid Acton syntax, so they will pass the first compilation step but gcc will barf and return some error. However, actonc doesn't honor that return code and continues anyway. 

Here's an example application that fails in that way:
```
actor Thing():
    def foo():
        return 1

actor main(env):
    var thingies = [Thing()]

    for thing in thingies:
        bar = thing.foo()
        if 1 > 0:
            break

    env.exit(0)
```

```
kll@ThinkYoga:~/terastream/acton/test$ ../actonc --root main for-break.act
/home/kll/terastream/acton/modules/test/for-break.c: In function ‘test$_$for_45_break$$l$6c$5cont’:
/home/kll/terastream/acton/modules/test/for-break.c:190:9: error: break statement not within loop or switch
  190 |         break;
      |         ^~~~~
/usr/bin/ld: /tmp/cc5uOMTi.o: in function `$ROOTINIT':
/home/kll/terastream/acton/modules/test/for-break.root.c:3: undefined reference to `test$_$for_45_break$$__init__'
/usr/bin/ld: /tmp/cc5uOMTi.o: in function `$ROOT':
/home/kll/terastream/acton/modules/test/for-break.root.c:6: undefined reference to `test$_$for_45_break$$main$new'
collect2: error: ld returned 1 exit status
kll@ThinkYoga:~/terastream/acton/test$ echo $?
0
kll@ThinkYoga:~/terastream/acton/test$ 
```

The resulting behavior can give quite subtle errors. When developing acton programs and iterating on the code, I can first have a working program, let's say the above program first looked like this:
```
actor Thing():
    def foo():
        return 1

actor main(env):
    var thingies = [Thing()]

    for thing in thingies:
        bar = thing.foo()

    print("Hello")
    env.exit(0)
```
(i.e. no break statement nested under a condition)

This works correctly:
```
kll@ThinkYoga:~/terastream/acton/test$ ../actonc --root main foo.act && ./foo
Hello
```

Now let's add back the break:
```
actor Thing():
    def foo():
        return 1

actor main(env):
    var thingies = [Thing()]

    for thing in thingies:
        bar = thing.foo()

    print("Hello")
    env.exit(0)
```

recompiling & running yields:
```
kll@ThinkYoga:~/terastream/acton/test$ ../actonc --root main foo.act && ./foo
/home/kll/terastream/acton/modules/test/foo.c: In function ‘test$foo$$l$5c$7cont’:
/home/kll/terastream/acton/modules/test/foo.c:127:9: error: break statement not within loop or switch
  127 |         break;
      |         ^~~~~
Hello
kll@ThinkYoga:~/terastream/acton/test$
```
that is, the compilation fail... and correctly does not output the foo binary nor the library in libActon but as we get return code 0 we happily continue to run.

plajjan:
We now inspect the return code of the subprocesses we spawn (gcc & ar)
to compile the generated C code. If there's a failure we report it as an
internal compiler error.

Fixes #19


plajjan:
I am going to go ahead and merge this. @nordlander I still think it would be good if you reviewed this and if something looks completely wacko, we can fix that in a later commit. I want to merge it because I have other pieces related to how we run tests that I want to merge and it depends on this.

PR #46, opened 2021-07-27
rts: Add runtime args for DDB

plajjan:
It is now possible to configure the RTS to use the distributed database
backend for storing the run time state of actors. It was previously
enabled through a build time flag and the database host and port was
hard-coded. There are two run time options:

--rts-ddb-host, the hostname to connect to
--rts-ddb-port, the TCP port to connect to, default is 32000

A third option, --rts-verbose, was also added to make the RTS a bit more
chatty about it is doing, in particular in regards to the backend.

An example, examples/count_forever, is added that shows how run time state persistence and restoration works.

I have been contemplating whether I should modify argc & argv before I pass it into the Acton application. My initial thought was no, let's not. Let's simply "reserve" that arguments starting with `--rts-` are for the run time system but still pass them on. This makes it possible for an Acton program to see how the RTS is started. But then I was thinking that we probably want to allow RTS introspection from an Acton program by `import acton.rts` and then something something. We don't have that though, so I figured passing args is fine... and as I was modifying I couldn't really make up my mind. Right now I feel I'm leaning towards stripping the rts args and passing modified versions of argc & argv to the Acton program but we don't have to do that in this PR either. This is already a marked improvement on things. Rome wasn't built in a day.

Do we need a test case? We already test that programs work without the RTS. I suppose we could write some simple program that works with the RTS. Again though, that could come in a different PR too.

### DIFF
--- a/rts/rts.c
+++ b/rts/rts.c
@@ -641,7 +641,7 @@ void BOOTSTRAP(int argc, char *argv[]) {
     clock_gettime(CLOCK_MONOTONIC, &now);
     $Msg m = $NEW($Msg, ancestor0, &$NewRoot$cont, now.tv_sec, &$WriteRoot$cont);
 
-    if (db != NULL) {
+    if (db) {
         create_db_queue(env_actor->$globkey);
         create_db_queue(ancestor0->$globkey);
         int ret = remote_enqueue_in_txn(($WORD*)&m->$globkey, 1, NULL, 0, MSG_QUEUE, (WORD)ancestor0->$globkey, NULL, db);
@@ -758,7 +758,7 @@ void FLUSH_outgoing($Actor self, uuid_t *txnid) {
             ENQ_timed(m);
             dest = 0;
         }
-        if (db != NULL) {
+        if (db) {
             int ret = remote_enqueue_in_txn(($WORD*)&m->$globkey, 1, NULL, 0, MSG_QUEUE, (WORD)dest, txnid, db);
             //if (dest)
             //    printf("   # enqueue msg %ld to queue %ld returns %d\n", m->$globkey, dest, ret);
@@ -1058,7 +1058,7 @@ void serialize_actor($Actor a, uuid_t *txnid) {
 void FLUSH_offspring($Actor current, uuid_t *txnid) {
     $Actor a = current->$offspring;
     while (a) {
-        if (db != NULL) {
+        if (db) {
             create_db_queue(a->$globkey);
             a->$consume_hd = 0;
             serialize_actor(a, txnid);
@@ -1087,7 +1087,7 @@ void *main_loop(void *arg) {
             $R r = cont->$class->__call__(cont, val);
             switch (r.tag) {
                 case $RDONE: {
-                    if (db != NULL) {
+                    if (db) {
                         uuid_t * txnid = remote_new_txn(db);
                         current->$consume_hd++;
                         serialize_actor(current, txnid);
@@ -1138,7 +1138,7 @@ void *main_loop(void *arg) {
                     break;
                 }
                 case $RWAIT: {
-                    if (db != NULL) {
+                    if (db) {
                         uuid_t * txnid = remote_new_txn(db);
                         serialize_actor(current, txnid);
                         FLUSH_outgoing(current, txnid);
@@ -1169,7 +1169,7 @@ void *main_loop(void *arg) {
                 if (ENQ_msg(m, m->$to)) {
                     ENQ_ready(m->$to);
                 }
-                if (db != NULL) {
+                if (db) {
                     uuid_t *txnid = remote_new_txn(db);
                     timer_consume_hd++;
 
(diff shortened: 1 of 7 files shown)


plajjan:
Hmm, I noticed something else... using getopt (in C, in the RTS) will modify and rearrange argv. At least one man page also backs that up: https://docs.oracle.com/cd/E36784_01/html/E36874/getopt-long-only-3c.html

On the main branch:
```
kll@Boxy:~/kod/acton/examples$ ./args --abc foo bar --foo
[./args, --abc, foo, bar, --foo]
```

On this branch:
```
kll@Boxy:~/kod/acton/examples$ ./args --abc foo bar --foo
[./args, --abc, --foo, foo, bar]
```

Note how `--foo`, which is specified last, is moved to the second argument position.

I can live with extra arguments being passed through, but rearranging the args feels a bit much, no?

plajjan:
Discussed this during yesterdays meeting. All agreed we want this. Some extra discussion on how to pass in various bits of config, but nothing that contradicts this PR.

@nordlander wrote the original code integrating the RTS with the database so he will review my code changes. Assigning him as reviewer!

Agreed with @nordlander and @aagapi that we will strip the rts arguments so they are not visible from within the Acton program. Thus, we need to create a new argc and argv that we pass to the root actor when bootstrapping and these new argc/argv should not contain the `--rts-ddb-host` etc arguments.

plajjan:
I have (force) pushed some updates. This PR is now rebased on the `main` branch to bring in some new stuff from there. I have addressed both the argument parsing concerns, i.e.:
- now telling getopt_long to return things in order rather than permute argv
- consuming / stripping the RTS arguments before passing on arguments to the Acton program

I accomplish this by creating a copy of argv in new_argv to which we only copy the arguments not consumed by the RTS. 

A small test case is added for the argument parsing.

No resources are freed, not sure if there is any point in trying to do that or not since I think we need new_argv throughout the entire program!?

Marking this PR as ready for review! @nordlander feel free to have a read and please be kind ;)

rts/rts.c
```diff
@@ -1225,7 +1221,77 @@ void $register_rts () {
...
+    // stop scanning once we've seen '--', passing the rest verbatim
+    int opt_scan = 1;
+    for (int i = 0; i < argc; ++i) {
+        if (strcmp(argv[i], "--") == 0) opt_scan = 0;
+        if (opt_scan) {
+            for (int j = 0; j < lo_len; j++) {
+                // compare at +2 since in argv it is --foo while only foo in long_options
+                if (strcmp(argv[i]+2, long_options[j].name) == 0) {
+                    if (long_options[j].has_arg == 1) i++;
+                    goto cnt; // abort inner loop
```
nordlander, line 1283: Better expressed as
```suggestion
                    continue; // abort inner loop
```

rts/rts.c
```diff
@@ -1225,7 +1221,77 @@ void $register_rts () {
...
+                }
+            }
+        }
+        size_t length = strlen(argv[i])+1;
+        new_argv[new_argc_dst] = malloc(length);
+        memcpy(new_argv[new_argc_dst], argv[i], length);
+        new_argc_dst += 1;
+        cnt:;
```
nordlander, line 1291: Label not needed if continue is used.

rts/rts.c
```diff
@@ -1225,7 +1221,77 @@ void $register_rts () {
...
+                if (strcmp(argv[i]+2, long_options[j].name) == 0) {
+                    if (long_options[j].has_arg == 1) i++;
+                    goto cnt; // abort inner loop
+                }
+            }
+        }
+        size_t length = strlen(argv[i])+1;
+        new_argv[new_argc_dst] = malloc(length);
+        memcpy(new_argv[new_argc_dst], argv[i], length);
+        new_argc_dst += 1;
```
nordlander, line 1290: Because there is no mutation of the strings pointed to by the argv elements, it's safe to simply reuse these pointers in new_argv. 
```suggestion
        new_argv[new_argc_dst++] = argv[i];
```

nordlander (approved):
Spot on, but with room for some C code simplifications!

nordlander:
I "approved" the patch with some comments, but maybe that just made the process unnecessarily complicated...

-- Johan

rts/rts.c
plajjan: Duh, that makes sense.

rts/rts.c
plajjan: I think you are actually fooled by my, admittedly bad, comment. I want a continue on the outer loop, so as to skip the copying etc that comes just after the inner loop. `continue` on the inner loop would still do the copying! I could of course do it with another temporary variable but not sure if that is better. I have improved the comment though!

plajjan:
I've incorporated @nordlander suggestion on reuse of argv values and improved the misleading comment. Also changed test case ever so slightly to ensure coverage. Merging.

PR #49, opened 2021-07-28
Link Acton libs in right order

plajjan:
We got to link with libActonProject first, since this is where our
project specific things are, which can reference things (builtins /
stdlib) in libActon. This means -lActonProject goes before -lActon!


plajjan:
This was BTW not a problem earlier since libActonProject didn't exist, it was my split of the libActon library into the system part (still libActon) and a users project part (libActonProject) that caused this problem.. or rather, my change was incorrect since it should have linked in the right order. This fixes it.

PR #50, opened 2021-07-28
Add simple time module

### LINKED ISSUE #18: Add time function
I need a time function of some sort in order to build benchmarks for Acton. Sure, I can measure externally from Acton but it would be so much better with a timing function. We don't need to implement a complete `datetime` style module. I'm talking about like simple UNIX timestamps - as long as they have enough precision!

```
kll@ThinkYoga:~$ python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import time
>>> time.time()
1626764135.0713706
>>> time.time_ns()
1626764135940527153
>>> 

```

so maybe build a `time` module but only implement the `time()` or `time_ns()` function!?

I don't think there are any special Acton concerns here, like talking to an external system. If an actor is migrated between two different compute nodes, then calling `time.time()` might give surprising results in case the machines time aren't synced up. Like the clock going backwards... but I think that can already happen, right? No guarantee of monotonic time.

plajjan:
This adds a time module to Acton. It exposes a single function, time(x),
which returns the current time as an integer with nanosecond precision.

time(x) currently takes an argument, though its value is irrelevant.
It's a dummy. I tried to remove the argument but  that breaks the build
and I don't get why. Halp.

I'm not sure how we should design our time module. Like we can model it
directly on Python's time module or we can do something else...
something better!? It's low level enough that I'm not sure it matters
making it more Pythonic / Actonic.

There is a simple test of time.time(x).

Closes #18 

## TODO
- [ ] remove argument from `time.time()`
- [ ] sanity check the code by @nordlander or @sydow

### DIFF
--- a/modules/Makefile
+++ b/modules/Makefile
@@ -7,6 +7,9 @@ __builtin__.ty: __builtin__.act
 math.ty: math.act __builtin__.ty
 	$(ACTONC) --stub math.act
 
+time.ty: time.act __builtin__.ty
+	$(ACTONC) --stub time.act
+
 numpy.ty: numpy.act math.ty
 	$(ACTONC) --stub numpy.act
 
--- a/modules/time.act
+++ b/modules/time.act
@@ -1,0 +1,36 @@
+# Copyright (C) 2019-2021 Data Ductus AB
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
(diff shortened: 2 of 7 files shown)


plajjan:
@nordlander I need your help in understanding why removing the argument from time.time(x) doesn't work... Either I'm missing something or there's a bug somewhere.

plajjan:
I've now re-read my code. Maybe it's really reading it for the first time since the first time I blindly charged through trying to implement the functionality I needed. I now see that I have a completely unecessary float class. It makes sense for math, from which I copied the whole shebang, but not for my time module. So this needs to be cleaned up...

plajjan:
Spoke to @nordlander about it. The more I look at the code and the more I listen to Johan the more I realize how wrong this way. Closing. Will try to sort this out and push another attempt.

PR #58, opened 2021-08-02
Add continuous releases

### LINKED ISSUE #36: Automatically create release from changelog.md
Check out https://github.com/marketplace/actions/create-github-releases-based-on-changelog

plajjan:
This will create and continuously update a GitHub release called
"latest" that holds the artifacts produced in the latest run CI test
job on the main branch. The text body of the release will be updated to
reflect the changes in CHANGELOG.md.

Closes #36 

Cloees #4

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,3 +62,36 @@ jobs:
           path: ${{ github.workspace }}/acton-linux-x86_64*
       - name: Run tests
         run: make -C ${{ github.workspace }} test
+
+  release-latest:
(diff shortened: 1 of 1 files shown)


plajjan:
While this isn't perfect, nor have I even tested with using tagged releases, like we will do for real versions, but it does solve the super immediate need and so I'll merge. It might be that I'll have to rework quite a bit to make it work for the normal releases too... let's see, but for now, we get "latestlies", like nightlies but for every build :)

PR #82, opened 2021-08-14
Make --hgen/--cgen output clean code

plajjan:
Rather than include a mostly useless header we just print the code which
makes it easier to pipe the result into a file. The CC command is also
moved to a separate option called --ccmd. I've found multiple times that
I just wanted the cc command and nothing else, which is now possible!

The downside is perhaps that previously one could do `actonc --hgen
--cgen`, get all the output and it would still be clear what parts were
what. If that functionality is important we could reimplement printing
of headers through the --verbose option or similar. Maybe we could
detect the output type, if it is a terminal or if we are piped to a file
and do output accordingly.

### DIFF
--- a/compiler/ActonCompiler.hs
+++ b/compiler/ActonCompiler.hs
@@ -61,6 +61,7 @@ data Args       = Args {
                     llift   :: Bool,
                     hgen    :: Bool,
                     cgen    :: Bool,
+                    ccmd    :: Bool,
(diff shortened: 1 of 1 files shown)


plajjan:
I doubt this is very controversial so going to merge. @nordlander please notice the change in behavior here though and if you feel strongly any other way we can change back or add some other option.

PR #83, opened 2021-08-14
Add time module with time() function

### LINKED ISSUE #18: Add time function

plajjan:
This adds a time module with a single function called time() that will
return the current time with nanosecond precision.

There's a simple test case to ensure it does what it should.

Closes #18.


plajjan:
Attempt #2. This feels a lot better than my previous attempt (#50).

@nordlander, you're up, plxz review :smiley_cat:

nordlander (approved):
No problems here. The concern over how to handle fatal C errors is of course very valid in general, but in case of clock_gettime I think the only possible errors would be caused by corrupt syscall arguments (which can never happen with the current code). So I'd even argue it's ok to ignore the return code altogether.

PR #84, opened 2021-08-14
Add doc to explain C library wrapping in Acton

plajjan:
This is a short and simple guide to wrapping C libraries in Acton that I
wrote while implementing the time module (see #83).

This sort of depends on #82 as it assumes --hgen and --cgen will output clean code without any headers or the C compilation commands.

### DIFF
--- a/docs/wrapping_c_libraries.md
+++ b/docs/wrapping_c_libraries.md
@@ -1,0 +1,118 @@
+# Wrapping C libraries in Acton
+
+This document attempts to provide a brief introduction to integration C libraries in Acton.
+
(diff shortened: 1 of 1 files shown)


plajjan:
@nordlander you wanna have a read before we merge? Anything to add? IMHO not super important that this is perfect - better to merge something than nothing at all.

nordlander (approved):

nordlander:
Good write-up, although there's of course much more to say. But I suspect that the more we write on the topic the more we'll realize that end users will really need better tool support to make C integration feasible. We'll have to approach this step by step, I guess.

plajjan:
Yes, there's so much more to say. Baby steps.

Since you approved I think you simply pressed the wrong button closing this instead of merging. So I'll reopen and merge.

nordlander:
Wrong button, indeed! Thanks for resolving.

PR #88, opened 2021-08-16
Separate build output from sources

plajjan:
This splits up the source of Acton from the built output, which goes in
the dist directory. Baking a release is then a simple operation of
taring up the dist directory.

When we invoke actonc, like for building tests, we do that in the dist
directory, thus producing output in that directory rather than mixed up
with the source code of Acton itself.

The release will be slightly slimmer, skipping certain .o files, .c
files etc that were previously included.


plajjan:
I'm going to merge this. Heads up @nordlander and @sydow, actonc has moved from the root dir into dist!

PR #93, opened 2021-08-17
Change --path to --syspath

### LINKED ISSUE #90: Rename --path to --syspath
I just think it's clearer... and better.

plajjan:
I think this just makes it clearer what it actually is. It is the path
to where the Acton system lives, syspath! The internal variable is
called syspath so this also aligns with that.

Closes #90.


plajjan:
@nordlander FYI, changing name of this...

Issue #86, opened 2021-08-14
Intermittent failures of queue_unit_tests!?


plajjan:
I don't know how to see old runs of a GitHub Action for a PR but for #82 I had two failures of a database related test and IIRC it was queue_unit_tests. We should try to reproduce somehow... I had to manually re-run the workflow twice to get it working, so I know something was failing. It was just failing under the Linux job and not on OS X but I'm not sure that has any real bearing, could just be intermittent!?

On the other hand, if we can't reproduce, then not sure what to do. So give this a shot I guess or give up and close issue.

plajjan:
Ran this locally:
```
GOOD=0; TOT=0; for I in `seq 1000`; do TOT=$(($TOT+1)); ./queue_unit_tests && GOOD=$(($GOOD+1)); echo "Total: $TOT"; echo "Good: $GOOD"; done
```
and stopped it just now with this output
```
Total: 277
Good: 267
```
i.e. there have been 10 failures. So yes, this does seem to fail intermittently!

plajjan:
Here's a failed run:
```
BACKEND: Queue table 0 created
Test create_queue_table - OK (0)
BACKEND: Queue 0/0 created
Test create_queue - OK (0)
Test create_producer_thread - OK (0)
Test create_consumer_thread - OK (0)
BACKEND: Inserted queue entry 0 in queue 0/0, status=0
Test create_consumer_replay_thread - OK (0)
BACKEND: Subscriber 0/0/0 subscribed queue 0/0 with callback 0x7f4a0b62ce10
Test subscribe_queue - OK (0)
BACKEND: Subscriber 0 read 1 queue entries, new_read_head=0
CONSUMER 0: successful_dequeues=1, last_entry_id=0
BACKEND: Subscriber 0 consumed entries, new_consume_head=0, read_head=0
CONSUMER 0: successful_dequeues=1, successful_consumes=1
BACKEND: Subscriber 0/0/1 subscribed queue 0/0 with callback 0x7f4a0ae2be00
BACKEND: Subscriber 1 read 1 queue entries, new_read_head=0
CONSUMER 1: successful_dequeues=1, last_entry_id=0
BACKEND: Inserted queue entry 1 in queue 0/0, status=0
BACKEND: Attempting to notify local subscriber 0 (0x7f4a0b62ce10/0x7f4a0b62ce30/0x7f4a0b62ce60/0x55a5b417a530)
Consumer 0/0/0 received notification for queue 0/0, status 0
[...]

PR #96, opened 2021-08-20
Disable the queue_unit_tests

plajjan:
It fails intermittently so until we've fixed that, I'm disabling the test.

Related to #86


plajjan:
FYI @aagapi

PR #124, opened 2021-08-24
Johan fixes

plajjan:


plajjan:
@nordlander FYI, these are the commits from the develop branch but without the merge commit of develop unto itself.

PR #125, opened 2021-08-24
Add regression test for async actor method call

plajjan:
When calling an actor method like we are here, it should be run async,
but we are instead seeing it is run synchronously.


plajjan:
@nordlander FYI, here is a smaller regression test of the actor method call async discrepancy I showed you.

Issue #110, opened 2021-08-23
Type inferencing cannot solve union of (None, int) while (None, str) works


plajjan:
### Code / what did you do

I tried this code:

```Acton
def foo():
    if 1 == 3:
        return None
    return 3

actor main(env):
    await async env.exit(0)
```

### Expected outcome
It should compile! The type checker should understand that the return value is a union of (None, int) or maybe that's a "maybe int".

### Actual outcome

Trying to compile, I get:

```
../../dist/actonc union-none-int.act --root main

******************** Type error
Cannot solve 'w$18' : 'T$4' (__builtin__.Number), None < 'T$4', 'T$4' < 'T$3'
make: *** [Makefile:22: union-none-int] Error 1
```

If I change the program to return a string instead, it will work, like so (note the quotes around `3` in the return):
```
def foo():
    if 1 == 3:
        return None
    return "3"

actor main(env):
    await async env.exit(0)
```

### Meta
This is on the currently latest version of the main branch
`actonc --version`:
```
acton 0.4.2.20210820.11.49.39
```

nordlander:
What a coincidence! I found this bug myself earlier this morning :-)

And the fix is traightforward, coming up...

plajjan:
This is fixed through a recent commit. Closing.

PR #126, opened 2021-08-24
Add two regression tests

plajjan:


plajjan:
@nordlander FYI, I wrote a regression test for #110 but forgot to push it yesterday. We can see that your fixes indeed shows up as this negative test was first failing. I had to mark the test as a positive one to make it pass! All good!

nordlander:
Great! I was expecting a negative test to fail after my fix but didn't couldn't find it. This explains why!

PR #132, opened 2021-08-25
Fix passing actor method as argument

### LINKED ISSUE #109: Calling actor method passed as argument from other actor segfaults
### Code / what did you do

I tried this code:

```Acton
actor Foo():
    def func(a):
        print("request", a)

actor Bar(f):
    f("test")

actor main(env):
    var f = Foo()
    var b = Bar(f.func)
```

### Expected outcome

I expected it to work, seeing "request test" in my terminal.
```
request test
```

### Actual outcome

I got a segfault instead

```
Segmentation fault
```

I can instead pass the actor and access the method more explicitly, which then works:
```
actor Foo():
    def func(a):
        print("request", a)

actor Bar(f):
    f.func("test")

actor main(env):
    var f = Foo()
    var b = Bar(f)
```

### Meta
This is on the main branch
`actonc --version`:
```
acton 0.4.2.20210820.11.49.39
```

plajjan:
Closes #109.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Fixed
 - Distributed RTS mode has been fixed with regards to actor bootstrap that
   previously lead to duplicat actors
+- An actor method (callback) can now be passed as an argument to another actor
+  method
+  - This previously lead to a segmentation fault but has now been fixed.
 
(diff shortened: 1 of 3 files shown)


plajjan:
@nordlander can you just check that the entry in the changelog looks ok?

PR #134, opened 2021-08-25
Add git workflow cheat sheet for develop branch

plajjan:


plajjan:
@nordlander ok, I have now added a second document explaining the feature branch workflow. Do you wanna have a read? Feel free to merge if it looks good... or return with feedback :)

nordlander:
Looks very good, with two exceptions: line 40 starts with a confusing "Instead", and lines 42-43 should probably just be removed now that we're talking about feature branches.

plajjan:
@nordlander indeed, I forgot remove as I wrote that "see below" and then immediately started writing the subsequent chapter. I fixed that now. Merging.

PR #138, opened 2021-08-25
Treat warnings as errors

plajjan:
There are currently some warnings being produced when compiling the C
code generated by actonc. We believe this likely indicates bugs and
thus, by treating these warnings as errors we are able to detect such
situations earlier. There are situations where the warnings are
irrelevant, but we should probably explicitly disable those warnings in
that case.

### DIFF
--- a/compiler/ActonCompiler.hs
+++ b/compiler/ActonCompiler.hs
@@ -66,6 +66,7 @@ data Args       = Args {
                     version   :: Bool,
                     stub      :: Bool,
                     rts_debug :: Bool,
+                    cpedantic :: Bool,
                     syspath   :: String,
                     root      :: String,
                     file      :: String
@@ -88,6 +89,7 @@ getArgs         = Args
                     <*> switch (long "version" <> help "Show version information")
                     <*> switch (long "stub"    <> help "Stub (.ty) file generation only")
                     <*> switch (long "rts-debug"<> help "Include RTS debug support in output program")
+                    <*> switch (long "cpedantic"<> help "Pedantic C compilation with -Werror")
                     <*> strOption (long "syspath" <> metavar "TARGETDIR" <> value "" <> showDefault)
                     <*> strOption (long "root" <> value "" <> showDefault)
                     <*> argument str (metavar "FILE")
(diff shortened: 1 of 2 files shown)


plajjan:
First of all, before we can merge this, the current warnings need to be fixed. I think there is just one - around how exceptions are raised. @nordlander  should be aware of it and working on a fix I believe.

But then, I'm actually a little torn on this one. I want to maintain a 0-warning policy. `-Werror` effectively creates a dependency on a specific version of the C compiler. Like, our code can work just fine on the current gcc 10.2 but with 10.3 a new warning exists and as we use `-Werror`, that is now a fatal error. We turn all of our users into guinea pigs - thee who upgrades to a new cc first will hit errors. That's not good! It forces our hand, in that we have to immediately look at the warning and address it, either by changing our code or by ignoring the warning.

I want to be able to choose when we deal with something. Thus, I think this needs to be an option, like `actonc --strict` or something. It can either be a opt-in or an opt-out. With opt-in, we can enable this option in our own CI, so we will find problems there and can maintain our 0-warning policy, but it will still work regardless for users who might be on other gcc versions. With opt-out, we will effectively use all of our users (about 0 right now, likely) as a test bed for finding new compiler warnings. With an opt-out option, we at least give them a way of working around it... but I'm not sure if that's good. Shouldn't users NOT have to think about something like that? That Acton produces C code should be internals and hidden, no?

I'm leaning towards opt-in and it being something we just run in CI. @nordlander thoughts?

nordlander:
Those are good arguments, I subscribe in full. The important fact is that actonc isn't a black box, it's a tool that must work in tandem with a C compiler that the end user controls. Of course one can imagine CC versions that misinterpret or even reject the code we generate, but if we allow for warnings we have at least created some margin for the user. At the same time we really want to be aware any C compiler warnings duringour own internal testing, since it's our intention that the code we generated is indeed fully valid C.

So an opt-in flag seems like the right solution!

nordlander:
LGTM, but we should change the flag name to c-pedantic at some point.

plajjan:
Rebased to resolve conflicts and renamed argument to `--cpendatic`! I'm enabling auto-merge on CI success.

plajjan:
I had to disable `--cpendatic` in the regression suite to make the CI tests pass. I've created #207 as a follow-up to enable it again after fixing the warnings!

PR #163, opened 2021-08-30
143 generalization fix

### LINKED ISSUE #143: Unexpected forward ref failure
### Code / what did you do

I tried this code:

```Acton
actor Foo():
    def func(a):
        print("request", a)

actor Bar(f):
    f("test")

actor main(env):
    f = Foo()
    b = Bar(f.func)
    await async env.exit(0)
```

### Expected outcome

Same as when compiling test/regression/109-passing-method-argument.act, which only differs by having `f` and `b` in actor `main` declared as state variables (with keyword `var`).

### Actual outcome

Instead, a compilation error occurs, that indicates some problem with forward references that is not present when `f` and `b` are state variables.

```
******************** Compilation error
 9:9-11
  |
9 |    f = Foo()
  |        ^^^
Name Foo is reserved but not yet defined
```

### Meta

`actonc --version`:
```
 0.4.2.20210824.12.22.11
```

nordlander:
Fixed #143

### DIFF
--- a/builtin/minienv.c
+++ b/builtin/minienv.c
@@ -1022,7 +1022,7 @@ void minienv$$__init__ () {
 }
 
 
-void $eventloop() {
+void *$eventloop(void *arg) {
   while(1) {
     struct kevent timer;
     pthread_setspecific(self_key, NULL);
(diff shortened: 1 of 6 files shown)


plajjan:
@nordlander FYI, the CI test times out. It is stuck on `./105-type-inference-order`. This is reproducible locally - it just looks like it hangs forever.

nordlander:
Yup, although it's actually intermittent when I test it locally. Found the error in the, which is independent of the fix (the test code just happens to be sensitive to the rts bug). I'll fix that in another branch, disabling the test meanwhile.

PR #168, opened 2021-08-30
Fix divmod return value

plajjan:
@nordlander PTAL

### DIFF
--- a/builtin/int.c
+++ b/builtin/int.c
@@ -198,10 +198,7 @@ $int $Integral$int$__index__($Integral$int wit, $int n) {
 $tuple $Integral$int$__divmod__($Integral$int wit, $int a, $int b) {
   int n = a->val;
   int d = b->val;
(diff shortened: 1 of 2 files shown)


plajjan:
@nordlander can the tuple thingy be written in such a way that we can't make mistakes like this?

plajjan:
@nordlander I've updated this, I believe in accordance with our conversation. PTAL :)

nordlander (approved):
Your change is quite correct but... mother of all horros, the pattern you've followed is still buggy and exists in numerous places in the builtin code!

The correct last line must be
```
   return $NEWTUPLE(2, to$int(n/d), to$int(n%d));
```
otherwise the memory allocated won't take the number of tuple elements into account!

So this is larger than the original divmod issue. I suggest you merge this fix as it is and I'll open another issue for the missing use of $NEWTUPLE at various places (including compiler ouput).

Issue #176, opened 2021-08-30
Cannot unify __builtin__.atom and __builtin__.int


plajjan:
### Code / what did you do

I tried this code:

```Acton
import random

actor Foo():
    def bar(a):
        rest = int(a["id"]) % 2

actor main(env):
    a = {"id": random.randint(0,10)}
    f = Foo()
    f.bar(a)
    await async env.exit(0)
```

### Expected outcome

I expected to compile it without problem

### Actual outcome

Instead I get a compilation error:
```
******************** Type error
Cannot unify __builtin__.atom and __builtin__.int
```

Not sure if my program is the absolutely minimal reproduction and there might be somewhat similar issues, like if the dict a["id"] = 3 instead I get a different error but that still seems related / sort of the same form:
```
******************** Type error
Cannot infer 'w$41' : __builtin__.atom (__builtin__.Number)
```

### Meta
`actonc --version`:
```
acton 0.5.2.20210827.16.46.48
```

plajjan:
Fixed through #183. Closing.

Issue #170, opened 2021-08-30
Intermittent non-termination


nordlander:
### Code / what did you do

Run regression test 105-type-inference-order repeatedly.

### Expected outcome

Successful output and termination each time.

### Actual outcome

Intermittent termination failures after output.

### Meta

acton 0.5.2.20210830.9.43.18

nordlander:
Investigation shows that while an actor is being constructed, messages it sends to other actors risk being lost. In test 105 actor `main` emits tow messages while it initializes: one message to `f.func` (indirectly via the `Bar` constructor), and one to `env.exit`. Some runs show that the `env` actor can be mysteriously dropped from the ready queue before any worker thread has picked it up, apparently correlated to the order the receiving actors end up in the ready queue when `main` does its `await`.

Further investigation reveals the error to be a disastrous reuse of the `next` link in the run-time representation of actors. The link helps form the linked list of ready actors waiting in the ready queue, as well as the list of actors that may wait for a particular message result (very rarely more than one). This reuse is ok because an actor can never be ready and waiting at the same time. But the `next` link is also used to form the list of actors created by a particular actor since the last checkpoint (called the `offspring` list). This is disastrous because an actor can very well be made ready to run while still being in the `offspring` list of its parent!

In test 105 this is exsctly what happens to actor instance `f`. It's still and offspring to `main` when `main` does its `await` (which forces a checkpoint), yet it's also made ready to execute by having a message sent to it. Now, if `f` happens to be linked before the `env` actor in the ready queue when `main` empties its offspring list, the `next` link of `f` will be cleared and thus `env` will effectively be dropped from the ready queue. As a result, the `env.exit` message will never get executed and the program hangs.

nordlander:
Note: this is a fundamental bug in the rts that will affect more programs than test 105 on a rainy day. I'm concluding that the most robust fix will be to remove the need to keep track of offsprings altogether. The formal semantics has no such concept; here all actor initializations are just ordinary synchronous messages. Time to make the implementation follow suit!

PR #182, opened 2021-08-31
174 checker fixes

### LINKED ISSUE #174: Variable not in scope in function
### Code / what did you do

I tried this code:

```Acton
def foo(a):
    print(a)

actor main(env):
    b = 1
    foo(b)
    await async env.exit(0)
```

### Expected outcome

I expected it to print '1'

```
1
```

### Actual outcome

Instead, I get a compiler error.

```
******************** Compilation error
Name 'w$25' is not in scope
```

I can fix this by explicitly casting the integer to a string, like so:
```
def foo(a):
    print(str(a))

actor main(env):
    b = 1
    foo(b)
    await async env.exit(0)
```
but the error message is completely misleading and can be increasingly hard to find in a larger program.

### Meta
`actonc --version`:
```
acton 0.5.2.20210827.16.46.48
```

nordlander:
Fixes #174 and #176

### DIFF
--- a/compiler/Acton/Env.hs
+++ b/compiler/Acton/Env.hs
@@ -864,31 +864,31 @@ castable env t1 t2                          = False
 -- GLB
 ----------------------------------------------------------------------------------------------------------------------
 
-glb env (TWild _) t2                    = t2
-glb env t1 (TWild _)                    = t1
+glb env (TWild _) t2                    = pure t2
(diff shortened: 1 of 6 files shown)


nordlander:
It's bug #170 again, causing intermittent hanging in test 176...

nordlander:
N.B.: hanging is unrelated to the fixes in this pull request.

PR #189, opened 2021-09-04
Add RTS debug mode

### LINKED ISSUE #172: Introduce debug / step mode in the RTS
I was looking at the RTS the other day and wanted to understand it a little better. I got the idea to introduce lots more verbose debug logging, like every time we pick up an actor from the ready queue, every time we persist state etc etc. This would allow "stepping through" the execution of the RTS.... well, I'm not suggesting to implement pause, just that we can see in detail what it is doing.

I suspect that could severely affect performance, even if we condition all that logging behind a `if (debug) log...` or so. The only way to be sure would be to have a compile time flag for it.... and then it struck me, that we could build an rts lib and a rts-debug lib, both with the same symbols, just compiled with and without. Thus, the acton system comes shipped with both and actonc can choose, when it is compiling an Acton program, to link with one or the other.

plajjan:
This enables various debug printing while the RTS is running. It is very
detailed and will most definitely have a major performance impact. Thus,
it can only be used for debugging purposes in controlled environments.
Debug output is enabled with the --rts-debug option.

To completely eliminate any overhead in normal operations the debug
prints are not just conditioned on the runtime --rts-debug flag but are
also conditioned behind a build time flag. We build the rts lib twice,
once with the flag enabled and once without. actonc chooses which one to
link with when compiling the .act file. That is, given the program
helloworld.act, to run with debug output, we first have to compile with:

  $ actonc --rts-debug --root main helloworld.act

Then also run the program itself with:

  $ ./helloworld --rts-debug

If helloworld was not compiled with --rts-debug, we will get an error:

  $ ./helloworld --rts-debug
  ERROR: RTS debug not supported.
  HINT: Recompile this program using: actonc --rts-debug ...
  $

Closes #172.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,9 +103,12 @@ numpy/numpy.o: numpy/numpy.c numpy/numpy.h numpy/init.h numpy/init.c \
 
 
 # /lib --------------------------------------------------
-LIBS=lib/libActon.a lib/libcomm.a lib/libdb.a lib/libdbclient.a lib/libremote.a lib/libvc.a
+LIBS=lib/libActon.a lib/libActonRTSdebug.a lib/libcomm.a lib/libdb.a lib/libdbclient.a lib/libremote.a lib/libvc.a
 
 lib/libActon.a: builtin/builtin.o builtin/minienv.o modules/math.o numpy/numpy.o rts/empty.o rts/rts.o modules/time.o modules/acton/acton$$rts.o $(MODULES)
+	ar rcs $@ $(subst $,\$,$^)
+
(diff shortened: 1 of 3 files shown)


plajjan:
I think this needs some more work. Like maybe verbose & debug should be sort of combine into a log level? Just enabling debug level without also enabling the currently verbose level might be a little weird?

The verbose variable is currently in main(), maybe it should be lifted to a global so we can use this wherever!?

Also noting that the verbose printfs I have in there since before won't work via rtsd_printf as they are not '\n' ended and we don't want to print the prefix multiple times.

nordlander:
LGTM

PR #204, opened 2021-09-08
Add database replication factor argument

plajjan:
We default to 3, which seems sensible for a distributed quorum-based
database. Read and write quorums are replication_factor/2+1.


plajjan:
I think this is simple enough and I don't think I've missed anything, so I'll go ahead and merge it.

PR #243, opened 2021-09-15
Moved timerQ handling to the eventloop thread, switched to microsecon…

plajjan:
…d time resolution internally.

### DIFF
--- a/builtin/minienv.c
+++ b/builtin/minienv.c
@@ -1066,7 +1066,7 @@ void *$eventloop(void *arg) {
         if (kev.filter == EVFILT_TIMER & kev.ident == TIMER_ID) {
             printf("## TIMER event\n");
             if (EV_ERROR & kev.flags)
-                printf("######## EV_ERROR %s\n", strerror(errno));
+                printf("######## EV_ERROR %s\n", strerror(kev.data));
             handle_timeout();
             continue;
         }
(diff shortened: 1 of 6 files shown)


plajjan:
FWIW, the latest commit (fcd881e), just like many of our attempts before, work just fine locally on my machine:
```
== Running test: 178-after
../../dist/bin/actonc 178-after.act --root main
./178-after
counter actor in test, i: 0
## Enqueueing new timed message at baseline 1631781769796331
## Waking up the eventloop
## WAKEUP event
## Current time is setting timer offset 0 sec, 999736000 nsec
error actor sleeping for 5 seconds
## TIMOUT event
## Dequeued timerQ msg with baseline 1631781769796331 (now: 1631781769796632)
counter actor in test, i: 1
## Enqueueing new timed message at baseline 1631781770796331
## Waking up the eventloop
## WAKEUP event
## Current time is setting timer offset 0 sec, 999656000 nsec
## TIMOUT event
## Dequeued timerQ msg with baseline 1631781770796331 (now: 1631781770796710)
counter actor in test, i: 2
counter actor finished in time, exiting happily
```

plajjan:
@nordlander I think we should have squashed away some of the unnecessary flip-flopping commits in the middle before we merged this, that way we would some pollution of git history :)

nordlander:
@plajjan Yup! I left lots of noise left in there...

PR #270, opened 2021-09-19
Build debian package

### LINKED ISSUE #60: Create debian packages
I'd like to produce a debian package of the Acton system so it is even easier to install!

This probably only makes sense after #44

plajjan:
This adds new make targets to build a .deb. We invoke this from CI and upload the resulting .deb as an artifact.

The only caveat is that the pre-release .debs do not reflect the actual version number, for example, actonc will report like 0.6.0.202010919.143412 whereas the .deb file will be just 0.6.0. This is due to the order in which this information is known. We currently generate this build date when compiling the acton compiler but the .deb build process needs to know about it up front. To make it work, we have to move when we generate this. We should take care to not rebuild unless something has actually changed, something that's not just the version number. Not sure if this is possible or not!? Anyway, that's for another issue.

Fixes #60.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,6 +72,33 @@ jobs:
       - name: Run tests
         run: make -C ${GITHUB_WORKSPACE} test
 
+  build-debs:
+    runs-on: ubuntu-latest
+    container:
+      image: debian:sid
+    steps:
+      - name: Set BUILD_RELEASE when we are building for a version tag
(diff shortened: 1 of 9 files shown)


plajjan:
I don't get how I can get CI to process these jobs. I think they are not running because I opened this from a branch on my fork of acton instead of from a branch on this repo. CLosing and opening new PR, see #273.

PR #286, opened 2021-09-21
Move __builtin__.c & .h

### LINKED ISSUE #265: Move __builtin__.c ?
I noticed we still keep this in the builtin directory. I suppose it could be moved to stdlib, in which case the make targets we already have there might work and compile it out of the box?

plajjan:
By moving these files to stdlib/src, our generic Makefile targets that
automatically figure out whether modules should be stub compiled or not
work correctly! Thus we can also remove the special targets to build
__builtin__.ty

Not all positive though, the header include paths are now these rather
weird relative ones. Doesn't look very nice at all.

Fixes #265.


plajjan:
All agree that this is better! Merging.

PR #287, opened 2021-09-21
Move defs to numpy.h

### LINKED ISSUE #278: Files from stdlib/c_src not included in dist
Stdlib C header files that include additional headers from stdlib/c_src no longer work when the including file is copied into dist/types. One example is stdlib/src/numpy.h, which in turn includes "../c_src/ndselect.h". The latter path is only valid inside stdlib/src, not from within dist/types.

Is the solution to copy stdlib/c_src into dist as well (and thereby reinterpret the meaning of "../c_src")? Or to expand the headers copied into dist/types so that any includes from c_src get inlined?

plajjan:
The files we keep in stdlib/src will eventually end up in the output
distribution system path under types/. A .h/.ty file in there is
effectively a module in Acton (and we expect the object file in
libActon.a). This means that the code for a module should either be
contained to that single file (well, .c & .h & .act) or that we somehow
include more files in some other location.

In this case stdlib/src/numpy.h previously included numpy/init.h.
Without qualification, init.h would look like a module called init.
Clearly we can't have that. The numpy internal files need to be hidden
away and preferably, we want to package it up in such a way that what we
do at development time with source file doesn't need to bleed through
to the operational environment (the Acton system path).

We now achieve that by:
- placing the entire interface of the numpy module in stdlib/src/numpy.h
  instead of including 6 other header files
  - including those header files meant that we had to propagate them
    into the final installed Acton system and... it's just messy
- partially linking the numpy.o module so that we don't have to resolve
  its dependencies (for example on libbsd) at actonc run time

This PR is not ready. If we decide to move forward with this proposal, we need to clean away the numpy/*.h files since they are effectively now in stdlib/src/numpy.h. We could also build stdlib/src/numpy.h by concatenating the numpy/*.h files together + some hackity hack.

Fixes #278.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,13 +90,15 @@ stdlib/out/types/%.h: stdlib/src/%.h
 
 stdlib/out/release/%.o: stdlib/src/%.c
 	@mkdir -p $(dir $@)
-	cc $(CFLAGS) -I. -Istdlib/ -Istdlib/out/ -c $< -o$(subst $,\$,$@)
+	cc $(CFLAGS) -I. -Istdlib/ -Istdlib/out/ -c $< -o$@
 
-NUMPY_HFILES=$(wildcard stdlib/c_src/numpy/*.h)
 NUMPY_CFILES=$(wildcard stdlib/c_src/numpy/*.h)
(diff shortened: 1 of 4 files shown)


plajjan:
Closing in preference of #293.

PR #293, opened 2021-09-22
Hide internals of numpy module

### LINKED ISSUE #278: Files from stdlib/c_src not included in dist

plajjan:
A module in Acton correspond to files in the types/ directory etc, which
in turn is the output built from sources in stdlib/src (for all the
built-ins / stdlib). Files are only allowed in this directory with a
direct correspondence to the Acton module.

For example, the numpy module "owns" the stdlib/src/numpy.* "namespace".
Normally there is a .act source file and for modules implemented in C,
there is also a .h and .c file. The .act file and .h both conceptually
contain the interface of the module, just on different levels; in Acton
and on a C level for our intermediate output.

For numpy, we originally chose to have 6 different header files, all
placed in a numpy directory at the root of the acton git repository. When we
recently restructured, we ended up with a single stdlib/src/numpy.h
which then included these 6 header files in stdlib/c_src/numpy/*.h with
some long and weird relative includes. The problem with this is that
includes in stdlib/src/numpy.h need to be resolved at actonc run time,
that is, we must include those libraries with the Acton system. Besides
the logistics inconvenience in including various extra files that follow
no particular structure, it is also bad because it leaks the module
internals to the outside.

To rectify this, the numpy header files in stdlib/c_src/numpy/*.h are
now removed and their content has been inlined into stdlib/src/numpy.h.
It is now the sole header file for the numpy module and contains its
entire interface.

Further, we avoid a similar bleed through of internals by partially
linking the numpy.o file with its dependencies, which in this case is
our own math.o and on linux also libbsd. Outside the compilation of
numpy.o, it is no longer visible what we depend on. As it should!

Fixes #278.


plajjan:
For some reason I'm getting a new warning message:
```
kll@mbp13:~/terastream/acton-mbp$ make -C test/ stdlib/test_numpy
../dist/bin/actonc --cpedantic --root main stdlib/test_numpy.act
warning: (x86_64) /var/folders/cf/z5njv0sd3pdb74937gpdw6b40000gn/T/numpy-34bf8c.o unable to open object file: No such file or directory
./stdlib/test_numpy
1
kll@mbp13:~/terastream/acton-mbp$
```

Tests succeed on this branch, so my fix is working and the above warning is just cosmetic. There are no new warnings from the existing tests, only for this new one in test_numpy.

plajjan:
Thought a little more about this and feel even stronger about this being in the right direction. The warning is something about generating debug symbols on MacOS. It is triggered by these changes, though I don't fully understand why. Before this PR we are inconsistent about how we generate debug symbols (not always doing it) whereas after this MR I think we are consistent, which then triggers a warning. Weird, but that's for later to fix. Feels like one needs to be a professor in macos clang to understand how it handles this.

I'll merge this.

PR #304, opened 2021-09-24
Fold backend/Makefile into main Makefile

### LINKED ISSUE #215: Fold backend Makefile into main Makefile to complete the DAG
Lots of back and forth compilation between these two as there are shared library dependencies... it's a mess. We need to fold the backend Makefile into the top level Makefile to fix it!

- [x] use `$@` and `$<` in make targets
- [x] replace `-I ../src` with `-I.`... I guess?
- [x] move `client.c` to `test_client.c`
- [x] move `server.c` to `actondb.c`
- [x] move backend tests into `backend/test`
- [x] combine all backend libraries into a single `libActonDB.a`
- [x] silence jemalloc check
- [x] move makefile content of `backend/Makefile` to `Makefile`

plajjan:
DAG is soon complete!

There's a bunch of simplifications, like instead of creating a number of archives we now have a single libActonDB.a that we can use everywhere.

The check for jemalloc is now lifted to the main Makefile which means the entirety of Acton will be using jemalloc in case it is installed. Since jemalloc is faster than the defaults I think this is a good thing.

There were many incorrect paths in the make targets that have been fixed / removed. Through variable reuse we also shrink the size of most targets.

Fixes #215.


plajjan:
FYI @aagapi , this moves stuff around a bit so you probably want to rebase an potential ongoing work ASAP. Apologies for possible  conflicts :/

Also hope I'm not stepping on any toes. I think this structure makes things easier, but there is probably an opposing view to that. LMK if we need to revert parts of this.

PR #307, opened 2021-09-24
Improve actonc (parallel) build & fix deb version

### LINKED ISSUE #271: Reflect correct version number in .deb files
.deb files built from the main branch as a pre-release build do not reflect the correct version number. actonc and the tar balls will have a version number that includes the date and time, like 0.6.0.202010919.143412 whereas the .deb file will be just 0.6.0.

Can we fix? We want to try to avoid rebuilding it all just for a new version number. I.e. if we inspect our dependencies, the version number should not affect it and result in always building a new build..

- [ ] top make file should set VERSION_INFO by:
  - check if VERSION_INFO is already set and if so, use it
  - check if actonc exists, ask it
  - set to unknown
- [ ] compiler makefile should only compute a new VERSION_INFO if one is not already set
  - or fold compiler/Makefile into top level makefile?

Follow up from #270.

plajjan:
DAG DAG DAG DAG! The compiler Makefile is now folded into the main
Makefile!

Turned the actonc make target into a non-phony target. Nice! I just hope
I've caught all the files that affect the stack build. Since various
other files, like .ty, depend on actonc, making it non-phony is good
since any target that depends on a phony is always run. Now make will
actually do nothing when there is nothing to do!

Also enabled parallel build of actonc itself using 4 threads. This
brings down the build time on my AMD 5950X from ~55 seconds (single
thread) to ~35 seconds. Higher values do not yield a noticeable
improvement. Given the completeness of the DAG, doing parallel make of
the whole project is quite fast. GitHub Actions don't have that many CPU
cores but 4 is low enough of a value to not trash everything with
context switching... I think.

The manner in which we generate the version number is revamped. We
previously generated it when compiling actonc and then let the rest of
the build query actonc --version to get the version number. In
particular when building a release, that version from actonc would get
inserted into the tar ball name. Debian packages didn't use the same
value for the version but rather just grabbed the VERSION variable from
common.mk, which works out okay for release builds but for tip builds it
wouldn't reflect the build time - bad. The tricky bit was that in a
debian build, the deb target is the first target that gets called, which
in turn sets up a new "clean room" build, so it takes the version number
as input. Since actonc isn't compiled yet, we can't query it!

We have now turned everything around, so we generate the version number
in the makefile. If we are just building actonc etc as normal, then
we'll insert that version number. If we are building a debian package,
then we'll take the version number and feed into the deb build. This
will in turn run the normal build, at which point we've feed in the
version number which will then be used.

Part of #271.


plajjan:
Hmm, it didn't work out completely. The .deb file indeed has the build time in the name but it is supposed to be the same version as actonc will identify itself with... it is not.

```
sudo dpkg -i acton_0.6.2.20210924.14.34.31_amd64.deb
kll@Boxy:~/Downloads$ sudo dpkg -i acton_0.6.2.20210924.14.34.31_amd64.deb 
[sudo] password for kll: 
(Reading database ... 393995 files and directories currently installed.)
Preparing to unpack acton_0.6.2.20210924.14.34.31_amd64.deb ...
Unpacking acton (0.6.2.20210924.14.34.31) over (0.5.3) ...
Setting up acton (0.6.2.20210924.14.34.31) ...
kll@Boxy:~/Downloads$ which actonc
/usr/bin/actonc
kll@Boxy:~/Downloads$ actonc --version
acton 0.6.2.20210924.14.34.33
cc: cc (Debian 10.2.1-6) 10.2.1 20210110
kll@Boxy:~/Downloads$ 
```

plajjan:
Anyway, the improvements to the makefile are good enough that I'll merge this even though it missed the mark on fixing the deb version completely.

PR #308, opened 2021-09-24
Fix deb version

### LINKED ISSUE #271: Reflect correct version number in .deb files

plajjan:
Finally! We now align the version contained in actonc, displayed with
actonc --version and the version of the deb package!

We compute the VERSION_INFO once and then pass it into the debian build
process to tag the inside content.

Fixes #271.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -320,4 +320,4 @@ debian/changelog: debian/changelog.in CHANGELOG.md
 
 .PHONY: debs
 debs: debian/changelog
-	debuild -i -us -uc -b
+	debuild --preserve-envvar VERSION_INFO -i -us -uc -b


plajjan:
Looks good!
```
kll@Boxy:~/Downloads$ sudo dpkg -i acton_0.6.2.20210924.16.58.23_amd64.deb 
[sudo] password for kll: 
(Reading database ... 393985 files and directories currently installed.)
Preparing to unpack acton_0.6.2.20210924.16.58.23_amd64.deb ...
Unpacking acton (0.6.2.20210924.16.58.23) over (0.6.2.20210924.14.34.31) ...
Setting up acton (0.6.2.20210924.16.58.23) ...
kll@Boxy:~/Downloads$ actonc --version
acton 0.6.2.20210924.16.58.23
cc: cc (Debian 10.2.1-6) 10.2.1 20210110
kll@Boxy:~/Downloads$ 
```

Merging!

PR #312, opened 2021-09-27
Do parallel build of Acton

plajjan:
This enables parallel build of Acton.


plajjan:
This actually appears to be slower than the serial build, so not going to merge as-is. GitHub Action runners do not have a lot of CPU. I think the Linux machines run with 2 CPU cores and something similar for the Mac ones. This PR does `-j`, which means unbounded parallel jobs, which means we could get lots of competition over CPU resources and thus, context switches, that bring down efficiency.

Maybe like `-j2` or `-j3` would work better.

plajjan:
This is just a hair faster than without `-j2`. Merging.

PR #327, opened 2021-09-29
Include __builtin__.h in correct path

### LINKED ISSUE #326: brew / apt system installed acton fails with missing include file
```
kll@Boxy:~$ actonc hello.act --root main
In file included from /tmp/actonc-011a0bf606970d6b/out/types/hello.h:2,
                 from /tmp/actonc-011a0bf606970d6b/out/types/hello.c:1:
/usr/lib/acton/builtin/builtin.h:16:10: fatal error: ../stdlib/src/__builtin__.h: No such file or directory
   16 | #include "../stdlib/src/__builtin__.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
ERROR: internal compiler error: compilation of generated C code failed
NOTE: this is likely a bug in actonc, please report this at:
NOTE: https://github.com/actonlang/acton/issues/new?template=ice.md
NOTE: acton 0.6.3 compiled by ghc 8.10 on linux x86_64
NOTE: cc: cc (Debian 10.3.0-11) 10.3.0
```

plajjan:
This reverts #286, which introduced a regression. The benefit of that was slightly more elegant Makefiles. The downside was hairy relative import... it didn't feel good at the time but I did merge it. Turns out it caused more problems, thus I'm reverting it.

I think these are mostly signs that our C files are not entirely well structured. These imports between builtin/* and __builtin__ are not good. Spaghetti-ish.

Fixes #326.


plajjan:
I want to test this by manually installing the produced .deb, before we merge this PR.

plajjan:
I've verified the .deb installed version works correctly!

PR #328, opened 2021-09-29
Get rid of make warning

plajjan:
Warnings make Johan nervous. It was really a cosmetic one but let's calm
him by getting rid of it.

We had a cyclic dependency in our make target definition. The target was
for a wildcard and all .ty files depend on __builtin__.ty, but they are
all built by the same target. Make is clever enough to figure out that
__builtin__.ty cannot depend on itself, so it drops the dependency and
then proceeds to build __builtin__.ty first, as it should. But it also
prints a warning.

By adding an explicit target for __builtin__.ty we get rid of the
warning. Otherwise we accomplish exactly the same thing.


plajjan:
As it turns out, this warning popped up when i restructured the __builtin__ stuff a bit in #286. That caused #326 which I'm fixing in #327 by simply reverting #286. That also means this warning is removed, so this PR is redundant. Closing.

PR #333, opened 2021-10-05
Make actonc build more idempotent

plajjan:
A bunch of targets that create .ty files depend on actonc. If actonc
is fresher than those targets source files, all those targets will be
rebuilt. Although stack is pretty smart with figuring out when to build
a target or not (it doesn't just inspect modification time of files), it
isn't as smart when it comes to "stack install". stack install will
always install the file, thus the target file (compiler/actonc in this
case) will always be refreshed, which in turn triggers running all the
targets that depend on actonc.

There's also the tricky bit of the version timestamp. We want to always
insert a new version timestamp, to show when actonc was built. However,
we don't want to update the version stamp if there are no other changes,
otherwise any build will be a rebuild as the version timestamp will
always change.

To achieve all this, we now have proper dependencies on the makefile but
also check the output of `stack build` and only if it reports that
something needs to build, we update the version timestamp and actually
do the build! Shiny ;)


nordlander:
Elegant!

PR #353, opened 2021-11-09
Add RTS monitor

plajjan:
This adds monitoring function to the run time system. It is enabled with --rts-mon /tmp/mon_socket after which the RTS will listen on the UNIX Domain Socket /tmp/mon_socket. For any incoming request, we will dump the current stats for all worker threads.

There's a new thread to deal with the monitor socket. We only accept a single connection at a time. I want the monitoring thread to be completely standalone, i.e. I do not want to mix in listening on various fds into our normal eventloop. This is simple, which is good.

The protocol is a quick hack of a line protocol. The client sends "WTS" (worker thread stats) and gets a single line reply, which is a JSON document. To encode data we use yyjson, which is now included in our code base.

There is a new utility in utils/actonmon which can connect to the monitoring socket and display thread stats. It does so continuously at 0.5 second intervals. The interval can be set with --interval. It requires the path to the monitoring socket as an argument.

To start examples/count with RTS monitoring enabled:

    examples/count --rts-mon ~/act_mon_socket

Then check the mon utility:

    utils/actonmon ~/act_mon_socket --interval 0.2

There's also a `--rich` option to actonmon to tell it to use the rich library to render a very pretty looking table that is then updated in place.

### DIFF
--- a/rts/rts.c
+++ b/rts/rts.c
@@ -25,6 +25,7 @@
 #include <uuid/uuid.h>
 #include <getopt.h>
 
+#include "yyjson.h"
 #include "rts.h"
 #include "../builtin/minienv.h"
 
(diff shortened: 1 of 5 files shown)


rts/rts.c
```diff
@@ -39,6 +40,54 @@ int opterr = 0;
 char rts_verbose = 0;
 char rts_debug = 0;
 
+long num_wthreads;
+
+char *rts_mon_path = NULL;
+
+struct wt_stat {
+    unsigned int idx;          // worker thread index
+    char key[10];              // thread index as string for convenience
```
plajjan, line 49: We should expose the env.argv[0], i.e. the name of the acton program that we are currently running. Maybe PID too? This way it'll be easier to follow in graphs. Maybe that shouldn't be under the WTS command though? Damn convenient though...

utils/actonmon
```diff
@@ -0,0 +1,344 @@
...
+        self.wfile.write(b'# HELP acton_rts_sleeps Number of times a thread went to sleep\n')
+        self.wfile.write(b'# TYPE acton_rts_sleeps counter\n')
+        for k,v in sorted(data['wt'].items(), key=lambda item: int(item[0])):
+            self.wfile.write(f'acton_rts_sleeps{{thread="{k}"}} {v["sleeps"]}\n'.encode('utf-8'))
+
+        self.wfile.write(b'# HELP acton_rts_conts Execution time of continuations\n')
+        self.wfile.write(b'# TYPE acton_rts_conts histogram\n')
+        for k,v in sorted(data['wt'].items(), key=lambda item: int(item[0])):
+            total['sleeps'] += v['sleeps']
+            for bk in conts_buckets:
```
plajjan, line 71: Too much metaprogramming. Better with the upfront style used in `richmon`. It's a little verbose but oh so much easier to read.

nordlander (approved):

rts/rts.c
plajjan: I've added argv[0] and the pid to the exported data. We're not really doing anything with it in actonmon for now... I want to play around a little more with it but first I want this PR merged, so leaving it like this for now. The future changes are likely in actonmon only and not to the on-the-wire format.

plajjan:
I have now fixed the last couple of things, including to properly free a temporary variable :P 

@nordlander approved so going to merge on CI success. Yay.

PR #360, opened 2021-11-15
Reduce worker threads wake ups

plajjan:
This reduces the number of places where we call new_work(), which wakes
up another worker thread. The basic idea is simple, as soon as we send a
message or do something that leads to new work, we call new_work() to
wake up another worker thread which can start processing that work,
potentially in parallel.

It turns out we woke up new workers in too many places. In particular
from the flush_outgoing function. When a continuation executes, outgoing
messages to other actors are collected in an outgoing queue and at the
end of the continuation, when we commit the result, we flush that
outgoing queue. We would then also call new_work() so that other actors
could start picking things up from the readyQ. It takes a little while
for the other thread to wake up. Since we are just done processing the
current continuation, we'll take another spin, check the readyQ for work
and pick up if any. By the time the other thread has woken up, the
readyQ is empty and it'll go back to sleep.

- T1 reaches end of continuation, flushes outgoing which adds to readyQ
- T1 calls new_work() which wakes up thread T2, though it takes some
  time for the T2 thread to actually wake up
- T1 takes another spin in the loop, checks the readyQ to find new work
  there, which is the item we just added from the previous continuation
- T2 has woken up and checks readyQ to find it is empty
- T2 goes back to sleep

This leads to many unnecessary wakeups, which in itself shouldn't really
be a problem, since work could theoretically still be carried out on one
CPU core while another CPU core is busy waking up etc. However, that's
not the case in practice; throughput overall is quite poor, possibly
because calling new_work(), which acquires a mutex lock, is expensive.

The new code doesn't wake any other threads when flushing the outgoing
queue. Instead we take another spin, check the readyQ, find our own work
and dequeue that from readyQ. Then we check if there is anything else on
the readyQ and only if there is we wake up another thread.

On an AMD5950X (16 cores / 32 threads) ring.act gets the following
results.

Before:

    Actor 0: 1962622
    Actor 1: 1962629
    In total across all actors: 3925251
    Total messages per second: 392525

    real	0m10.048s
    user	0m8.628s
    sys	0m22.109s

After:

    Actor 0: 23865464
    Actor 1: 23865504
    In total across all actors: 47730968
    Total messages per second: 4773096

    real	0m10.230s
    user	0m8.576s
    sys	0m1.657s

About 10x faster with much lower sys time.

ring.act is a rather artificial test and not representative of "real"
programs, however, this change really should not behave worse for any
other scenarios either.

### DIFF
--- a/rts/rts.c
+++ b/rts/rts.c
@@ -142,6 +142,12 @@ pthread_mutex_t sleep_lock = PTHREAD_MUTEX_INITIALIZER;
 pthread_cond_t work_to_do = PTHREAD_COND_INITIALIZER;
 
 void new_work() {
+    // We are sometimes optimistically called, i.e. the caller sometimes does
+    // not really know whether there is new work or not. We check and if there
+    // is not, then there is no need to wake anyone up.
+    if (!readyQ)
+        return;
+
     pthread_mutex_lock(&sleep_lock);
     pthread_cond_signal(&work_to_do);
     pthread_mutex_unlock(&sleep_lock);
@@ -756,7 +762,6 @@ void FLUSH_outgoing($Actor self, uuid_t *txnid) {
             $Actor to = m->$to;
             if (ENQ_msg(m, to)) {
                 ENQ_ready(to);
-                new_work();
             }
             dest = to->$globkey;
         } else {
@@ -1099,6 +1104,7 @@ void *main_loop(void *arg) {
     while (1) {
         $Actor current = DEQ_ready();
         if (current) {
+            new_work();
             pthread_setspecific(self_key, current);
             $Msg m = current->$msg;
             $Cont cont = m->$cont;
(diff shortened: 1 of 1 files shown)


plajjan:
Using the RTS monitoring feature (from #353) I have created a pretty cool graph that shows the speedup here:

2021-11-15-161820_1225x807_scrot

I've run `test/perf/ring.act`. On the left in the graph I am using the old RTS before this patch. We can see how we execute roughly 650k continuations per second. 20% of time is spent in the continuation and about 80% is in bookkeeping. Summing up continuation execution & bookkeeping we see values like 993 ms, so very close to the expect 1 second if we were truly tracking all time spent. Note how this is across many threads, so we could see use of more than 1 second but our work here is really single threaded in nature, thus my expectation of seing around 1 second...

After this patch, we process around 8 million continuations per second by spending about 550ms in continuation execution while bookkeeping falls to ~250ms. We can see how we are nowhere near 1 second, just under 800ms in total. This means we fail to account for some time. We don't measure time doing `DEQ_ready` and some other stuff - but could that really account for this much? Or is it just the time measuring code itself that now runs so often that it starts to account for a larger percentage? I don't know. Not sure it's super valuable to find out either as this test is quite artifical in nature. For more natural programs, it won't be this extreme. I still think it nicely shows the performance improvement from this patch.

nordlander (approved):

Issue #361, opened 2021-11-16
async not working!?


plajjan:
### Code / what did you do

UPDATED, use this code (also added to repo in `test/test_async.act`):

```python
import acton.rts
import time

actor Act(idx):
    def sleep():
        print(idx, "sleeping...")
        acton.rts.sleep(1)
        print(idx, "after sleep")
        return True

actor main(env):

    def test_async1():
        a1 = Act("1")
        print("-- async1 test start")
        t1 = time.time_ns()
        a1.sleep()
        t2 = time.time_ns()
        diff = (t2-t1) / 1000000000
        if diff > 0.1:
            print("a1.sleep() seemingly runs sync :/")
            await async env.exit(1)
        print("-- async1 test done")

    def test_async2():
        a1 = Act("1")
        print("-- async2 test start")
        t1 = time.time_ns()
        r = async a1.sleep()
        t2 = time.time_ns()
        diff = (t2-t1) / 1000000000
        if diff > 0.1:
            print("async a1.sleep() seemingly runs sync :/")
            await async env.exit(1)
        print("-- async2 test done")

    test_async1()
    test_async2()

    await async env.exit(0)
```

Old code, which should also illustrate the problem:

```python
import acton.rts

actor foo():
    def bar(pfx):
        print(pfx, "sleeping...")
        acton.rts.sleep(1)
        print(pfx, "after sleep")
        return 42

actor main(env):
    def test():
        f1 = foo()
        f2 = foo()
        f1.bar("1")
        f2.bar("2")

        await async env.exit(0)

    after 0: test()

```

### Expected outcome

I expected async execution of the bar() method! That is, I should see the first print from each actor method invocation, they should then both sleep for 1 second after which I should see the 'after sleep' message from both.

```
1 sleeping...
2 sleeping...
1 after sleep
2 after sleep
```

### Actual outcome

Instead, all methods appear to run synchronously.

```
kll@Boxy:~/terastream/acton/examples$ actonc --root main test_async.act && ./test_async 
1 sleeping...
1 after sleep
2 sleeping...
2 after sleep
kll@Boxy:~/terastream/acton/examples$ 
```

### Meta

`actonc --version`:
```
acton 0.6.4
compiled by ghc 8.10 on linux x86_64
cc: cc (Debian 11.2.0-10) 11.2.0
kll@Boxy:~/terastream/acton/examples$ 
```

plajjan:
[...]

PR #362, opened 2021-11-17
Add async test

plajjan:
Simple test to check if async functionality is working or not.

We are not running the test since it is currently failing.

@nordlander please have a look if you agree that this test actually tests async, like should these method calls be async or not?

Part of #361


plajjan:
Spoke to @nordlander on the phone about this, who agrees this should work. I'll merge.

PR #370, opened 2021-11-17
Add acton-by-example book

plajjan:
This is by no means complete but a decent start to a book which aims to
explain Acton by example. I want to get the scaffolding in place and we
can add to it bit by bit.


plajjan:
I forgot to explain one detail; I deliberately added this book to the main acton git repository despite it being significantly easier to add it to the web repo (actonlang/www.acton-lang.org) since the final output is to be built and published on the web page. The reason is that I want it to be easy to update the documentation and keep it in line with the actual language. It is significantly easier to do this when they are in the same repository. Corollary, it is easy to forget when documentation lives in a separate repo. Also, I want to run through all the examples in CI to ensure they are actually correct.

PR #387, opened 2021-11-23
Switch linux test to run on Ubuntu 18.04

plajjan:
We continue building the .deb files on Debian but the standard Linux
test is better run on Ubuntu 18.04, since it is an older platform, so we
cover more ground than running everything on Debian.

The homogeneity we had with debian was mostly because it was simple. Now
that it is working we want to improve coverage across versions etc.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,9 +48,7 @@ jobs:
 
 
   test-linux:
-    runs-on: ubuntu-latest
-    container:
-      image: debian:bullseye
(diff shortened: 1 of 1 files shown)


plajjan:
Due to some bug in the gcc version that ships with Ubuntu 18.04, this fails to build.

It is currently not a very important goal to support older releases and so I will close this. We might revisit it in the future, trying to work around the issue, but given all the areas in which we have a fair bit of work cut out for ourselves, it's not very likely.

PR #388, opened 2021-11-23
Built & test on debian:buster

plajjan:


plajjan:
Due to some bug in the gcc version that ships with debian:buster, this fails to build.

It is currently not a very important goal to support older releases and so I will close this. We might revisit it in the future, trying to work around the issue, but given all the areas in which we have a fair bit of work cut out for ourselves, it's not very likely.

PR #403, opened 2021-12-09
Expand on dev / rel compiler profiles

### LINKED ISSUE #399: Don't do -O3 everywhere
I guess we really want this to be dependent upon the compiler profile, but that would require compiling everything twice. Not sure if we are ready for this?

If not, it would be better to do -O0 today to simplify debugging.

plajjan:
We now have two libAction:
- libActon_dev, for the development profile
- libActon_rel, for the release profile

This is more consistent than the previous solution where we always used
libActon and override certain modules (rts.o) that were specially
compiled with libActonRTSdebug. The two new archives are complete and do
not rely on each other in any way.

Thus, actonc will now link using libActon_dev OR libActon_rel.

All o-files are now compiled as a _dev.o and _rel.o file, using
CFLAGS_DEV and CFLAGS_REL for the respective flags. Right now, those are
-g for dev mode and -O3 for release mode.

Fixes #399

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,9 @@ else
 export VERSION_INFO?=$(VERSION).$(BUILD_TIME)
 endif
 
-CFLAGS+=-g -I. -Ideps -Wno-int-to-pointer-cast -Wno-pointer-to-int-cast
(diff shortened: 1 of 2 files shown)


plajjan:
@aagapi so FYI, with this, you can now get that debugging simply by telling actonc to use the dev profile with `actonc --dev ...` :)

PR #460, opened 2022-01-12
Exit App / RTS cleanly

### LINKED ISSUE #454: Rework app exit (env.exit())
Calling `env.exit()` from within an Acton application will run `exit(x)` (the C function) and thus exit the entire program more or less immediately. It'd probably be nice to let currently running continuations finish before we wrap up and call it a day.

plajjan:
This changes how env.exit() is implemented and rather than directly
calling the C exit() function and abruptly killing everything, we now
let the RTS worker threads process any currently running continuation
until completion and then shut them down one by one until no worker
threads remain.

Fixes #454.

### DIFF
--- a/rts/rts.c
+++ b/rts/rts.c
@@ -35,8 +35,10 @@
 
 char rts_verbose = 0;
 char rts_debug = 0;
-
 long num_wthreads;
+
+char rts_exit = 0;
+int return_val = 0;
 
 char *rts_mon_path = NULL;
 
@@ -1173,6 +1175,15 @@ void *main_loop(void *idx) {
 
     struct timespec ts1, ts2, ts3;
     while (1) {
+        if (rts_exit) {
+            wt_stats[(int)idx].state = WT_NoExist;
+            rtsd_printf(LOGPFX "Worker thread %d exiting\n", (int)idx);
+            // Wake up all sleeping threads so they too can exit
+            pthread_mutex_lock(&sleep_lock);
+            pthread_cond_broadcast(&work_to_do);
+            pthread_mutex_unlock(&sleep_lock);
+            break;
+        }
         $Actor current = DEQ_ready();
         if (current) {
             new_work();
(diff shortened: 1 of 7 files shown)


plajjan:
@nordlander can you please review this?

rts/rts.c
```diff
@@ -1699,23 +1710,26 @@ int main(int argc, char **argv) {
...
     }
     num_threads = num_threads + num_wthreads;
 
-    for(int idx = 0; idx <= num_threads; ++idx) {
+    // Only joining the worker threads, we don't care about gracefully shutting
+    // down IO / Mon threads
+    for(int idx = wthread_start; idx < num_threads; ++idx) {
         pthread_join(threads[idx], NULL);
+        pthread_mutex_lock(&sleep_lock);
+        pthread_cond_broadcast(&work_to_do);
```
plajjan, line 1731: I do not really know why this is necessary but without it, I saw intermittent hangs in our CI tests on Mac OS which I could also reproduce locally on my Mac, though not on Linux.

I'm quite confident this current code works, I have tested hundreds of runs on my local Mac without hangs.

plajjan: @nordlander if you can tell me why the pthread_cond_broadcast in main_loop is not enough, I'm all ears...

nordlander: Hmm. I wonder if this is a possible scenario: worker threads 1 and 2 are busy, all the others are sleeping. Thread 1 executes a continuation that calls exit, so the flag is set and all the sleeping threads are woken up -- except #2 which is already awake! So all threads except #2 now terminate as they should, but #2 is unaware of all this and eventually goes to sleep when it finds the readyQ empty.

So it's probably necessary to check the exit flag before going to sleep. Maybe that's where the exit handling code really should be, and not at the head of the loop?

plajjan: I can't remove it from the head of the loop because that would mean a worker thread will continue running as long as there is something on the readyQ - we have to be able to get out of that loop, but you are right about the sleep stuff I think.

plajjan:
Pushed new version which checks exit inside of the mutex before sleep... Now works reliably on both Mac os and Linux (done 10k rounds). Previous to fix it would fail mostly in under 10 tries on Mac, sometimes <50. This feels good.

plajjan:
Spoke to @nordlander. Agreed this is all good. Merging.

PR #469, opened 2022-01-15
Fix epoll disconnect detection

### LINKED ISSUE #441: TCP socket disconnect leads to continuous on_receipt() being called with empty content
EPOLLHUP vs EPOLLRDHUP

plajjan:
We have to look for EPOLLRDHUP in addition to EPOLLHUP!

Fixes #441.


plajjan:
@nordlander although I have already merged this (it works for me :tm:), I do think it would be good if you could have a look at it as well.

nordlander:
Looks good to me! And in any case, this is code I don't have the ability to test so I trust your judgement :-)

PR #473, opened 2022-01-17
Add DB client lib statistics

### LINKED ISSUE #452: Add DDB client statistics to RTS
- counters
  - sent RPCs
  - received RPCs
  - timed out RPCs
  - no quorum reached within timeout? (is there a difference between this and above?)
  - I think all DB client functions use the same underlying RPCs, so it's better to instrument the common RPC sending functions than each individual higher level function (like enqueue_bla), right?

- do we keep statistics per worker thread or globally for the RTS?
- statistics per DDB remote server node?
  - would be nice to be able to say "node 2 is slow / down"

plajjan:
This adds some statistics for the DB client library so we can get an
idea of how much time we are spending waiting on the DB.

All the functions / operations in the DB client library have been
instrumented to track the following:
 - called;     // Number of calls of this op, incremented on start
 - success;    // Number of successful calls of this op
 - error;      // Number of failed calls of this op, includes all errors
 - no_quorum;  // Number of failed calls with NO_QUORUM_ERR
 - no_msg_cb;  // Number of failed calls with NO_SUCH_MSG_CALLBACK
 - time_sum;   // nanoseconds spent waiting for op to complete
 - time_100ns; // bucket for <100ns
 - time_1us;   // bucket for <1us
 - time_10us;  // bucket for <10us
 - time_100us; // bucket for <100us
 - time_1ms;   // bucket for <1ms
 - time_10ms;  // bucket for <10ms
 - time_100ms; // bucket for <100ms
 - time_1s;    // bucket for <1s
 - time_10s;   // bucket for <10s
 - time_100s;  // bucket for <100s
 - time_inf;   // bucket for <+Inf

The statistics are then exposed via the normal RTS mon functionality.

Fixes #452.

### DIFF
--- a/backend/client_api.c
+++ b/backend/client_api.c
@@ -20,6 +20,77 @@
 
 #include "client_api.h"
 
+struct dbc_stat dbc_stats;
+
+void zeroize_dbc_ops_stats(struct dbc_ops_stat *ops_stat, const char *name) {
+	ops_stat->name = name;
+	ops_stat->called = 0;
+	ops_stat->completed = 0;
+	ops_stat->success = 0;
+	ops_stat->error = 0;
+	ops_stat->no_quorum = 0;
+	ops_stat->time_sum = 0;
+	ops_stat->time_100ns = 0;
+	ops_stat->time_1us = 0;
+	ops_stat->time_10us = 0;
+	ops_stat->time_100us = 0;
+	ops_stat->time_1ms = 0;
+	ops_stat->time_10ms = 0;
+	ops_stat->time_100ms = 0;
+	ops_stat->time_1s = 0;
+	ops_stat->time_10s = 0;
+	ops_stat->time_100s = 0;
+	ops_stat->time_inf = 0;
+}
+
+void init_dbc_stats() {
+
+#define X(ops_name) \
+	dbc_stats.ops_name = malloc(sizeof(struct dbc_ops_stat)); \
+	zeroize_dbc_ops_stats(dbc_stats.ops_name, #ops_name);
+LIST_OF_DBC_OPS
+#undef X
+
+}
+
+void stat_start(struct dbc_ops_stat *ops_stat, struct timespec *ts_start) {
+	ops_stat->called++;
+
+	clock_gettime(CLOCK_MONOTONIC, ts_start);
+}
+
+void stat_stop(struct dbc_ops_stat *ops_stat, struct timespec *ts_start, int status) {
+	ops_stat->completed++;
+
+    struct timespec ts_stop;
+	clock_gettime(CLOCK_MONOTONIC, &ts_stop);
+	long long int diff = (ts_stop.tv_sec * 1000000000 + ts_stop.tv_nsec) - ((long long int)ts_start->tv_sec * 1000000000 + (long long int)ts_start->tv_nsec);
+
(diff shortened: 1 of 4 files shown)


plajjan:
I want to do this per RPC type already from the beginning, so converting this to draft for now.

aagapi:
This looks great so far, nice work! Please add a run-time configuration switch and/or a compile-time macro to disable stats bookkeeping if we need to do that for performance reasons.

backend/client_api.c
```diff
@@ -641,15 +712,19 @@ int send_packet_wait_replies_sync(void * out_buf, unsigned out_len, int64_t nonc
 		return ret;
 
 	// Wait for signal from comm thread. It will come when 'db->quorum_size' replies have arrived on that nonce:
+	int res = wait_on_msg_callback(*mc, db);
```
plajjan, line 715: Unnecessary change. This was from an earlier attempt - revert!

backend/client_api.c
```diff
@@ -20,6 +20,77 @@
 
 #include "client_api.h"
 
+struct dbc_stat dbc_stats;
+
+void zeroize_dbc_ops_stats(struct dbc_ops_stat *ops_stat, const char *name) {
+	ops_stat->name = name;
+	ops_stat->called = 0;
```
plajjan, line 27: Add `completed` counter that counts how many completed operations we have. It should be equal to `called` or the difference are currently outstanding operations. Success + error should equal `completed`.

plajjan:
I just realized this isn't thread safe. For the other stats collect the information is kept per thread so there is ever only one thread incrementing a counter. Here there will be many worker threads that use the same db client and thus write to same counter.

I'm inclined to move ahead with this code anyway. I think the risk is quite small and even if the data isn't entirely correct it is still valuable. I think computing the number of in flight requests is probably the thing that runs the highest risk of being wrong. The other data still makes sense if somewhat approximate.

Last, since the RTS is based on the ideas of the seastar library, whose sort of core idea is a shared nothing architecture, I always imagined we would move to separate the database client per worker thread so there's nothing shared between then. Then this becomes a non issue...

plajjan:
Thought about this some more, in particular the non thread safety, potential performance impact etc and find that the benefit of having this outweigh any downsides. I'm fine rewriting all of this later or discarding it entirely if it's not up to par. As for performance and adding compile time flags to disable this, I don't feel it's worth it right now. I've done some performance measurements and I think this code is negligible in the big picture, I'll try to post some figure in another issue / PR on that. I just want to keep things simple.

aagapi:
Good work.
Regarding thread safety: A good "shared nothing" solution that doesn't involve any thread synchronization, locks etc, would be: 1) keep counters per-thread and  update / increment them by each thread to their private counter separately; 2) At "read time" -> i.e. when statistics are generated to be output in the JSON object, add the per thread counters to a global total. This would be a pretty simple solution with no synchronization overhead and almost no performance overhead.
Regarding counter correctness, I think this is might still be important for some counters, especially when looking for error conditions. I.e. it is OK to have approximate values for things like RPC throughput, but it makes a difference if we have 0 or 1 requests in-flight that haven't been ACK-ed yet. So I think there is some value in making the above change.
This is still true even if we chose to have separate DB drivers per-thread, we might still be interested in global stats per-node.

PR #487, opened 2022-01-20
Segfault on mac

aagapi:
Fix segfault on OSX on uuid_compare()


backend/skiplist.c
```diff
@@ -33,7 +33,8 @@ int long_cmp(WORD e1, WORD e2) {
 
 int uuid_cmp(WORD e1, WORD e2)
 {
-	return uuid_compare(*((uuid_t*) e1), *((uuid_t*) e2));
+	return memcmp(e1, e2, sizeof(uuid_t));
```
plajjan, line 36: Isn't this more of a workaround rather than a true fix? No pointer problems so we were looking in the wrong place? Or why use memcmp over uuid_compare?

plajjan (commented):
Please squash the ping-pong commits of going -> memcmp -> uuid_compare -> memcmp :)

PR #492, opened 2022-01-26
Generate & check env.c & env.h

### LINKED ISSUE #485: Generate env.c
The environment stuff is built into Acton and consists of manually written C code. One of the big challenges we have with this is that the manually written C code needs to match the code that actonc generates. Simple in practice, but as we improve actonc and our output changes ever so slightly, we must keep env.c up to date. As it turns out, this is a challenge and we have failed numerous times.

Can we improve the situation and generate parts of env.c? I think so.

My idea is roughly:

1. rename env.c to env_in.c
2. generate env_cgen.c from stdlib/src/__builtin__.act
3. extract function signatures from env_cgen.c and env_in.c
   - error out if mismatch - they must match!
4. remove functions in env_cgen.c that match function signatures in env_in.c
5. produce env.c by merging env_cgen.c and env_in.c

- [x] not possible to stub & cgen compile stdlib/src/__builtin__.act #486 
- [ ] should we cgen entire stdlib/src/__builtin__.act or first cut out just the Env part of it? including only:
  - actor Env
  - actor Connection
  - actor RFile
  - actor WFile

I did try to extract just the Env part into a new act file, like and stub&cgen compile that but got another error:
```
kll@Boxy:~/terastream/acton/stdlib/src$ ../../dist/bin/actonc --stub --cgen env.act

******************** Type error
 10:36-40
   |
10 |    openR        : action(str) -> ?RFile
   |                                   ^^^^^
Cannot infer RFile < __builtin__.RFile

kll@Boxy:~/terastream/acton/stdlib/src$ 
```

plajjan:
We have suspected that our env.c and env.h files have been quite out of date and as a consequence of, we have seen various bugs, such as segfaults. env.[ch] were originally generated from actonc by stub compiling __builtin__.act, extracting relevant portions pertaining to the env and manually modifying functions to implement the desired functionality. As the compiler has been updated, the generated&modified files and what the compiler outputs has been able to drift apart which could have catastrophic consequences. It is important that they align on key points, like arguments and return values.

This PR adds a script that can both check and generate parts of the content in env.c and env.h. To avoid having separate files - one for generated content and another for the manually edited content, we choose to have it all in one file and instead rely on markers to partition the file into one part for generated content and the rest being artisanally edited by hand. Any function found in the hand-edited part will not be inserted into the generated part.

The --check option verifies all function signatures and structs. For some structs that we know are actor struct, we apply a "prefix check" where we only verify that the first fields of the struct match between the generated and hand-crafted versions. This makes it possible to extend a struct with further fields to contain the state necessary for an actor to do its job. Such fields can be used to hold internal state of the actors, only visible from the underpinning C code.

The generation is not perfect. It currently produces new() functions which is produced for all classes, even if a class happens to be an actor, in which case only a newact function should be produced. It's really sort of cosmetic since nothing will ever call the new() function of an actor. However, since our hand-crafted actors do not use the normal CPS converted initialization code, our __init__ functions return type and arguments differ. This is OK but the generated new() function rightly assume the CPS converted version. Currently fixed by manually removing the new() functions - though they will reappear on the next --gen run. Should fix via #497 eventually.

We do not run the generation of code or even the check automatically as either would fail. The --check would need exceptions for the __init__ functions as per above. We could implement that but I want it to be pretty narrow, not to have a blanket exception for the __init__ signature. I want it to except the particular diff we are seeing now.

I think we've already achieved one of the goals that we set out to do, which is that we have sort of gone through the env code half manually but with the help of these tools. It is actually looking pretty good.

The script itself is rather ugly now, but I think that's okay as it's the product of experimentation and we really didn't know what we set out to achieve. We have honed in quite a bit. My plan is that we can just wait for the next time we change something in the compiler and we run into a env problem, then look at the diff / errors and sort of slowly have this script converge on what we might see as an emerging usage pattern.

Fixes #485.


plajjan:
@nordlander please have a look at the code here and see if you can sort out the differences and make this compile :)

PR #499, opened 2022-02-01
Add monitor interface to ActonDB

### LINKED ISSUE #401: Add mon socket to actondb
I propose we do the same thing as for the RTS mon socket, i.e.:
- a local unix domain socket
- JSON encoding
- right now, one simple command to return server status with membership information

plajjan:
This adds a monitor interface to the ActonDB, just like we support in
the RTS. It uses a unix domain socket to expose JSON encoded data via
netstrings as a line protocol.

There is only a single supported query command now called "membership"
and it returns the local node ID and the vector clock view. By comparing
this view between the nodes in a cluster we can see if they all agree on
whose a member.

Unlike the RTS, the implementation here does not use a separate
monitoring thread but is rather part of the one big main loop that we
have. This means we do not have to worry about locking data structures.

Similar to the RTS, we limit ourselves to a single client at a time. The
second client that tries to connect will simply have to wait for the
first one to finish.

actonmon supports fetching the DB membership information but doesn't
bother pretty printing it, rather just dumping JSON to the user. My plan
is to wait a bit and see what other information we add before going all
fancy with formatting. Not currently sure what output format is best...

Fixes #401.


aagapi:
It looks good, please merge when ready.

PR #504, opened 2022-02-03
Separate env.listen error callback

### LINKED ISSUE #471: Separate listen failure from connection handler
When we establish a new listen socket we specify a connection handler as the second argument. This handler is called when there is a new connection but it is also called when there was an error binding as can be seen in the code
https://github.com/actonlang/acton/blob/main/builtin/env.c#L704-L705

I don't like the connection handler dealing with all of this. I think it would be better if we had a separate callback or just let env.listen throw an exception or return some form of error?

plajjan:
env.listen(port, cb) has previously taken a single callback cb that is
invoked both when there are incoming connections to the listening socket
as well as for errors, in which case the callback was called with a None
argument. This is confusing, bad, bla bla. Better to have a separate
callback, which we are now introducing, so the new signature is:

    env.listen(port, cb_on_connect, cb_on_error)

An error when trying to establish the listening socket will now be
reported by calling the callback function provided via cb_on_error.

There's a new test case to cover precisely this test case.

Fixes #471.

### DIFF
--- a/builtin/env.c
+++ b/builtin/env.c
@@ -285,22 +285,25 @@ $l$3lambda $l$3lambda$new($Env p$1, $str p$2, $int p$3, $function p$4) {
     return $tmp;
 }
 struct $l$3lambda$class $l$3lambda$methods;
-$NoneType $l$4lambda$__init__ ($l$4lambda p$self, $Env __self__, $int port, $function cb) {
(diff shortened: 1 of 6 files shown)


plajjan:
@nordlander can you please have a look at this to see if it makes sense? or perhaps @sydow ? :)

plajjan:
@nordlander is ok with this as per video call, merging.

Issue #440, opened 2022-01-07
Trigger socket disconnect on App recovery from DB


plajjan:
The idea for dealing with sockets during Application recovery from the database is to simply invoke the error handler and let the application deal with the problem, typically reconnecting / starting the listen socket again.

In order to do this, we need to save a tad more state in env... a list of the error handler callbacks, right?

plajjan:
If we can get this in, we have a good interface for communicating with our Acton applications in a programmatic way (stdin/stdout sucks, as we've discussed). This in turn enables us to write much better tests that are both more robust and can run much faster. As you indicated this was a fairly small thing, I'm putting this right at the top of the priority list and assigning you :)

PR #506, opened 2022-02-04
Add ListenSocket

plajjan:
This further builds on #504 (so the diff includes those changes too, check out the last commit in this PR to get the really relevant stuff) and takes us one step closer to app resume from DB of TCP listen sockets.

env.listen() sets up a TCP listening socket but we never had a handle on
the listening socket itself, so for example we could not close it. There
is now a new type of actor called ListenSocket, which provides precisely
this functionality. env.listen() now returns a ListenSocket.

The error handling is a little weird, like env.listen() is defined to
maybe return a ListenSocket (otherwise None), which we could use to
signal that there was an error setting up the socket. Right now we
always return a ListenSocket though. Since the user provides an error
callback to env.listen(), we use that to signal back if there was a
problem setting up the listening socket, which is sort of weird control
flow. But if we replace that by just returning None, we don't know what
the error is. Should we throw an exception instead? Should the error
callback get an argument with the type of error?

We need the error callback regardless, because that is how we can send
errors later on, after the socket has been opened. That doesn't normally
happen on a unix system (I don't know what it would at least) but it
would happen when we resume app state from the DB...

I'm doing a pass on the error management for now. I think the most
important design criteria is that it becomes simple for the application
developer to handle errors both for initial socket setup as well as for
handling later errors (app resume from DB). Also, I think we should
clean up fd_data in case there is an error in env.listen().

We save the cb_on_error callback in the state of the ListenSocket. This
is what needs to be called when we resume from DB to tell the app the
ListenSocket is broken and needs to be replaced by calling env.listen()
again. I'm not sure where or how in the deserialization code we can call
this callback, do we need a new function, like __resume__?

The TCP test app now saves the ListenSocket in a variable. Had to
restructure the program somewhat to avoid what appears to be a deadlock.
I think it's the deactorizer acting up, which is being worked on
separately anyway.

Another issue is that the ListenSocket actor does not appear to be
serialized to the database at all. We never call any method on it, so
the RTS doesn't serialize it. We force serialization in rts.c of the env
and root actor. We could do something similar in env.listen() but that
requires quite a restructuring of code because we can't reach the right
methods from env.c.


plajjan:
@nordlander please have a look! Note how this includes the same changes that are in #504. Look at https://github.com/actonlang/acton/pull/506/commits/b0da761b299f8ce05ed97b7e87fbf210b4091982  to just get the commit that adds ListenSocket.

As noted in the commit, we have some further challenges to address before we have a solution to #440 but I think this is a step in the right direction. I made a pass on error handling - I think we can discuss that and modify accordingly, but there is no need to let this improvement await consensus on improved error handling.

I worked around the async deadlock by restructuring the test program. I presume your code will fix this eventually.

PR #512, opened 2022-02-15
Added helper function...

nordlander:
...so that manually constructed actors can be conveniently serialized after initialization.


plajjan:
While it initially failed CI, I'm pretty sure that has nothing to do with this patch, so I'm merging this. I'll fix the other thing separately.

PR #517, opened 2022-02-17
Add actor resume method

plajjan:
The __resume__ method is run immediately after an actor has been
deserialized from the database upon resumption of the system and allows
us to add extra hooks when necessary. For example, for socket related
actors, we want to invoke the error callback so that the user
application code can recreate the socket.


plajjan:
I've held this off because the DB tests were broken. That's fixed now with #518. I've rebased this branch to ensure tests pass with the fixed tests. Merging.

PR #518, opened 2022-02-17
Correctly handle return values from DB tests

plajjan:
What a blunder... we tested, but we sort of ignored the return value,
duh....

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -47,7 +47,7 @@ test_db_app_recovery:
 	./test_db.py TestDbApps.test_app_recovery
 
 test_db_app_tcp:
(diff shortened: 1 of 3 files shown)


plajjan:
Okay, so fixing the test script to correctly propagate errors reveleaed that we actually fail the DB tests a lot, in particular on MacOS. Investigation of that happened in #523 and outcome was segfaults due to unitialized actor attributes. I rewrote the test applications to avoid that pattern. I've also improved retry handling in the test script. It's a little messy but seems to work fairly well.

It's passed CI twice - I did a re-run - so hopefully that's a relatively good indicator this actually works. We wouldn't want a lot of intermittent failures in CI.. anyway, merging!

PR #528, opened 2022-02-22
Do not name main thread

### LINKED ISSUE #526: Do not name the main thread
A while ago I started naming all thread with pthread_setname. I think it's a mistake to name the main thread. It is now literally called `main` and this is what shows up in dmesg when it segfaults and in some top / ps output. If we don't explicitly name it, we should just see the process name rather than the thread name.

So let's revert this. It's better to see the process name, which is unique per acton program, than "main" which all acton programs will have.

plajjan:
Naming the main thread masks out the process name in some top / ps
output and similar. It is more useful to show the process / binary name
rather than the thread name (which would always be 'main').

Fixes #526.

### DIFF
--- a/rts/rts.c
+++ b/rts/rts.c
@@ -1641,12 +1641,6 @@ int main(int argc, char **argv) {
     // Do line buffered output
     setlinebuf(stdout);
 
-#if defined(IS_MACOS)
-    pthread_setname_np("main");
-#else
-    pthread_setname_np(pthread_self(), "main");
-#endif
-
     // Ignore SIGPIPE, like we get if the other end talking to us on the Monitor
     // socket (which is a Unix domain socket) goes away.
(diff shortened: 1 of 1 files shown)


plajjan:
segfault now looks good, we see name "count" rather than the generic "main". I did the before / after in reverse order... and I made the RTS deliberately segfault!

```
[3031058.943825] count[1023984]: segfault at 0 ip 0000000000403871 sp 00007ffd4bc690f0 error 6 in count[403000+60000]
[3031084.775960] main[1024165]: segfault at 0 ip 0000000000403885 sp 00007fff80bb3dd0 error 6 in count[403000+60000]
```

Setting to merge on CI completion.

plajjan:
Rebased to pull in some more fixes to DB testing that's causing CI to fail. Not related to this patch.

Issue #524, opened 2022-02-22
Remove actor parameter pruning analysis when full actor code is not known, i.e. with externally defined code (in C)


plajjan:
We only want to prune actor parameters when we have full knowledge of the actor code, since only then are we able to do a reasonable job at it. If the module is implemented in C (which we today roughly call "stub mode") or has external functions (also in C), i.e. marked as NotImplemented and there is a .ext.c file, we do not have visibility into the full actor code and thus the pruning analysis cannot possibly do a correct job. We should err on the safe side and let all parameters remain.

plajjan:
Hmm, this probably needs to be reworded. We only want to prune actor parameters when we have full knowledge of the actor code. If the module is implemented in C (which we today roughly call "stub mode") or has external functions (also in C), i.e. marked as NotImplemented and there is a .ext.c file, we do not have visibility into the full actor code and thus the pruning analysis cannot possibly do a correct job. We should err on the safe side and let all parameters remain.

plajjan:
Fixed in 8145aac6b8658eaa48d643e22b87be46e60e4c6c, closing.

PR #529, opened 2022-02-22
Call error handler on resume of ListenSocket

### LINKED ISSUE #440: Trigger socket disconnect on App recovery from DB

plajjan:
When a ListenSocket actor is resumed from the database, its error
handler is now called so that the Acton application is notified that the
ListenSocket is no longer working.

We unconditionally call the error handler because there is no way that
the ListenSocket survives a resume. We serialize it many times to the
database in normal operations but we only deserialize and resume it when
we start up fresh (or in the future when we migrate an actor). Either
way, the listen socket has a local fd in the kernel and that will not
survive either circumstance so it assuming it broken and reporting via
the on_error callback is the only sane recourse. The Acton application
should have an error handler that sets up the listening socket anew.

We really should not need to create the cb_err attribute, the actor
argument cb_on_error should be saved and available, however the compiler
currently prunes it away. It runs an analysis to see if actor arguments
are used by methods in the actor, in which case they are persisted.
Since this is a abstract actor with stub methods, the analysis comes up
with the wrong result that the arguments are not needed. We will try to
address this separately, see #524.

Fixes #440.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,10 +79,10 @@ backend/actondb: backend/actondb.c lib/libActonDB.a
 		$(LDLIBS)
 
 backend/%.o: backend/%.c
-	$(CC) -o$@ $< -c $(CFLAGS)
+	$(CC) -g -o$@ $< -c $(CFLAGS)
 
 backend/failure_detector/%.o: backend/failure_detector/%.c
-	$(CC) -o$@ $< -c $(CFLAGS)
+	$(CC) -g -o$@ $< -c $(CFLAGS)
 
 # This target is just to override the above target and turn it into a NOP for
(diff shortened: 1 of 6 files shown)


plajjan:
Alright, so a lot of retries and fixes here. Turned out that ASYNC didn't work as it does pthread_getspecific which wasn't initialized. As we're fairly used to by now, the value defaults to 0 on Linux and thus works whereas on Mac it is some random value which broke thing. Now fixed by explicitly initializing. Merging!

Issue #542, opened 2022-02-24
TCP Connection receive handler is never called on Linux


plajjan:
When we initiate an outgoing connection with `env.connect`, get back a connection and then install a receive handler using `on_receive`, that handler is not called when we receive data. This only happens on Linux, so is likely an epoll problem.

PR #543, opened 2022-02-24
Verify epoll & kqueue return values

plajjan:
There appears to be bugs lurking, so now we check that the return values
are as expected...

Part of troubleshooting #542.

### DIFF
--- a/builtin/env.c
+++ b/builtin/env.c
@@ -18,6 +18,7 @@
 #endif
 #endif
 
+#include <assert.h>
 #include "env.h"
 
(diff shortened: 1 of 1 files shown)


plajjan:
This isn't passing because we get some errors back, but perhaps that is ok. I don't think we are going to move forward on this one, rather we are going to attempt migrating to libuv, so it's better we try to ensure we deal with errors in that new implementation rather than trying to slap it on our own epoll/kqueue stuff. Closing this.

PR #545, opened 2022-02-24
Fix epoll stuff for TCP client connection

### LINKED ISSUE #542: TCP Connection receive handler is never called on Linux

plajjan:
We're getting an error when we try to add the same fd to epoll twice but with different filters. By clearing out the old stuff first, we can then register our interest in EPOLLIN.

Fixes #542.

### DIFF
--- a/builtin/env.c
+++ b/builtin/env.c
@@ -114,6 +114,11 @@ void EVENT_del_read(int fd) {
     fd_data[fd].event_spec.data.fd = fd;
     epoll_ctl(ep, EPOLL_CTL_DEL, fd, &fd_data[fd].event_spec);
 }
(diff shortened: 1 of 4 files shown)


plajjan:
So we now condition some stuff on running on Linux, since the kqueue & epoll API are not 1:1 with each other so we actually need to do different things. @nordlander I would like to get your feedback on how to handle this. It's an ifdef for now and I'll merge this so we can move forward, but not sure how to keep this code base maintainable over time... seems iffy with ifdefs...

PR #559, opened 2022-03-02
Fix duplicate nonces in msg_callbacks skiplist

aagapi:


plajjan:
@aagapi I noted that the last failures wasn't due to segfault. It looks like the TCP connection to the DB server isn't working... I don't think I've ever seen that before. It seems unlikely that it was introduced by this branch since there are only changes to client_api.c, still something is happening. Could we trigger a new code path in the server by the client changes?

I think we need to get more debug output here...

plajjan:
@aagapi FYI, as for the segfault core collection, I did add that in #564 but I have yet to trigger a seg fault so haven't been able to verify if it works correctly or not.

aagapi:
> 
Hm, that's interesting. It is possible in principle that a new issue was triggered somehow by fixing this one.. Do you have a stack trace (or connection failure logs) from MacOS?

plajjan:
I don't think the DB server is segfaulting because I should capture all core files in the test/ directory as artifacts from the CI jobs if it fails and I don't see any core files.

The output looks like this:
```
TcpCmdError: Process is dead, returncode: 0  stdout: b'#RTS# Detected 3 CPUs: Using 4 worker threads, due to low CPU count. No CPU affinity used.\n#RTS# Using distributed database backend replication factor of 3\n#RTS# Using distributed database backend (DDB): 127.0.0.1:16000\n#RTS# Using distributed database backend (DDB): 127.0.0.1:16001\n#RTS# Using distributed database backend (DDB): 127.0.0.1:16002\n#RTS# Checking for existing actor state in DDB.\n#RTS# Found 7 existing actors; Restoring actor state from DDB.\nHost disconnected , ip 255.127.0.0 , port 21536 \n'  stderr: b'ERROR writing to socket: Bad file descriptor\n'
```

The last messages are:
```
Host disconnected , ip 255.127.0.0 , port 21536
```
which I believe is https://github.com/actonlang/acton/blob/10aa5e6d2bf21aafc5b32932fd07f7f8a1e2997f/backend/client_api.c#L159. It appears to suffer from some little / big endian order problem, maybe not correctly mapped from network order to host order, because surely the IP address is not 255.127.0.0...

There is an error message just like it in actondb, but this is the output of the Acton app and not from the DB server...

and on stderr:
```
ERROR writing to socket: Bad file descriptor
```
which should be https://github.com/actonlang/acton/blob/10aa5e6d2bf21aafc5b32932fd07f7f8a1e2997f/backend/client_api.c#L613

The DB servers should write a log file. I'll add that to the list of files to upload as artifacts...

```

PR #570, opened 2022-03-29
Cleanup fds

plajjan:
Add a simple test case for light stressing of our TCP server stack. It now connects 2000 times (though sequentially) to a test server. This doesn't work unless we clean up fds since the default open file descriptor limit is 1024.

We now clean up fds when a client disconnects!

Our MAX_FD is now 1024, to align with the default fd limit of 1024. This does need to be improved further though, since a user can increase the fd limit beyond 1024 and we will still crash.

### DIFF
--- a/builtin/env.c
+++ b/builtin/env.c
@@ -1362,20 +1362,22 @@ void *$eventloop(void *arg) {
                 exit(-1);
             }
             EVENT_del_read(fd);
-        }
-        switch (fd_data[fd].kind) {
+            $init_FileDescriptorData(fd);
+            close(fd);
+        } else {
+            switch (fd_data[fd].kind) {
             case connecthandler:
                 if (EVENT_is_read(&kev)) {              // we are a listener and someone tries to connect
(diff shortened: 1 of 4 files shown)


builtin/env.c
```diff
@@ -1362,20 +1362,22 @@ void *$eventloop(void *arg) {
                 exit(-1);
             }
             EVENT_del_read(fd);
-        }
-        switch (fd_data[fd].kind) {
+            $init_FileDescriptorData(fd);
```
plajjan, line 1365: @nordlander I'm thinking about this in particular. close(fd) seems fairly safe / right thing to do, but what about our fd_data, how do we say that a particular fd slot is unused?

nordlander: Setting `fd_data[fd].kind = nohandler` is the proper way to indicate that the slot is unused. But we should also null out the closure pointer fields for the benefit of our future GC (that's `chandler`, `rhandler`, `errhandler` and `conn`). What we do with the remaining fields is irrelevant, as far as I can see.

PR #588, opened 2022-04-19
Use log.c lib in ActonDB

### LINKED ISSUE #581: Use logging library in DDB server & RTS
Rather than print everything to stdout, we should use some logging library. We want:
- ability to write to file but also to stdout as today
- timestamps, with millisecond or preferably microsecond precision so it is easy to correlate between DB and client what was going on
- small footprint, simple compile, simple to use
- quite fast
- thread safe

I don't think picking a library needs to be that big of a deal. We are aiming for output in a human format, so if the format changes a bit, that's not a big deal and how we call the log library in code is easy to change, like we can probably regexp replace if we ever switch from one library to another. Thus, we don't need to consider all and every feature we might ever need around logging, we can just focus on what we need right now and if our requirements every change, we can just switch logging library.

plajjan:
Rather than printf() log message, we now use the same log.c library that
is used in RTS, though with a slightly different set of modifications.

Log output is now sent to stderr rather than to stdout as previously.

Not all printf() statements have been replaced but most have. I've
mapped most log lines to log_debug() but have lowered the level of some
messages to log_info(). Rather than scrub all this in one go, I figured
we could over time adjust the levels as we see fit.

We support writing the log to a file with `--log-file=foobar.log`. If
file logging is enabled, no log output is written to stderr.

It might seem weird to have two log libraries and in particular with
different set of modifications but that was quite deliberate, since the
use in the DB and RTS are quite different. For example, RTS is threaded
while the DB is not (although it could become). The RTS log.c prints the
worker thread id, so it's possible to track what log lines belong to
what worker thread. The DB doesn't have worker threads, so such a change
would otherwise have to be made conditional. Keeping them separate is
thus, for the time being, simpler. There might be more changes that we
want to apply. If we see the two log libs being rather stable over time
and can find a way to combine, that would be fine too.

Example output:
```
09:12:40.680991 INFO  backend/actondb.c   : 2235: Using args: portno=32000, gportno=34000, verbosity=1, seeds:
09:12:40.681060 DEBUG backend/actondb.c   :  189: Create ACTORS_TABLE - OK (0)
09:12:40.681074 DEBUG backend/actondb.c   :  193: Create MSGS_TABLE - OK (0)
09:12:40.681080 INFO  backend/actondb.c   : 2265: State schema successfully created
09:12:40.681087 INFO  backend/queue.c     :   68: Queue table 2 created
09:12:40.681093 DEBUG backend/actondb.c   :  206: Create MSG_QUEUE table - OK (0)
09:12:40.681098 INFO  backend/actondb.c   : 2272: Queue schema successfully created
09:12:40.681170 INFO  backend/actondb.c   : 2313: Started [0.0.0.0:32000, 127.0.0.1:34000], my_lc = VC(99536:0)
09:12:44.400878 INFO  backend/actondb.c   : 2563: SERVER: accepted connection from client: localhost (127.0.0.1:51288)
09:12:44.401147 DEBUG backend/queue.c     :  796: BACKEND: Queue 2/0 created
09:12:44.401233 DEBUG backend/queue.c     :  651: BACKEND: Subscriber 0/0/0 subscribed queue 2/0 with callback (nil)
09:12:44.401299 DEBUG backend/queue.c     :  796: BACKEND: Queue 2/-11 created
09:12:44.401353 DEBUG backend/queue.c     :  651: BACKEND: Subscriber 0/0/-11 subscribed queue 2/-11 with callback (nil)
09:12:44.401660 DEBUG backend/queue.c     :  796: BACKEND: Queue 2/-12 created
09:12:44.401715 DEBUG backend/queue.c     :  651: BACKEND: Subscriber 0/0/-12 subscribed queue 2/-12 with callback (nil)
09:12:44.401760 DEBUG backend/queue.c     :  202: BACKEND: Inserted queue entry 0 in queue 2/-12, status=0
09:12:44.401772 DEBUG backend/queue.c     :  272: BACKEND: Notified remote subscriber -12
09:12:44.402747 DEBUG backend/queue.c     :  422: BACKEND: Subscriber -12 read 1 queue entries, new_read_head=0
09:12:44.402902 DEBUG backend/queue.c     :  202: BACKEND: Inserted queue entry 0 in queue 2/0, status=0
09:12:44.402920 DEBUG backend/queue.c     :  272: BACKEND: Notified remote subscriber 0
09:12:45.402799 DEBUG backend/queue.c     :  422: BACKEND: Subscriber 0 read 1 queue entries, new_read_head=0
09:12:45.403267 DEBUG backend/queue.c     :  202: BACKEND: Inserted queue entry 1 in queue 2/-12, status=0
09:12:45.403290 DEBUG backend/queue.c     :  272: BACKEND: Notified remote subscriber -12
09:12:45.403527 DEBUG backend/queue.c     :  202: BACKEND: Inserted queue entry 0 in queue 2/-11, status=0
09:12:45.403541 DEBUG backend/queue.c     :  272: BACKEND: Notified remote subscriber -11
09:12:45.405014 INFO  backend/actondb.c   :  839: Host disconnected, ip 127.0.0.1, port 51288, old_status=0, closing fd 5
```

Fixes #581.

Fixes #588.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ backend/actondb: backend/actondb.c lib/libActonDB.a
 		$(LDLIBS)
 
 backend/%.o: backend/%.c
-	$(CC) -g -o$@ $< -c $(CFLAGS)
+	$(CC) -DLOG_USE_COLOR -g -o$@ $< -c $(CFLAGS)
 
 backend/failure_detector/%.o: backend/failure_detector/%.c
 	$(CC) -g -o$@ $< -c $(CFLAGS)
@@ -239,7 +239,7 @@ DB_OFILES += backend/db.o backend/queue.o backend/skiplist.o backend/txn_state.o
 DBCLIENT_OFILES += backend/client_api.o rts/empty.o
 REMOTE_OFILES += backend/failure_detector/db_messages.pb-c.o backend/failure_detector/cells.o backend/failure_detector/db_queries.o backend/failure_detector/fd.o
 VC_OFILES += backend/failure_detector/vector_clock.o
-BACKEND_OFILES=$(COMM_OFILES) $(DB_OFILES) $(DBCLIENT_OFILES) $(REMOTE_OFILES) $(VC_OFILES) deps/netstring_rel.o deps/yyjson.o
+BACKEND_OFILES=$(COMM_OFILES) $(DB_OFILES) $(DBCLIENT_OFILES) $(REMOTE_OFILES) $(VC_OFILES) backend/log.o deps/netstring_rel.o deps/yyjson.o
 OFILES += $(BACKEND_OFILES)
 lib/libActonDB.a: $(BACKEND_OFILES)
 	ar rcs $@ $^
--- a/backend/actondb.c
+++ b/backend/actondb.c
@@ -2192,7 +2192,7 @@ int main(int argc, char **argv) {
   int ret = 0;
   char msg_buf[1024];
   int verbosity = SERVER_VERBOSITY;
-  FILE *logf;
+  FILE *logf = NULL;
 
   pid = getpid();
 
--- a/backend/actondb.c
+++ b/backend/actondb.c
@@ -2064,6 +2064,7 @@ typedef struct argp_arguments
   int no_seeds;
   char * local_iface;
   char *mon_socket_path;
+  char *log_file_path;
 } argp_arguments;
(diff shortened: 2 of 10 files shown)


plajjan:
@aagapi PTAL!

plajjan:
Talked about in meeting, all happy. Merging.

PR #598, opened 2022-04-22
Make actonc projects dev/rel aware

plajjan:
libActonProject.a is now called libActonProject_dev.a or _rel.a
condition on the --dev flag to actonc, just like we already do for
libActon. We change both which library we use when compiling as well as
the library we produce.


plajjan:
Superseded by #599. Closing.

PR #605, opened 2022-04-25
Add str methods

nordlander:


plajjan:
Please do rebase rather than merging in the main branch, the git history looks so weird now.

For example:
```
git checkout main
git pull origin main
git checkout add-str-methods
git rebase main
git push --force origin add-str-methods
```

PR #618, opened 2022-04-28
Implement build command for actonc

### LINKED ISSUE #594: Add `build` command to actonc
In an Acton project it should build all source files!

Something like
- find all source files
- figure out dependencies between source files and sort
- compile each file in dependency sorted order

For building binary executables (currently manualled via --root arg) we shall inspect the source file and if there is a actor definition named `main`, we will automatically build a binary! If there are multiple source files with a `main` actor, we will build multiple binaries!

plajjan:
Running `actonc build` will find all .act files in the project source
directory and compile them! There's a big hole in this as we cannot
build any binaries! Binaries are built by specifying the --root actor
but since we can't send in arguments per module, that won't work for a
project build. We can still build those by specifying the file though,
like `actonc src/foo.act`.

Fixes #594.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,20 +181,6 @@ deps/yyjson_rel.o: deps/yyjson.c
 
 # Building the builtin, rts and stdlib is a little tricky as we have to be
 # careful about order. First comes the __builtin__.act file,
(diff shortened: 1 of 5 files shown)


plajjan:
This is a wasteful approach, as outlined in the comments of #619, so going to close in preference of #623.

PR #619, opened 2022-04-28
actonc build alternative

plajjan:
Outline why first approach is wasteful and what an alternative could look like...


plajjan:
This PR outlines what's bad of #618 and through comments describes what we could do.. some of those ideas are now implemented in #623, so going to close this. No point in keeping a PR for docs.

Issue #358, opened 2021-11-10
Default to using `main` as root actor


plajjan:
If an acton source file defines an actor named `main`, we should detect that and use it as the root actor, thus automatically and implicitly building a binary executable!

plajjan:
Our current default is an empty string (`""`). I tried just changing it to `root` which then fails when we try to compile `__builtin__.act`. I should have seen that coming! All libraries etc won't have a `root` actor.

We already have the concept of being in a project or not. If we are in a project, I think we can expect the project config file or the location of a source file to indicate whether a binary should be built or not. If we are not in a project, a.k.a. in script mode, then we should always build a binary, right? Thus, setting a default actor value like `root` makes sense, no?

nordlander:
But isn't the point having a default available in the cases where we quickly want to build a binary? That is, when we're outside a project. I think it's correct to assume we want to build a binary in those cases, and a default root actor `main` sounds better in my ears than `root`.

But apart from this I believe we'll also want to preserve the ability to run the compiler on a particular file, just to check its static correctness if nothing else. So if the `--root` option turns into a default behavior we should perhaps add a "check-only" option to cover the special case. Why not call that option `--check`, with `-c` as a short-hand?

plajjan:
Blargh, I meant to suggest we default to `main` as the root actor, not the value `root`... :) 

So I think we agree. When **not** in a project, we want to default to `main`.

Yeah, we can do a `--check` option, or I guess you could just do `--root ""` to get the current behaviour?

plajjan:
@nordlander is there a way of parsing a file and inspecting the AST to determine whether there is an actor named `main`... or I mean, I presume that is possible, but how?

plajjan:
We previously wrote in this issue that we should only do this outside of a project... I think that was simply because we focused on the simpler case outside of a project... but really, it doesn't make sense to restrict this, on the contrary, detecting a main actor and using it as root makes sense in a project too I think. It's like a sensible default, which we can then override in the project config.

sydow:
This is implemented by #726 (except for --check).

PR #628, opened 2022-05-03
--root now takes a qualified name

plajjan:
--root takes the name of the actor that shold be used as the root actor.
It used to be unqualified, like "main" and actonc would look in the
source file specified. Now that we support building multiple files in a
acton project, an unqualified name is ambiguous. Thus, by switching to
qualified name we get around this!

We still do attempt to build a qualified name if an unqualified name is
given and do this by the "old rules" so it behaves the same way as it
used to for single file compilation, for example:

  actonc examples/count.act --root main

Here 'main' will be qualified to 'count.main' and is thus equivalent to
giving the full name directly:

  actonc examples/count.act --root count.main

Using unqualified names for a project build is not deterministic and
should be avoided.

There are some simple test cases to check that both qualified and
unqualified names work in a project and outside.

Related to #358 and #594.

### DIFF
--- a/compiler/ActonCompiler.hs
+++ b/compiler/ActonCompiler.hs
@@ -36,6 +36,7 @@ import qualified InterfaceFiles
 import Control.Exception (throw,catch,displayException,IOException,ErrorCall)
 import Control.Monad
 import Options.Applicative
+import Data.List.Split
 import Data.Monoid ((<>))
 import Data.Graph
 import Data.Version (showVersion)
@@ -129,6 +130,9 @@ main = do
       else
         case (cmd args) of
             "build" -> do
+                iff (not (null $ (root args)) && (length $ splitOn "." (root args)) == 1) (
+                    errorWithoutStackTrace("Project build requires a qualified root actor name, like foo.main")
+                    System.Exit.exitFailure)
                 -- find all .act files in src/ directory, parse into tasks and
                 -- submit for compilation
                 curDir <- getCurrentDirectory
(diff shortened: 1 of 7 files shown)


compiler/ActonCompiler.hs
```diff
@@ -290,6 +291,8 @@ findPaths actFile args  = do execDir <- takeDirectory <$> System.Environment.get
                                     return $ (False, False, pre, drop 1 ds)
 
 data CompileTask        = ActonTask  {name :: A.ModName, src :: String, atree:: A.Module} deriving (Show)
+-- TODO: why is this needed? why can't we use a A.GName?
+data RootActor          = RootActor {modname :: A.ModName, actname :: A.Name}
```
plajjan, line 295: @nordlander so this code works but it seems redundant to have to define a new record here. I want a record so that I easily can map over a list of such records and still pass all the necessary things.

I have a GName and later in buildExecutable we need a QName. Those records have the same field names, can I just convert / case one to the other?

I tried to pass GName, then extract the mname & noq and then build a QName, but it complains about trying to use mname. Is the mname field selector associated with QName? :confused:

plajjan:
Add check to ensure we get qualified name when doing project build!

compiler/ActonCompiler.hs
plajjan: Discussed; replace record with one for BinTask or similar that also contains binFileName in addition to a A.GName

plajjan: Ok, done.

plajjan:
I've added a check that the root actor is qualified when we're doing a project build. Also added test cases for this. Really feel like restructuring the test suite as... not a Makefile. Parallel test execution etc would be nice.

PR #629, opened 2022-05-05
Remove exception handling when building binary

plajjan:
The exception handler 'handle' is written to be used when compiling a
module. When invoked, it will remove the resulting .ty file, in order
not to expose a broken .ty file. When building binaries, this is not a
possible error scenario, since the .ty file is produced earlier.

Not sure what kind of errors could really pop up during buildExecutable
but I think they deserve to be handled differently and meanwhile it's
better to just print raw exceptions than to have incorrect handling.


plajjan:
@nordlander what do you think? Is this better? Alternative is to use a separate exception handling function that's meant to be used for buildExecutable, or we can add an argument to `handle` to instruct it whether to remove the file or not?

PR #631, opened 2022-05-05
Add actonc tests using Tasty

plajjan:
We have been using a Makefile to structure our tests for quite some time
now. Certainly better than nothing, but there are certainly improvements
too.

This is the first step towards better testing for actonc. We are now
using the Haskell Tasty package for running tests. It offers lots of
features, like nice compact output, parallel test execution, rich HTML
output and lots of other things.

The first couple of tests implemented are ports of the actonc project
build tests from the test Makefile.


plajjan:
@nordlander / @sydow FYI, if we want to add tests for the compiler, I think this is now a good place! These current tests are based on invoking actonc on .act files or projects, but we might just as well import some Haskell Acton module and run tests on low level functions if we want to. The sky is the limit!

PR #645, opened 2022-06-03
Expose and render distributed acton membership information

### LINKED ISSUE #643: Expose DB topology via mon socket
The gossip protocol now distributes a complete topology of both DB servers and DB clients (the RTSes). We should expose that through the mon socket interface, perhaps both in the RTS & DB server, so it can be queried and we can get look at the topology!

plajjan:
actonmon now has a better view for rendering membership information for a distributed Acton system that utilizes rich to display a periodically updating table.

The membership information can be pulled from either ActonDB or an RTS. Both ActonDB and RTS have been extended with a list of actual members.

Fixes #643.


plajjan:
@aagapi PTAL

aagapi (commented):
Please also change 256 -> 262 in function update_listen_socket(), in the following line where the id gets updated:

-    snprintf((char *) &rs->id, 256, "%s:%d", hostname, portno);
+    snprintf((char *) &rs->id, 262, "%s:%d", hostname, portno);

PR #659, opened 2022-06-08
Properly declare prerequisites for backend targets

plajjan:
All the Makefile targets related to the backend now have proper
prerequisite declarations so make should behave better. We should now
never need to use clean-backend or similar but any change should be
identified and rebuilt accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,17 +91,41 @@ backend/actondb: backend/actondb.c lib/libActonDB.a
 		-lActonDB \
 		$(LDLIBS)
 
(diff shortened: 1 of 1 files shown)


plajjan:
@aagapi ok, I think I've covered all the backend targets now with proper dependencies. Please only use `make -j` when you are working on backend stuff and if there are ever build errors due to incorrect dependencies, e.g. we are not rebuilding files despite them having changed, then please let me know so we can fix it! :)

PR #663, opened 2022-06-09
changed TCP content to be bytes instead of strings

sydow:
This is for you to review; there is one problem that should be checked before we merge:

The file examples/client.act does not compile:

acton> actonc examples/client.act --root client
Compiling client.act for release
actonc: No match in record selector rtail
acton> 

I believe that this has very little to do with this change (but it compiled before), but this is for Johan to judge.

The only changes are that in Connection, method .write takes a bytes parameter instead of a str. The same for the first callback parameter in  on_receive.

Applications like server/client in examples and ddb_test_server_client/server  in test/rts need to do some calls to encode(() and decode().

I  believe that the test suite passes, but there is a lot of output which I don't understand completely...

### DIFF
--- a/builtin/env.c
+++ b/builtin/env.c
@@ -438,14 +438,14 @@ $l$7lambda $l$7lambda$new($Env p$1, $str p$2) {
     return $tmp;
 }
 struct $l$7lambda$class $l$7lambda$methods;
-$NoneType $l$8lambda$__init__ ($l$8lambda p$self, $Connection __self__, $str s) {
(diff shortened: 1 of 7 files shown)


plajjan:
I'm an expert in rtail workarounds ;)  I'll take a look and see if we can get that to compile...

nordlander:
It’s a bug! Found it in the CPS pass, will submit a fix soon.

— Johan

9 juni 2022 kl. 20:32 skrev Kristian Larsson ***@***.***>:

﻿

I'm an expert in rtail workarounds ;) I'll take a look and see if we can get that to compile...

PR #664, opened 2022-06-09
Fix cps def in branch

nordlander:


nordlander:
Fixes issue detected during PR #663.

PR #671, opened 2022-06-10
Fix hex parse

sydow:
Implements splitting of strings when needed after two hex digits in hex escape sequences.


sydow:
I tried to fix this, but the linux test fails. In particular, the new examples/StringsAndBytes.act fails. I cannot see the exact reasons for failure in the Details, so I need help.

plajjan:
Interesting; not quite sure why it fails and yeah, there's no printout of the error itself. I tried to add that but don't understand how it's supposed to work with this Tasty library. For successful tests we want to suppress output; only when a test fails do we want to print some stuff but I don't see how.

Anyway, I checked out the branch on my machine and after rebasing on main it runs just fine. I (force) pushed the rebased version... let's see what CI thinks of that.

plajjan:
@sydow in #673 I implemented so that the test suite now prints output on test failures. I have rebased this branch so we can now see the failure, see https://github.com/actonlang/acton/runs/6830353409?check_suite_focus=true here included for convenience:

```
 ERROR: actonc return code (ExitFailure 1) not as expected (ExitSuccess)
STDOUT:
STDERR:
actonc: /__w/acton/acton/examples/StringsAndBytes.act: hGetContents: invalid argument (invalid byte sequence)
```

I don't get it at all. Could this depend on things like locale which might be different in CI than from my local Linux machine?

sydow:
I have now reduced examples/StringsAndBytes.act so that it passes also the Linux tests. The hex escape fix that was the aim of this whole push request remains but all tests of en- and decoding, switching betwwen non-trivial unicode  strings and their bytes representations are removed. I will now merge anyhow, and think of how to reintroduce more tests next week...

PR #676, opened 2022-06-11
Restructure regressions tests to use Tasty

plajjan:
Our regressions tests have now been moved over to be run by Tasty which brings nicer printout and concurrent testing. With some other changes we have reduced the time it takes to test examples, regressions etc to under 1 second on my 16-core AMD 5950X machine.

Tests are now better organized. We automatically discover tests in:
- examples/
- test/regression, where tests are expected to compile and run successfully (exit 0)
- test/regression_build, where tests are expected to fail compilation
- test/regression_segfault, where tests are expected to compile but segfault at run time
- test/regression_run, where tests are expected to compile but fail at run time (non-0 exit code)

For example, if we fix a test in test/regression_build, we move the .act file over into test/regression since it is now expected to succeed. Just like before, currently failing tests are marked as such and if they start succeeding but are still "marked" (implied by their location) the test will fail. A test that has failed but is now succeeding should be moved!


plajjan:
Looking at some past CI runs I see tests on Linux take 30, 31, 34, 35, 36 seconds. In this pipeline it took 22 so definitely some improvement. Out of this, the tasty tests are less than 3 seconds. Majority of time is still spent in the Makefile driven tests. I believe some of the DDB tests take the longest.

The number of threads used to run tests is the same as the number of CPU cores in the machine and I think the CI machines only have 2 core, so not a lot of parallelism going on. Could perhaps tune this as well. Doesn't make a world of difference right now, but as we add even more and longer tests to run with Tasty, it could make a difference.

PR #677, opened 2022-06-11
Test builtins

### LINKED ISSUE #683: "hGetContents: invalid argument (invalid byte sequence)"
We are seeing this error in CI on Linux:
```
... hGetContents: invalid argument (invalid byte sequence)
```

I think it is for source files that contain unicode characters.

There's only one hGetContents in the Acton code base (https://github.com/actonlang/acton/blob/main/compiler/InterfaceFiles.hs#L31) and I tried to enforce the encoding on that file handle (https://github.com/actonlang/acton/pull/677/commits/6de2b92bb4bfd4c24042ad68948ba4df974a9c66) but that didn't help.

I've now been able to reproduce the issue locally using a container and I've now even removed the hGetContents from my version (clearly breaking actonc...) but I still get the same error, so there must be some other hGetContent... in a library we use?

plajjan:
Also fixes #683.


plajjan:
@sydow I did a fix, enforcing utf-8 encoding on source files, so your test stuff, that I relocated into a test in test/builtins_auto/strings_and_bytes.act now seems to work! :)

PR #685, opened 2022-06-12
added method __repr__ to class value and subclasses

sydow:
Method __repr__() added to class value and subclasses (i.e. all builtin types). Main reason is to get better printing of contasiner types.


sydow:
Oops! the numpy test now fails. I had forgotten to add repr to classes in numpy and math, but there is still a problem, which I do not understand.
test_numpy.act segfaults. If I compile with --dev to be able to debug with llvm, linking fails and complains about similar things as I experienced yesterday with numpy in general. But I have not commited any changes to Makefiles or similar so I cannot understand why this test would fail like this. Have to give up for now.

PR #688, opened 2022-06-13
Add stdlib tests for math & numpy modules

plajjan:
Only doing the very basic testing of importing these libs separately and
then in combination (which causes problems).


plajjan:
This is superseded by #692 which adds the stdlib tests for numpy & math and #693 for fixing the linking problem with duplicate symbols. Closing.

PR #698, opened 2022-06-15
Add new libuv based IO thread to RTS

### LINKED ISSUE #691: Replace our IO eventloop with libuv
We want to modularize our IO stuff, like having an acton `net` module for network related things. This entails moving out various IO related functions from env.c so that they can be accessed from other acton modules, and by extension, other C files.

Rather than actually moving out some stuff from env.c, so that it can be imported, as suggested in #555, we have pondered (#561) and decided that we should migrate our IO stuff to use libuv. Let's a go!

- add new thread for libuv IO, same affinity as current eventloop
- write something simple, like a client TCP connection (`net.connect`) using libuv, see #550 
- alternatively, use libuv for something new, like a netconf module based on libnetconf2

plajjan:
This adds a new IO thread based on libuv. It doesn't really do anything besides existing but lays the foundation for using libuv in the RTS, env and modules. While we have yet to properly use it, I think this is correct, like linking is done as it should etc.

libuv header files are in dist/include. Previously we've had header files that are required at actonc run time in dist/builtin and dist/rts but I think it would be good to collect them in a single place so rather than create a uv specific directory, I added dist/include and endgame might be to move rts.h and builtin/*.h in there!?

libuv_.a is placed in dist/lib. We do not compile it differently for rel/dev profile (as of yet). It is linked in at actonc run time. I think this is the only way, for example; we could try to link io.o with libuv at compile time but as we would also need to link any module interacting with libuv with libuv, we would end up with duplicate symbols. Thus I think it's better we ship the libuv.a archive as part of the Acton system.

Fixes #691

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,7 @@ jobs:
             ~/.stack
           key: ${{ matrix.os }}-${{ secrets.CACHE_VERSION }}
       - name: "Install build prerequisites"
-        run: brew install argp-standalone haskell-stack protobuf-c
+        run: brew install argp-standalone haskell-stack libuv protobuf-c
       - name: "Build Acton"
         run: make -j2 -C ${{ github.workspace }} BUILD_RELEASE=${{ env.BUILD_RELEASE }}
       - name: "Build a release"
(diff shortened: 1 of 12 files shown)


plajjan:
I've verified this is working by implementing DNS lookups in an acton module. I won't merge that code just yet since it's wrong; it runs synchronously and blocking which is the opposite of what we want. Nonetheless, it proves the libuv loop works as it should so it validates the code in this MR, thus I'll go ahead and merge this. There are some build issues currently, which I'm working to fix.

PR #702, opened 2022-06-16
added Eq instances to list and dict

sydow:


sydow:
This request adds Eq instances for list[A] if A supports Eq and for dict[A,B] if B does so (A always does). Ord is left for another time and Eq for tuples to an uncertain future.

After spending too much time on trying to understand why I got type errors in my small test example involving map, I finally sat down with the "Cannot solve ..." message and realized that we hadn't implemented Eq for bool. 

Then I had a similar experience with a dict example until I realized that we hadn't implemented Hashable for bool. So now both these are also implemented.

PR #704, opened 2022-06-16
Use system ghc

plajjan:
This is an attempt at speeding up CI by using the system installed GHC...


plajjan:
Nah, this won't yield much. FWIW, this option does not force the use of system GHC, it only enables it. The version requirements based on the stack release, like 18.12-lts must align, so if it wants GHC 8.8 and 8.8 is installed on the system it can use it. But if the system has a different version of ghc, stack will still pull it down.

Anyway, I don't think this is a good way forward so closing.

PR #710, opened 2022-06-16
Allowed docstrings in NotImplemented functions.

### LINKED ISSUE #709: Docstring prevents ext-C NotImplemented functionality from working
#706 looks really cool, good job @nordlander !   but...

```python
def foo() -> int:
    """My foo function
    """
    NotImplemented
```

inserting a docstring (we still want nice documentation!) prevents this thing from working. Can we fix that? :)

nordlander:
Fixes #709


plajjan:
woohoo

PR #722, opened 2022-06-19
Cache build-debs job

plajjan:
This should speed up the build-debs job by a fair bit. We only do it
when not running on the main branch, since then we want to be absolutely
cerain we're doing a fresh build.


plajjan:
This doesn't seem to work. build-debs is taking just as long as before, probably due to some flag explicitly instructing the build not to use cached data.

It's not a biggie, this job doesn't take *thaat* long anyway, so closing this PR...

PR #723, opened 2022-06-19
Include libuuid in acton system distribution

### LINKED ISSUE #76: Ensure static linking of actonc output
I want to make sure we statically link the final binary that actonc produces. Check and fix if necessary!

plajjan:
And remove it as a run time dependency on actonc.

Fixes #76


plajjan:
Hmm, so there is no statically compiled .a file shipped with libuuid... and MacOs comes with libuuid as a builtin. I guess it's fairly likely many systems already have uuid, so let's skip trying to include it. It can remain a run time dep.

Issue #728, opened 2022-06-22
Root actor discovery fails for partial recompilation


plajjan:
Rebuilding a project that's been built before where only some of the source files have been updated now fails. I believe it's because we don't read in the .ty files and so have no content loaded for a module, yet go looking for actors in it..

plajjan:
Fixed by #729. Closing.

PR #729, opened 2022-06-22
Add failing test case for rebuilding a project

plajjan:
There's a problem with the new root actor discovery mechanism. When a
project is rebuilt, actonc crashes with:

    kll@ThinkYoga:~/acton/test/compiler/rebuild$ ../../../dist/bin/actonc build
    Compiling rebuild.act for release
    Building executable /home/kll/acton/test/compiler/rebuild/out/rel/bin/rebuild
    actonc: Maybe.fromJust: Nothing
    CallStack (from HasCallStack):
    error, called at libraries/base/Data/Maybe.hs:148:21 in base:Data.Maybe
    fromJust, called at ActonCompiler.hs:570:43 in main:Main

Which I'm guessing has to do with that we do not read in all modules
from .ty files unless we need to. When building from scratch, all
modules are read in, since we are compiling them all, but when some are
already built, we skip them unless they are dependencies of other
modules.

Test case for #728

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -361,7 +361,7 @@ test-builtins:
 	cd compiler && stack test --ta '-p "Builtins"'
 
 test-compiler:
-	cd compiler && stack test --ta '-p "actonc"'
+	cd compiler && stack test --ta '-p "compiler"'
 
 test-db:
 	cd compiler && stack test --ta '-p "DB"'
--- a/compiler/Acton/CodeGen.hs
+++ b/compiler/Acton/CodeGen.hs
@@ -15,6 +15,7 @@
 module Acton.CodeGen where
 
 import qualified Data.Set
+import qualified Data.List
 import qualified Acton.Env
 import Utils
 import Pretty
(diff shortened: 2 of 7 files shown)


plajjan:
So my test case that is supposed to be failing, and which is always failing on my laptop... actually passes CI testing. I'm not sure how that can be... very weird.

Hopefully it fails on your computer @sydow, so you can reproduce this problem and can work on a fix.

plajjan:
The failures on Mac are also not related I think. Probably something intermittent.

sydow:
It fails on my laptop :-)

plajjan:
Ah yes, I believe hierarchical was natural but for libs we end up adding it all in one .a archive, which doesn't support directories, so it's preferable if the filenames are unique on their own. IIRC, it doesn't really matter, like we can have two `foo.o` in the .a, since it is just searched for symbols anyway. But it is easier for debugging etc if the name reflects the hierarchy. `$` was chosen because that is what is used internally in C. I agree that `.` seems like a better choice.

PR #751, opened 2022-06-30
Restructure DNS lookup functions as DNS actor

### LINKED ISSUE #749: Ensure ioloop is started before running programs
I believe we have a race now between the ioloop starting up and actor code registering callbacks. Programs that make use of libuv backed modules are sort of flaky and I believe this is the cause.

We should make sure the ioloop has started before we do anything else!

### LINKED ISSUE #750: Rework net lookup functions
Cannot be functions as that means we cannot call callbacks since that requires CPS and module functions are not...

I think the only way to do this is to first create  a DNS actor and place the lookup methods on it.

plajjan:
I think this is a better fit. Also brings the added benefit of being able to call impure functions (which wasn't the case before due to wrong type signatures).

The test program is now correct (thanks to being able to use impure callbacks) and runs much faster. It used to take 1 second due to a after 1 but now runs in millieseconds.

We now await the ioloop to start up before proceeding with starting worker threads. This has eliminated a race which lead to deadlocks.

Fixes #750.

Fixes #749.

### DIFF
--- a/rts/io.c
+++ b/rts/io.c
@@ -1,6 +1,7 @@
 #include "io.h"
 
 #include <uv.h>
+#include <unistd.h>
 
 #include "log.h"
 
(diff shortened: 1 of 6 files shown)


plajjan:
I still think it feels like overkill with a DNS actor but it does work well today. Maybe we can revisit this, trying to see if there is a way to model this as just functions in a module, but I suspect we should at the very least await the new deactorizer.

PR #754, opened 2022-07-01
Fix printf format

plajjan:


plajjan:
Bah, bummer. This kind of stuff sucks. My mac complains and thus this PR but then all the other platforms cc complain... platform dependent types suck. Closing this.

PR #766, opened 2022-08-01
Rewrite C lib guide for new style

### LINKED ISSUE #98: C lib wrapping: explain when something is sync vs async!
I believe the following is true:
- modules with free functions (not in a class or actor) will always be synchronous
- methods on classes are also always synchronous, since they are really instantiated within an actor and everything within an actor is always synchronous
- modules that implement actors will default to the actor behaviour, which is that when an actor method is called without binding the return value, like `foo.bar()`, then it is called asynchronously, whereas if the return value is assigned, like `a = foo.bar()` then the actor method call is synchronous

@nordlander is the above true? Also realized while writing this, that this is not specific to when wrapping C libraries - this is the generic Acton sync behavior, but I guess it is somehow extra relevant to point out when implementing a library.

I know it is possible to explicitly do something async or sync by adding `await` / `async` keywords at the call site. Is it possible to change the sync/async behaviour by specifying something at the function definition? No, right? I think I've asked this before but have forgotten.

Should add this to the guide!

plajjan:
Fixes #98

### DIFF
--- a/docs/wrapping_c_libraries.md
+++ b/docs/wrapping_c_libraries.md
@@ -1,101 +1,73 @@
 # Wrapping C libraries in Acton
 
-This document attempts to provide a brief introduction to integration of C libraries in Acton.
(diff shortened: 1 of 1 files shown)


plajjan:
@garberg FYI!!

PR #774, opened 2022-08-02
Test run actonc on debian buster (10)

plajjan:


plajjan:
This won't work. The acton package is built on debian 11 and thus with a newer version of glibc than debian 10 has. The glibc dependency is automatically generated to match that of the platform on which it was built, so we can't just hack around this (though that would also be wrong).

So, to support running on debian 10, we need to build the .deb packages on something older. We can't actually build on buster due to a gcc bug, but perhaps we can work around that, either by upgrading the gcc version or by building on Ubuntu, maybe 20.04?

plajjan:
There are no backports of newer gcc versions to debian:10, so that's a no go. Ubuntu 20.04 uses glibc 2.31 while debian:10 is on glibc 2.28, so building on ubuntu 20.04 won't do anything either. I'm giving up on this. Closing.

Issue #795, opened 2022-08-09
Complete process module


plajjan:
- [x] pid
- [x] kill
- [x] write
- [x] close stdin
- [x] pass env
- [x] pass working directory
- [x] use on_error cb for errors instead of raising exceptions
- [x] malloc args & env
- [x] rename wdir to workdir

plajjan:
Think we're pretty much done here!? Closing...

PR #803, opened 2022-08-12
Use on_exit callback for errors

plajjan:
Rather than raising an exception when we encounter an error in spawning
the process, we call the on_exit callback. It now receives a new
argument which is the error message. If there is no error and the
process exited for some other reason, the error variable is None.

The reasoning for this approach is that it's likely we need to write
less code and have a more consistent control flow when using callbacks.
If a process exits abnormally, we would write a on_exit callback to
handle this and perhaps restart it. If a process doesn't start / spawn
properly, that is sort of in the same category of events and should
probably lead to a restart.

Part of #795.

### DIFF
--- a/stdlib/src/process.act
+++ b/stdlib/src/process.act
@@ -5,7 +5,7 @@ class ProcessAuth():
     def __init__(self, auth: WorldAuth):
         pass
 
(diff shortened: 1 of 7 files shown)


plajjan:
Hmm, changed my mind. Better to just add another callback called `on_error`. In the end, we want to have a handler object with multiple method and for that I think it's cleaner to just have an error event that we can react to... we don't support that sort of interface yet, so callbacks will have to do. Having on_error callback method means we can easily move that to be the error handler later, so like it will be 1:1 method mapping, thus it is better then reusing on_exit since that would require more refactoring later on.

PR #822, opened 2022-08-23
Removed the use of ... in $function and $Cont signatures.

nordlander:


plajjan:
This requires changes to the compiler too, right? In the generated code we still get `$function` which ought to be some function with arity, right? It can be fixed in hand-written modules by casting but in our generated code it's still a problem, no?

plajjan:
Oh wait, I see, you don't want to complete the whole arity thing here, just prepare the main branch with arity and do the compiler fixes in the `deactorizer` branch, right?

PR #878, opened 2022-09-19
Use deploy key for pushing tag

### LINKED ISSUE #877: Fix CI release job
Use other token. We cannot use default github token. Using that token to create a tag will not cause other workflow to start based on tag.

Also same for Homebrew Acton repo!

plajjan:
The push used the defaults credentials provided when checking out the repository, which equates to github.token. For some reason, perhaps to avoid cyclic errors, the CI runner is not supposed to trigger itself and so using github.token for auth will not trigger any workflows. By switching to our own SSH deploy key for checking and pushing the repo we can trigger workflows!

Fixes #877

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,17 +5,22 @@ on:
 
 jobs:
   if_merged:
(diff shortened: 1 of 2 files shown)


plajjan:
Ugh, these things are to hard to write and debug. Hope I get this right...

PR #892, opened 2022-09-21
Add homebrew on include path

plajjan:
Homebrew on x86 places headers in /usr/local/include whereas on M1 it is placed in /opt/homebrew/include. gcc / clang includes /usr/local/include per default but for M1 we must explicitly add the include path.

We now also explicitly list the standard include paths so we can work with compilers, like zig, that do not seem to have any default include paths?


plajjan:
@sydow so this patch fixes the problem for me on my M2 computer, please check if it works for you!

sydow:
Yes, it fixes the problem for me, too.

PR #899, opened 2022-09-24
fixes issue 869

sydow:
The bug is caused by a dark corner in C macro expansion.  The macro $NEWTUPLE in tuple.h uses the tuple pasting operator ## in order to avoid a spurious "," in the call of __init__ for an empty tuple. This is a gcc extension which does not work for nested macro calls.

Fixed by introducing a special macro $NEWTUPLE0 for empty tuples and removing the ## operator in $NEWTUPLE.

Probably one should think through other macros in common.h which also use ## and would fail on nested calls.


sydow:
In the above comment, "tuple pasting" should have said "token pasting"...

sydow:
In the previous comment, "tuple pasting" should be  "token pasting"...

Issue #907, opened 2022-09-28
Timer queue recovery fails sometimes


aagapi:
When starting and recovering an actor that uses timers, the actor sometimes blocks after recovery. This can be verified with a single database server, and the basic counter actor count.act:

./count 1000 --rts-ddb-host localhost:32000 --rts-verbose --rts-ddb-replication 1 --rts-host localhost --rts-node-id 1

When repeatedly killing and restarting the counter actor, the blocking will eventually be reproduced.

Initial investigation:

I added extra logging in the RTS recovery code and noticed that:
1) Counter state is always correctly recovered from the DB upon restart
2) The timers in the timer queue are always correctly recovered from the DB upon restart
3) The RTS, upon recovery, sometimes (correctly) updates the timer deadline to 'now' if too much time has passed since the crash
4) The RTS, upon recovery, always inserts the timer message recovered from the DB in the local RTS timer queue.
5) Still, at times the RTS blocks after recovery, as if it does not handle the actor timer in the local queue

@plajjan mentioned during meeting the issue was possibly introduced by changes related to RTS libuv integration. Assigning to him.

Log with example failure:
 
### Outcome
```
11:10:44.160308 DEBUG rts/rts.c           :  938: RTS   :    # read msg from queue -12 returns 1, entries read: 1
11:10:44.160424 DEBUG rts/rts.c           :  938: RTS   :    # read msg from queue -12 returns 1, entries read: 0
11:10:44.160657 DEBUG rts/rts.c           :  938: RTS   :    # read msg from queue -11 returns 1, entries read: 0
11:10:44.160940 DEBUG rts/rts.c           :  938: RTS   :    # read msg from queue 0 returns 1, entries read: 1
11:10:44.160967 DEBUG rts/rts.c           : 1142: RTS   : # baseline (1664363443672819) < now (1664363444160693)
11:10:44.160992 DEBUG rts/rts.c           : 1146: RTS   : # Adding Msg -17 to the timerQ
11:10:44.161085 DEBUG rts/rts.c           :  938: RTS   :    # read msg from queue 0 returns 1, entries read: 0
11:10:44.161106 DEBUG rts/rts.c           : 1137: RTS   : # Failed to read msg from DB TIMER_QUEUE
11:10:44.161122 INFO  rts/rts.c           : 2248: RTS   : Actor state restored from DDB.

=> HERE the actor recovers correctly and does not block

4
5
[...]

PR #911, opened 2022-09-28
Change actondb default gossip port

plajjan:
I think it's a short-coming in the gossip protocol so we only gossip one of the ports, the normal port that clients connect to, so to find the gossip port we just take that port + 1. Thus, users are not really free to set the gossip port however they want. Maybe it shouldn't even be configurable given this constraint... however, changing the default port is at least a simple way of leading users to the right thing.

### DIFF
--- a/backend/actondb.c
+++ b/backend/actondb.c
@@ -83,7 +83,7 @@ void error(char *msg) {
 #define SERVER_VERBOSITY 1
 
 #define DEFAULT_DATA_PORT 32000
-#define DEFAULT_GOSSIP_PORT 34000
+#define DEFAULT_GOSSIP_PORT 32001
 
 #define RANDOM_NONCES
 int64_t requests = 0;


plajjan:
@aagapi FYI

aagapi:
> @aagapi FYI

Yes, this is the correct change to make for now, to allow server to be run with default args, as per the discussion thread in https://github.com/actonlang/acton/issues/907. Please go ahead and merge it.

plajjan:
Hell, maybe gossip port shouldn't even be configurable given that things will just not work if you change it...

aagapi:
> Hell, maybe gossip port shouldn't even be configurable given that things will just not work if you change it...

I think the correct solution is probably to allow both ports to be configurable, and advertise both ports in gossip messages, such that RTSs learn RPC ports of servers dynamically as servers join via gossip. This would allow most flexibility in case people have special constraints on what ports they can keep open for servers in their particular deployment.
At the moment, only gossip ports are gossiped between servers (and configured as seed contact points initially), and the RPC ports are derived by clients (RTSs) by convention on base of the gossip port, which is a bit more rigid and requires the user to be aware of the implicit port offset convention when deploying, which I don't really like.

PR #913, opened 2022-09-28
Compute gossip port as offset from RPC port

plajjan:
The gossip protocol only includes the RPC port and does not explicitly
say what port is used for gossip, so other parts of our system assumes
it is the RPC port +1. Thus, we change the actondb default to also be
based on RPC port +1.

The long term solution is to make the gossip port configurable and to
propagate that information in the gossip protocol. Meanwhile, since we
do not support any other port than +1, we make it impossible to
configure but leave most of the argument parsing code in so that it's
easy to change one day.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@
 ### Added
 - Bash completion is now part of the Debian packages & brew formula
 
+### Changed
+- actondb now uses a default value for gossip port of RPC port +1 [#913]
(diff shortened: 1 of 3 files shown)


plajjan:
@aagapi FYI I improved it a bit further by making the default value an offset from the RPC port rather than a fixed default, so now if you do `actondb -p 1234` we get a gossip port of `1235`. It is still possible to override but you get a stern warning...

PR #916, opened 2022-09-28
Only communicate with live nodes

### LINKED ISSUE #402: Remove garbage output from actondb
After we've encountered an error (like from #209) it appears we get garbage output from the database. Fixelifix!

### LINKED ISSUE #910: Garbage output from DB client?
### Code

Run any acton program towards the DDB. However, do not actually run the DDB and you'll get garbage output after the RTS fails to connect to the DB.

### Outcome
Note how I am not actually running an `actondb` which is why it fails to connect...
```
kll@Boxy:~/terastream/acton$ examples/count --rts-ddb-host=127.0.0.1 --rts-ddb-replication=1
get_remote_server: ERROR connecting to 127.0.0.1:32000
ERROR: Failed joining server 127.0.0.1:32000 (it looks down)!
9
:5
ÿÿ
localhost ÿÿ(ÿÿÿÿÿÿÿÿÿ0ÿÿÿÿÿÿÿÿÿ§¼àNo quorum (0/1 replies received)
")
                                                                                                       þÿÿÿÿÿÿÿ ŀа(No quorum (0/1 replies received)
*K
                                                                                                        ÿÿÿÿÿÿÿÿÿ ÿÿÿÿÿÿÿÿÿ(ÿÿÿÿÿÿÿÿÿ0ÿÿÿÿÿÿÿÿÿ8ÿÿÿÿÿÿÿÿÿJPüXNo quorum (0/1 replies received)
*0                                                                                                       (0ÿÿÿÿÿÿÿÿÿ8ÿÿÿÿÿÿÿÿÿJPݭÀòXNo quorum (0/1 replies received)
*T                                                                                                      õÿÿÿÿÿÿÿÿ
        ÿÿÿÿÿÿÿÿÿ ÿÿÿÿÿÿÿÿÿ(ÿÿÿÿÿÿÿÿÿ0ÿÿÿÿÿÿÿÿÿ8ÿÿÿÿÿÿÿÿÿJPàXNo quorum (0/1 replies received)
*B                                                                                                      õÿÿÿÿÿÿÿÿ (õÿÿÿÿÿÿÿÿ0ÿÿÿÿÿÿÿÿÿ8ÿÿÿÿÿÿÿÿÿJPñŀÆXNo quorum (0/1 replies received)    
```

### Expectation

I do expect an error but I don't know where all the garbage output is coming from, it should not be there!

plajjan:
The DB client interaction functions have unconditionally tried to communicate with all remote servers in our list. If a server is dead, it's socket is invalid and trying to communicate with it doesn't make any sense. We now only communicate with servers that we think are live.

This has been a long running issue and a rather annoying one where we've seen lots of garbage written to the screen when RTS fails to connect to a DB server. It happened because the sockfd gets rewritten to 0, which is stdin, and then the message meant for a DB server is written to our terminal. It's a binary protocol so it looks like a bunch of garble garble.

Now also using -1 for sockfd so we get an error if we try to write to it and tightened up so we only set NODE_LIVE after connection is successful.

Same changes also goes for the actondb server which has had similar problems with garbage output in the past.

Fixes #910.

Fixes #402.

### DIFF
--- a/backend/client_api.c
+++ b/backend/client_api.c
@@ -1129,11 +1129,7 @@ void error(char *msg) {
 
 int send_packet(void * buf, unsigned len, int sockfd)
 {
-	if(sockfd == 0)
-	{
-		log_error("ERROR: Attempted to write packet to fd 0!");
-		return -1;
-	}
+	assert(sockfd > 0);
     int n = write(sockfd, buf, len);
     if (n < 0)
     {
--- a/backend/client_api.c
+++ b/backend/client_api.c
@@ -1215,6 +1215,8 @@ int send_packet_wait_replies_async(void * out_buf, unsigned out_len, int64_t non
 	for(snode_t * server_node = HEAD(db->servers); server_node!=NULL; server_node=NEXT(server_node))
 	{
 		remote_server * rs = (remote_server *) server_node->value;
+		if (rs->status != NODE_LIVE)
+			continue;
 #if SYNC_SOCKET > 0
 		pthread_mutex_lock(rs->sockfd_lock);
 #endif
--- a/backend/client_api.c
+++ b/backend/client_api.c
@@ -179,9 +179,9 @@ int handle_socket_close(int * childfd)
 	printf("Host disconnected , ip %s , port %d \n" ,
 		  inet_ntoa(address.sin_addr) , ntohs(address.sin_port));
 
-	//Close the socket and mark as 0 in list for reuse
+	//Close the socket and mark as -1 in list for reuse
 	close(*childfd);
-	*childfd = 0;
+	*childfd = -1;
 
 	return 0;
 }
@@ -1129,6 +1129,11 @@ void error(char *msg) {
 
 int send_packet(void * buf, unsigned len, int sockfd)
 {
+	if(sockfd == 0)
+	{
+		log_error("ERROR: Attempted to write packet to fd 0!");
+		return -1;
+	}
     int n = write(sockfd, buf, len);
     if (n < 0)
     {
@@ -1215,7 +1220,7 @@ int send_packet_wait_replies_async(void * out_buf, unsigned out_len, int64_t non
 	for(snode_t * server_node = HEAD(db->servers); server_node!=NULL; server_node=NEXT(server_node))
 	{
 		remote_server * rs = (remote_server *) server_node->value;
-		if (rs->status != NODE_LIVE)
+		if (rs->status != NODE_LIVE || rs->sockfd <= 0)
 			continue;
 #if SYNC_SOCKET > 0
 		pthread_mutex_lock(rs->sockfd_lock);
--- a/backend/comm.c
+++ b/backend/comm.c
@@ -506,9 +506,7 @@ remote_server * get_remote_server(char *hostname, unsigned short portno,
 
         if(do_connect)
         {
-            rs->status = NODE_LIVE;
-
-        		rs->sockfd = socket(AF_INET, SOCK_STREAM, 0);
+			rs->sockfd = socket(AF_INET, SOCK_STREAM, 0);
 			if (rs->sockfd < 0)
 			{
 				fprintf(stderr, "ERROR opening socket!\n");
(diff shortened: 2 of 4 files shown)


plajjan:
@aagapi can you please have a look at this and merge if it looks good?

I'm using -1 as a value for invalid socket instead of 0, since 0 is actually a valid socket (stdin) and so we would attempt to write to it. This should act as a safeguard in case there are more similar situations. This goes both for server and client.

Second, I'm inspecting the node status in the DB client before trying to send data to it. We only send to live nodes.

plajjan:
Rebased to fix conflict...

aagapi:
Good find! I added a commit to this pull request with some small changes:

1) I am also setting the socket fd to -1 in the handle_socket_close() function, which gets called by the  comm loop when the other end closes the socket on us.
2) In the send_packet_wait_replies_async() function, I am also checking for rs->sockfd <= 0 in addition to NODE_LIVE, because we might be in a window between the time the server crashed / closed socket (marked at point 1) and the time the gossip view is learned by us.

Note that writing to sockfd 0 was indeed an issue that could happen in the client because of the lack of check in send_packet_wait_replies_async(), however it couldn't actually happen in the server, as there I am always checking for both node status, and enforcing rs->sockfd > 0 (so excluding 0) whenever servers communicate, for the reason mentioned at 2) above. E.g. https://github.com/actonlang/acton/blob/63d447274e398871f5322539227d909ebac88348/backend/actondb.c#L1491

I also changed back the setting of rs->status in get_remote_server(), as it was equivalent, but kept your change to sockfd = -1.

BTW, I also noticed a different issue when testing this, but I will make a separate PR and link it here, it is too big to piggyback to this PR.

Good find again!

backend/comm.c
```diff
@@ -506,7 +506,9 @@ remote_server * get_remote_server(char *hostname, unsigned short portno,
 
         if(do_connect)
         {
-			rs->sockfd = socket(AF_INET, SOCK_STREAM, 0);
+			rs->status = NODE_LIVE;
```
plajjan, line 509: I don't understand why you are bringing this back. I moved setting `rs->status = NODE_LIVE` down to after the connect, when we actually know if the connect succeeded or not. Meanwhile, `rs->status = NODE_UNKNOWN` as set on line 481. Isn't that more correct? We don't actually know if the server is live until after we connected, thus NODE_UNKNOWN is the status we should use?

backend/client_api.c
```diff
@@ -1129,6 +1129,11 @@ void error(char *msg) {
 
 int send_packet(void * buf, unsigned len, int sockfd)
 {
+	if(sockfd == 0)
```
plajjan, line 1132: Is this even necessary though?

I initially added an `assert(sockfd != 0);` here so we'd just exit in these situations rather than continue... but after changing sockfd to -1 everywhere I removed the assert since -1, unlike 0, is actually an invalid fd and the write() a few lines down will catch it with its error handling.

aagapi: Right, I was thinking to specifically catch sockfd value setting errors in the spirit of this PR, to specifically see if we update the fd properly everywhere, since the write() below could error for other reasons than the sock being invalid. I agree it's optional.

backend/comm.c
plajjan: Okay. I figured that as the comms is one thread and the RTS threads are running concurrently, we can access things in parallel and see these transitions as they happen and thus, marking the NODE_LIVE too early would have the RTS threads attempt to send data to this server (which isn't connected yet). But I don't have the full workings of this in my head so I'm probably have a mistake in my reasoning somewhere...

backend/client_api.c
plajjan: Okay, then I vote for an assert which I think is more explicit and since it's clearly indicative of coding error we should just exit and not continue with a bad state...

backend/comm.c
aagapi: The concurrency concern is valid, but by the way we update the local view in the client API, the active list of servers never gets updated with a new remote_server pointer until we have settled on whether the server is NODE_DEAD or NODE_LIVE, so we would never see a transitory state. Also, concurrent updates to that list are protected by db->gossip_lock (although at the moment updates only happen from the comm thread, so that's redundant at the moment).
The initial idea of the NODE_UNKNOWN state was not to handle access concurrency, but to support an intermediate state in inter-server gossip where nodes are learned about, but their liveness hasn't yet been verified by connection. However, I realized in practice that state is superfluous, NODE_DEAD will do just fine for that.

backend/client_api.c
aagapi: OK!

aagapi:
Ready to merge this?

plajjan:
@aagapi I turned that sockfd check into an assert instead, think it's better we catch it that way more explicitly. Enabling auto-merge when CI completes...

aagapi:
> @aagapi I turned that sockfd check into an assert instead, think it's better we catch it that way more explicitly. Enabling auto-merge when CI completes...

Great!

PR #929, opened 2022-10-03
Revamp remaining RTS DB tests

plajjan:
Rewrite the two remaining RTS DB test; the simple test app and the TCP client resume test in the style of the TCP server test, which I previously rewrote to use a new test orchestrator (written in Acton).

Pretty messy code, though it works and the tests execute very fast and print most relevant debug information if anything goes wrong.

I'm able to run many test iterations without problems!


plajjan:
I'm a little stumped. After merging the first revamped test for the TCP server, I have had several failed PR test rounds where the test times out. I've run thousands of iterations on my machine which have worked just fine. There are some segfaults but they are rare (<1% which is less than what we had before)... in this PR I've added some better log output but also extended the timeout. Are the test machines so damn slow? A test run completes in <0.4 seconds on my machine and the timeout was 2 seconds. It's now 5 seconds...

Issue #939, opened 2022-10-04
Handle DB errors / return codes of DB client API in RTS


plajjan:

plajjan:
It seems there are DB client API functions that do not return errors the way they should, for example https://github.com/actonlang/acton/blob/main/backend/client_api.c#L1635. Shouldn't that be `return NO_QUORUM_ERR;` ?

aagapi:
> It seems there are DB client API functions that do not return errors the way they should, for example https://github.com/actonlang/acton/blob/main/backend/client_api.c#L1635. Shouldn't that be `return NO_QUORUM_ERR;` ?

Fixed by https://github.com/actonlang/acton/pull/945. Thank you for noticing this. There were 4 functions in the API that do non-range search of DB rows in the client API did not return a status, but a pointer to the row, an NULL if this fails. Changed them to also return status. I also changed the DB unit tests, however these functions were not currently used in the Acton RTS.

plajjan:
@aagapi remote_new_txn also returns uuid of transaction rather than a status. Should it too be changed?

aagapi:
> @aagapi remote_new_txn also returns uuid of transaction rather than a status. Should it too be changed?

This is an interesting one. In principle that function call creates a new transaction state locally, including a new UUID txnid generated on the client side. While UUID collisions are *very very* unlikely, it then still checks for UUID collisions on a quorum of servers to make sure a previous txn doesn't exist with the same ID (this is more to future-proof the code in case we decide to use something else than UUIDs for txn_ids). Even in the very unlikely case a collision is detected on the servers, the function internally retries with a new generated UUID until success (no collision).
So the only reason this function would ever fail is actually for a lack of quorum (and again the check of collisions on the servers is probably superfluous in the first place if we assume no collisions from the UUID library). This is the reason I chose not to include a status and simply return NULL in the case of a lack of quorum (since the status can only ever be 0 or NO_QUORUM_ERR), to make calling the init function for a txn simpler / less tedious.

plajjan:
Great work @aagap!! 🎉

Closing this issue.

PR #943, opened 2022-10-04
Prefer symbol

plajjan:
This changes return code from -2 to -3 but I think that's what we mean!?

Part of #939


plajjan:
I wonder, does this mix up closing a client transaction vs a server transaction? are they -2 and -3?

aagapi:
> I wonder, does this mix up closing a client transaction vs a server transaction? are they -2 and -3?

No, it simply should have used the NO_SUCH_TXN return status (-3) as opposed to -2 (NO_SUCH_MSG_CALLBACK). Good find!

PR #944, opened 2022-10-05
Add no quorum error handling of all DB calls

plajjan:
All function calls to the distributed database now checks if we reach a quorum or are lacking one for each function called. We will retry individual calls indefinitely.

Tidied up a few tab indents as well.

Part of #939

### DIFF
--- a/rts/rts.c
+++ b/rts/rts.c
@@ -160,10 +160,10 @@ int sched_getaffinity(pid_t pid, size_t cpu_size, cpu_set_t *cpu_set)
 }
 
 kern_return_t thread_policy_set(
-					thread_t thread,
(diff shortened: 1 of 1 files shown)


aagapi:
I believe all changes in this PR are covered by the changes in https://github.com/actonlang/acton/pull/957 that we discussed, jointly with schema repairs on DB servers that re-join a quorum and other fixes.
@plajjan , please review and confirm if there is anything in https://github.com/actonlang/acton/pull/944 that is missing in https://github.com/actonlang/acton/pull/957

plajjan:
Agree this is covered so removing this PR and my branch.

PR #953, opened 2022-10-06
Add garbage collection

### LINKED ISSUE #952: Implement garbage collection using libgc
### Proposal

Use libgc for garbage collection!

### Motivation

Acton needs a GC!

### Alternatives

Naturally the long term solution is a custom GC for Acton that interoperates well with other parts of Acton providing the best solution for performance etc. In the meantime, any GC is better than no GC.

plajjan:
This uses libgc for garbage collection in the Acton RTS. libgc is a
garbage collector for C and C++ programs and by modifying our RTS and
generated Acton programs we can make use of it.

I've tested this primarily with the perf/ring program, which is a form
of benchmarking program that sends messages between actors as fast as
possible. Before this patch, on my laptop, ring consumes around 12GB of
RAM after running for about 1 minute sustaining around 800k messages per
second. With libgc, memory consumption stays constant at 4.5MB while
maintaining roughly the same message rate. I haven't collected a lot of
statistics and the variation between test rounds is probably around 10%.
I cannot see a clear pattern between using GC and not on this
performance.

As far as I understand, libgc does pause all threads every now and then
in order to collect a correct view of registers and similar. This means
that concurrency likely will be hit harder in Acton than the kind of
single threaded / low concurrency applications used in the above
mentioned benchmarking. ring is actually quite "single threaded" in that
there is only one outstanding message in total being passed between the
actors so we might not see those effects clearly.

Anyhow; GC seems to be working. WE HAVE A BLOODY GC ALMOST FOR FREE!!!

The longer term plan is obviously to implement our own GC that does
things the Acton way and will be way more efficient in various but until
then... this'll do I think :)

Fixes #952


plajjan:
Closing in preference of #964.

PR #957, opened 2022-10-08
Handle schema mismatch repairs, RTS txn and queue ops retries, queue flush side effects

aagapi:
This PR implements the following changes:

1) In the DB client API, in addition to the main return status (computed from the majority replies based on each RPC's semantics), also returns to client, for queue ops and txn commits, the minority failure status, if such a status exists, in variable called minority_status. This is useful to advise the client (RTS) to re-create missing or mismatching schema on minority nodes as they re-join the quorum.

This addresses the following problem: In the current RTS/DB implementation, while schema tables are automatically created on servers upon a DB node startup, schema mismatches can still happen in the form of missing actor input queues on DB servers that freshly joined a quorum (since the set of actors is dynamic, not part of the static schema). This caused the issue that newly restarted DB nodes were not able to re-join the quorum properly, as they missed this schema.

By the changes in this PR, the RTS now monitors the returned minority_status from all relevant operations (queue operations and txn commits) and triggers schema recreation on minority (or even majority) nodes if schema mismatches are observed.

2) In the RTS, we also handle DB ops and / or txn retries when either:
   2.1) Transactions are aborted by the servers (e.g. due to serializability conflicts with other txns)
   2.2) There is no quorum temporarily for a RPC (servers do not respond)
   2.3) There is a quorum, but there are schema mismatches on a majority of servers (in this case operations are re-tried after schema recreation as well)

3) Also in relation to 2.1), this PR refactors the function FLUSH_outgoing(), which is flushing an actor's outgoing messages after running a continuation to all destination actors affected, in a single txn (that also includes the actor's state update and input queue consume). In the previous form, the function could produce side effects in the local RTS in-queues of the other actors, if the DB txn containing the update is aborted by the system (e.g. due to concurrency failures), then retried until success (effects on local queues were not rolled back upon a txn abort). To solve this, this PR refactors FLUSH_outgoing in 2 separate functions, one that retries the DB transaction to the global queues first until success (FLUSH_outgoing_db), producing no effects on the local queues. Only after the global txn succeeds, the local queues are updated via the second function (FLUSH_outgoing_local).

4) Fixes a segfault in handle_txn_commit() due to unknown txn_ids when nodes rejoin a quorum with no knowledge of an ongoing txn

5) Handles schema errors between transaction validation and commit steps: Queue schema mismatch errors are now already detected in the txn validation round (via is_write_invalidated and table_verify_cell_version), and a new status, VAL_STATUS_ABORT_SCHEMA, is introduced. Like before, txns are only validated and committed when there is a quorum of VAL_STATUS_COMMIT responses, but the new status advises callers on schema mismatches on nodes that freshly joined a quorum to trigger repairs.

6) When waiting for quorum of server replies to a sync op, do not count invalid replies (e.g. schema mismatch errs) towards quorum to avoid timing out on valid replies.


aagapi:
Note: In the current form, only the DB operations that could actually produce schema mismatches in the current implementation are handled (minority_status set in the DB client API and schema repairs + retried triggered in the RTS). These are the queue operations (read_queue, consume, enqueue, subscribe etc), as well as the txn commit. Schema inconsistencies would only happen for these operations in the current implementation, because they stem from actor input queues being unknown to DB servers that freshly join a quorum (the set of actors is learned dynamically only once at RTS start up).
The other DB operations the RTS uses (e.g. inserts) would not currently produce schema inconsistencies, since the table schema is automatically created at server startup.
However, for completeness, I will update this PR before merging to also handle the remaining DB operations, like inserts, similarly (add the minority_status variable in the API, and also handle the other types of retries in the RTS, e.g. lack of quorum, for them).
Marking this PR as draft until this is completed.

plajjan:
Very cool! While I will read your code (not today though) I think I should actually spend more time in improving the test case to include various error scenarios, like partitioning db servers.

Have you manually tested stuff or do you have some test script or some process you could share?

plajjan:
If you add handling of schema errors for inserts, doesn't that mean we could start db servers with no schema and we would then create the schema from RTS as part of error handling for the first failed insert? That sounds pretty appealing to me! Only creating schema on startup is brittle and places heavy assumptions on environment. No assumptions and dealing with errors in generic ways make for robust programs!

aagapi:
> Very cool! While I will read your code (not today though) I think I should actually spend more time in improving the test case to include various error scenarios, like partitioning db servers.
> 
> Have you manually tested stuff or do you have some test script or some process you could share?

Yes, it would be very cool to have regression tests include partition scenarios, including some sequences of partitions which could cause issues.

To test, I used the test/count.act actor, and generated partitions by killing servers and/or the RTS in some sequences of failures, while making sure state from the actor is correct throughout (counter increases without skipping any numbers), and there are no crashes on either servers or RTSs.
Also, that the RTS blocks when a quorum is lost and it resumes when a quorum is re-gained.
And also, that servers and RTSs re-join the views in the gossip protocol when they come up and are marked as dead in gossip messages when they die. The gossip views and messages are printed in server logs, as well as RTS logs currently.

Some interesting sequences of failures to try are:

1) Killing one DB server (quorum remains live)
2) Bringing that server back (should re-join view as live, as well as quorum - respond to queries)
3) Killing the RTS and bringing it back (should continue counting from where it left off)
4) Killing one DB server, killing the RTS, bringing the RTS up, bringing the DB server up (restore state from an incomplete replication set)
5) Killing one DB server, killing the RTS, bringing the DB server up, bringing the RTS up (restore state from a replication set that includes servers that have been restarted)
6) Scenarios 4) and 5) but with killing 2 servers, along with the RTS (here the RTS should block while there is no quorum, but recreate schemas on re-joining servers, unblock and resume counting from where it left off once there is a quorum again)
7) Scenarios 4) and 5) but with killing all 3 servers, but not killing the RTS. Here, the RTS should block while there is no quorum and resume counting after there is. However, if in addition we kill the RTS everything, it will currently resume from the beginning, since there is no disk persistence.

aagapi:
> If you add handling of schema errors for inserts, doesn't that mean we could start db servers with no schema and we would then create the schema from RTS as part of error handling for the first failed insert? That sounds pretty appealing to me! Only creating schema on startup is brittle and places heavy assumptions on environment. No assumptions and dealing with errors in generic ways make for robust programs!

Right, since lack of schema errors will be propagated for inserts now, it would be possible to use that to trigger tables for complete schema creation as well. Just to note that right now what the RTS triggers when is started up is only the creation of input queues for actors (since the set of actors is dynamic and only known by the RTS). The "static" parts of the schema (tables for actor state, table of messages etc) are independently created by DB servers when they start up (similarly to a pre-installed schema server-side), so they currently involve no communication between RTS and DB servers at initial wire-up. We can discuss further options during the meeting.

PR #964, opened 2022-10-11
Add garbage collection using libgc

plajjan:
This adds garbage collection to the Acton run time system through the
use of libgc. We use link time redirection of malloc and friends in
order to catch all occurrences. Thread creation is hooked by including
gc.h in RTS and the DB client (which are the two places where we create
threads).

An earlier attempt was made to change our source code from using malloc
& friend to GC_MALLOC & friends but as it is exceedingly difficult to
find all mallocs, it regularly produced programs that segfaulted. libuv
and yyjson allows using a custom malloc but other libraries, like
utf8proc, doesn't appear to have functionality for specifying which
malloc to use and so it becomes rather difficult. Thus, this approach
was abandoned in preference of link time redirection of `malloc`.

Thread creation is hooked in source code as this is the only way that
works on MacOS. GC_USE_DLOPEN_WRAP allows wrapping dlopen so that on
Linux, hooks are installed at run time when the application is
dynamically linked with libpthreads. This facility does not exist on
MacOS which is why we are instead hooking it via source (by setting
GC_THREADS 1 and including gc.h). It ought to be more stable setting
this in source too. This is however untested, libgc even has an error in
gcconfig.h defined for this, so there's some uncertainty if there might
be bugs. It seems to work though.

libgc allows control via environment variables, most notably export
GC_DONT_GC=1 will disable the GC. This can be useful in case we suspect
the GC causes some problems. Once we've gained confidence with the GC,
we should disable this at compile time by setting -DNO_GETENV.

I've tested this primarily with the perf/ring program, which is a form
of benchmarking program that sends messages between actors as fast as
possible. Before this patch, on my laptop, ring consumes around 12GB of
RAM after running for about 1 minute sustaining around 800k messages per
second. With libgc, memory consumption stays constant at 4.5MB while
maintaining roughly the same message rate. I haven't collected a lot of
statistics and the variation between test rounds is probably around 10%.
I cannot see a clear pattern between using GC and not on this
performance.

As far as I understand, libgc does pause all threads every now and then
in order to collect a correct view of registers and similar. This means
that concurrency likely will be hit harder in Acton than the kind of
single threaded / low concurrency applications used in the above
mentioned benchmarking. ring is actually quite "single threaded" in that
there is only one outstanding message in total being passed between the
actors so we might not see those effects clearly.

The longer term plan is obviously to implement our own GC that does
things the Acton way and will be way more efficient in various but until
then... this'll do I think :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ ZIG_VERSION:=0.10.0
 CC=$(TD)/dist/zig/zig cc
 CXX=$(TD)/dist/zig/zig c++
 ZIG=dist/zig
+LIBGC=lib/libactongc.a
 export CC
 export CXX
 
@@ -113,7 +114,7 @@ BUILTIN_HFILES=$(filter-out $(ENV_FILES),$(wildcard builtin/*.h))
 BUILTIN_CFILES=$(filter-out $(ENV_FILES),$(wildcard builtin/*.c))
 
 DBARCHIVE=lib/libActonDB.a
-ARCHIVES=lib/dev/libActon.a lib/rel/libActon.a lib/libActonDeps.a
+ARCHIVES=lib/dev/libActon.a lib/rel/libActon.a lib/libActonDeps.a lib/libactongc.a
 
 DIST_BINS=$(ACTONC) dist/bin/actondb dist/bin/runacton
 DIST_HFILES=\
@@ -164,7 +165,7 @@ backend/txn_state.o: backend/txn_state.c backend/txn_state.h $(DEPSA)
 backend/txns.o: backend/txns.c backend/txns.h $(DEPSA)
 	$(CC) -o$@ $< -c $(CFLAGS_DB)
 
-backend/client_api.o: backend/client_api.c backend/client_api.h backend/log.h backend/hashes.h $(DEPSA)
+backend/client_api.o: backend/client_api.c backend/client_api.h backend/log.h backend/hashes.h $(DEPSA) $(LIBGC)
 	$(CC) -o$@ $< -c $(CFLAGS_DB)
 
 backend/failure_detector/db_messages.pb-c.o: backend/failure_detector/db_messages.pb-c.c backend/failure_detector/db_messages.pb-c.h $(DEPSA)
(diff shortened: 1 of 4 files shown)


plajjan:
We are seeing intermittent failures on this branch and I've been running some tests in order to get a better view of this.

I've run 1000 iterations for each case
```
OS              branch  GC      test        FAIL
debian:testing  libgc   yes     db_app         6
debian:testing  libgc   yes     tcp_client     6
debian:testing  libgc   yes     tcp_serv      12
debian:testing  libgc   no      db_app         5
debian:testing  libgc   no      tcp_client    14
debian:testing  libgc   no      tcp_serv      17
ubuntu:20.04    libgc   yes     db_app         0
ubuntu:20.04    libgc   yes     tcp_client     1
ubuntu:20.04    libgc   yes     tcp_serv     132
ubuntu:20.04    libgc   no      db_app         0
ubuntu:20.04    libgc   no      tcp_client     1
ubuntu:20.04    libgc   no      tcp_serv       1
macos12         libgc   yes     db_app         1
macos12         libgc   yes     tcp_client     3
macos12         libgc   yes     tcp_serv       0
macos12         libgc   no      db_app         0
macos12         libgc   no      tcp_client     0
macos12         libgc   no      tcp_serv       1
macos12         main    no      db_app         0
macos12         main    no      tcp_client     1
macos12         main    no      tcp_serv       1
```

the notable outlier is the tcp_server test on Ubuntu 20.04 which sees 13.2% failures and that is also what I "feel" from CI runs, it fails fairly regularly. With the GC turned off we see a single failure, so that's a stark difference - the GC has definitely made things worse.

However, on other platforms we don't see that. I think it is interesting to note that we have some level of failures even without the GC (see reference tests on main branch at bottom of list).

On debian:testing, which is what I run on my main machine, there is certainly a visible increase of failures with GC turned on. Since we see failures even without the GC my gut feeling is that the test is somehow timing related and the GC makes this worse since it influences timings. This is a guess though, while I have looked at the tests a bit I haven't been able to discern what makes them fail. At least it's not a segfault!

I want to emphasize that I haven't seen a single segfault with this GC configuration! If the GC was broken or didn't interoperate well with our code, I would expect segfaults. As that doesn't happen, I think it's basically sound... although it leaves the increased failure rate a mystery.

plajjan:
It's looking somewhat optimistic. I've now run 2000 test invocations on my local computer without a single failure. It usually fails way earlier, so things are indeed looking good.

All CI tests have passed too, but I'll run through a couple of more runs. We could perhaps consider adding more extensive testing, like adding Ubuntu 20.04 back in, before merging this since we've seen failures on it. On the other hand, I really do think we've addressed those failures with #1088, so perhaps we should just move forward and we'll get around to testing more on older Ubuntu later. As I said, I'll run through quite a few more complete CI runs anyway before we merge this.

plajjan:
For the record, my tests failed around 2500 runs but I don't think the failure is like any other we've seen. It complains about getting the wrong data back from the test server:
```
testERR: Got unexpected data from server in state 2: 1
```

I dunno why that is but I reckon it is more likely it has to do with incomplete handling in the test runner script than with Acton itself.

PR #976, opened 2022-10-16
Cmdlineparser

### LINKED ISSUE #893: Rework acton compiler argument parsing to use sub-parsers for commands
The argument parsing of actonc is currently somewhat limited. In essence, it assumes that the compiler is invoked to compile an .act file. While that is the main use, it is not our only intended use.

We are currently using options.applicative (https://hackage.haskell.org/package/optparse-applicative-0.17.0.0/docs/Options-Applicative.html).

I believe we should convert to using the sub-parser feature so that each command can have its distinct arguments.

To implement, move all current functionality related to building in under a `build` command. We support dumping .ty file content, which should go in under a separate `dump` command.

Finally, a new command "cloud" should be introduced with the hierarchy:
- actonc cloud
  - run
  - list
  - show
  - stop

sydow:
Fixes #893

### DIFF
--- a/compiler/Acton/CommandLineParser.hs
+++ b/compiler/Acton/CommandLineParser.hs
@@ -1,0 +1,145 @@
+module Acton.CommandLineParser where
+
+import Options.Applicative
+
+parseCmdLine        :: IO CmdLineOptions
+parseCmdLine        = execParser (info (cmdLineParser <**> helper) descr)
+
+data CmdLineOptions = CompileOpt [String] CompileOptions
+                    | VersionOpt VersionOptions
+                    | CmdOpt Command
+                    deriving Show
+
+data VersionOptions = VersionOptions {
+                        version :: Bool,
+                        numeric_version :: Bool
+                     } deriving Show
+                     
+data Command        = New NewOptions
+                    | Build BuildOptions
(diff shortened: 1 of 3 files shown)


sydow:
After two stupid mistakes I now arrived at the failures I experienced before pushing. One test case fails, and I canot understand why. I certainly can do the following:

qualified_root> pwd
/Users/sydow/acton/compiler/test/actonc/project/qualified_root
qualified_root> actonc build --root test.main
Building project in /Users/sydow/acton/compiler/test/actonc/project/qualified_root
  Compiling b.act for release
   Already up to date, in    0.001 s
  Compiling test.act for release
   Already up to date, in    0.000 s
Building executable out/rel/bin/test

which seems to me to be the test that fails. I can't get further with this.

plajjan:
The test is wrong. The function buildThing checks if a "thing" is an acton project or a single .act source file and builds accordingly. If it's a project, it compiles the project by running `actonc build`, i.e. buildThing adds the `build` argument and sets the working directory to the project.

In this case, the test case itself sets arguments to `build --root test.main` so the final cmd args to actonc become `build build --root test.main`.

plajjan:
I squashed the two extra commits... and added a fix.

PR #977, opened 2022-10-16
OLD: Add `xml` module to stdlib

plajjan:
- [ ] move / rewrite examples/xmltest.act to test/stdlib_auto/test_xml.act
  - [ ] make sure we have some automatic test cases
- [ ] fix dependencies, install, copy libs etc
- [ ] decide something for indent, either remove it or make it optional through argument?
- [ ] empty nodes for storing text? why? shouldn't it go in content field?
- [ ] should we have decodeFile? isn't it better to leave xml module completely pure with no IO and leave it to acton file module to read data? then we also go through acton for content encoding
  - libxml2 depends on libicu (on Linux) or iconv (on mac) for internationalization and conversion between formats
  - seems more acton-way to make 
  - if we compile libxml2 ourselves we can keep down its size by not doing internationalization - thus only supporting utf-8, but since that is what we do internally anyway, that's ok?
    - if we want to support other formats later we should add that as a generic feature in acton, not specific to xml

### DIFF
--- a/compiler/ActonCompiler.hs
+++ b/compiler/ActonCompiler.hs
@@ -739,7 +739,7 @@ buildExecutable env args paths binTask
         buildF              = joinPath [projPath paths, "build.sh"]
         outbase             = outBase paths mn
         rootFile            = outbase ++ ".root.c"
-        libFilesBase        = " -lActonProject -lActon -lActonDB -lprotobuf-c_a -lutf8proc_a -luv_a -lpthread -lm -ldl "
+        libFilesBase        = " -lActonProject -lActon -lActonDB -lprotobuf-c_a -lutf8proc_a -luv_a -lpthread -lm -ldl -lz -llzma -liconv -lxml2 "
         libPathsBase        = " -L " ++ sysPath paths ++ "/lib -L" ++ sysLib paths ++ " -L" ++ projLib paths
 #if defined(darwin_HOST_OS) && defined(aarch64_HOST_ARCH)
         libFiles            = libFilesBase
--- a/examples/xmltest.act
+++ b/examples/xmltest.act
@@ -1,0 +1,13 @@
+import xml
+import file
+ 
+actor main(env):
+    fa = file.FileAuth(env.auth)
+    rfa = file.ReadFileAuth(fa)
+    rf  = file.ReadFile(rfa,env.argv[1])
+    bs = rf.read()
+    s = bs.decode()
+    t : xml.Tag = xml.decode(s)
+    print(xml.encode(t))
+    env.exit(0)
+
--- a/examples/xmltest.act
+++ b/examples/xmltest.act
@@ -2,12 +2,13 @@ import xml
 import file
  
 actor main(env):
-    fa = file.FileAuth(env.auth)
-    rfa = file.ReadFileAuth(fa)
-    rf  = file.ReadFile(rfa,env.argv[1])
-    bs = rf.read()
-    s = bs.decode()
-    t : xml.Tag = xml.decode(s)
+#    fa = file.FileAuth(env.auth)
+#    rfa = file.ReadFileAuth(fa)
+#    rf  = file.ReadFile(rfa,env.argv[1])
+#    bs = rf.read()
+#    s = bs.decode()
+#    t = xml.decode(s)
+    t = xml.decodeFile(env.argv[1])
     print(xml.encode(t))
     env.exit(0)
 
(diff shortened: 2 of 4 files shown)


plajjan:
@sydow I'm creating a PR so that:
- we have a natural place to talk about this code
- see CI run

It's in draft status to avoid accidental merge (although it would have to pass CI anyway to be mergable)

plajjan:
For Linux I believe we need to install `libxml2-dev` but it also looks like it places header files in `/usr/include/libxml2/libxml` rather than in `/usr/include/libxml` as I would have expected. This means we'd need to add `-I/usr/include/libxml2` to cc args.

For pure C modules, we support this by custom make targets, which you can see in `stdlib/Makefile`... however, since this is a mixed mode `.ext.c` style, we don't run those make targets. I'm not sure what is the best hook point for this. There's always adding it to ActonCompiler.hs, which isn't pretty at all but I'm also inclined not to block on tasks like this but rather solve it short term with a hack and figure out a better way over time. It's a pity though since it will affect all files whereas we really only need this during the compilation of our xml module.

plajjan:
@sydow I think we should have a argument to encode to control whether it should pretty print the resulting XML with indent and new lines. It's nice for debug output but for the normal programmatic use cases, it's not needed and I'd argue even unwanted. We just want the dense output, i.e. simplest possible output. What do you think?

I see you've added a decodeFile. Is that wise? Shouldn't we rely on our file module to read files and then send the data as a acton value to our xml module so it can be pure rather than do IO? I assume it doesn't hook into libuv but does direct IO... I'd like to avoid that if possible I think.

plajjan:
I'm confused about the empty nodes for text. Is this how libxml2 returns it?

I figured that for:
```xml
<a>hello</a>
```
we would get:
```
Tag(name="a", content="hello")
```

and for something more complex, with interleaved tags, like:

```xml
<a>hello<b>FOO</b> world</a>
```
we would get:
```
Tag(name="a", content="hello world")
```

content could also be a list of str for when it's interleaved like this!? but introducing extra Tag objects seems more complex for little gain, or am I missing something? :)

plajjan:
@sydow I agree that copying data around is unattractive but I think that pros of reusing other acton components still outweighs that unattractiveness, and thus I think we should remove decodeFile :)

sydow:
Hi,

I am confused about in particular your second example, "<a>hello<b>FOO</b> world</a>”.   With this input, libxml2 builds a tree 

(an xmlNode, see https://gnome.pages.gitlab.gnome.org/libxml2/devhelp/libxml2-tree.html#XMLNODE) as follows.

The root node has name “a”, type  XML_ELEMENT_NODE and no content. It has three children:
- a node with name “text”, type  XML_TEXT_NODE         and content “hello”
- a node with name “b”,     type  XML_ELEMENT_NODE and no content
- a node with name “text”, type  XML_TEXT_NODE         and content “world”

Is this not what is wanted? To me the merging of text you suggest seems strange, but...

Björn

plajjan:
> Hi, I am confused about in particular your second example, "<a>hello<b>FOO</b> world</a>”. With this input, libxml2 builds a tree (an xmlNode, see https://gnome.pages.gitlab.gnome.org/libxml2/devhelp/libxml2-tree.html#XMLNODE) as follows. The root node has name “a”, type XML_ELEMENT_NODE and no content. It has three children: - a node with name “text”, type XML_TEXT_NODE and content “hello” - a node with name “b”, type XML_ELEMENT_NODE and no content - a node with name “text”, type XML_TEXT_NODE and content “world” Is this not what is wanted? To me the merging of text you suggest seems strange, but...

@sydow I guess I'm influenced by Python's xml etree module, which I think does roughly what I describe... at least it has the text of a node in a `.text` attribute (akin to our .content).

I suppose you're right in that merging text is a little strange, perhaps foremost because it would be hard to make it reversible and thus encode / decode to be 1:1 mappings back and forth. I still think it would be convenient but if we have a "merged text" field, it would need to be as a "extra" thing, out of convenience.

Reusing the Tag both for actual tags and text is weird to me though. Is the solution to add a second object type, so we have Tag and Text (which would map to XML_ELEMENT_NODE and XML_TEXT_NODE)? Should our Tag be called Node?

plajjan:
@sydow apologies, my attempt at fixing the build for Linux probably broke your build on Mac. I've removed those commits from this branch and will only do that sort of work separately (using my `xml` branch). Use `git reset --hard 3c60cfe58a8aa557d9d7afabce89c24f7f360690` to remove my commits on your branch locally.

plajjan:
Closing in preference of #990

PR #987, opened 2022-10-20
Bigint

sydow:
This push changes the type int to contain arbitrary size integers, built on BSDNT. It is not ready for merging yet, for several reasons:

- the BSDNT library is not yet incorporated into the build framework. I guess that this will happen soon.

- serialisation does not work. It has to be changed for int, and since many builtin types have an int component, those do not work, either. I have problems in testing serialisation; there is some test code in builtin/serialization_tests, but it seems to be outdated. Has someone tested serialisation recently in some way?

- less important is that formatted output does not work for ints. I have tried to look at the code in CodeGen.hs, but find it difficult. What must be accomplished is that the code cannot anymore access the long value of a variable n : int using n->val, but must do from$int(n). So this is simple once one understands function castLit. Johan?

I push this now anyhow, since there are several things to discuss. First, I should say that most things work well. All tests in the test suite pass, and I have been able to compute the value of e (the base of the natural logarithms) to 10000 decimals (or, rather and equivalently, computed the value of e*10^10000 as an int). On the other hand, integer computations that only need 64 bits become slower, naturally.

As it stands now, I followed Kristian’s suggestion and let int be the default integer type. So far, I have also i64, but the other bounded types should be very simple to add. The problem is that, as it stands, i64 is a rather useless type. To make it useful, I think one needs to change several of our protocols and change parameter types from int to Integral. I am thinking of indexing and slicing, but also lengths of Collections, and others. Not only protocols should change, but also e.g. range. I think that this is also easy, but tedious. I believe that the type checker does not need to change, but the protocol instances become more complicated. This will affect both complexity of the generated C code and the speed of execution. But
- Acton coding with bounded types will be simpler; by a signature for just one variable, the default will be overruled and all computations done in a bounded type.
- in the not too distant future I hope we will be able to simplify generated code, and often remove witnesses altogether.
- in the a bit more distant future, we may be able to sometimes eliminate boxing altogether, and achieve C speed for bounded integer arithmetic.

So, I recommend to do this change, but that could come later. We could start by letting  int be a non-bounded type very soon and gain experience.

Another, related, issue is subtyping of integer types. As of now, i64 is not a subtype of int. This could not be handled by implicit coercion; an int is a pointer to a struct consisting of
- an array of unsigned longs.
- a size field which also codes for sign (a negative size for negative numbers).
- a field which gives the allocated size of the array (one may have allocated a bigger array than necessary (BSDNT ints are mutable; incremental updates (+= etc) really updates the first operand)).

As for i64, i32, i16 etc, I get headache by just considering how subtyping would work with endianness, memory allocation for structs, C code, etc.

Finally, I have not implemented Logical for int. This is not supported in BSDNT’s high level interface zz, but only in the lower level nn. It would be possible, but I tend to think that logical operations are better suited for bounded ints. If someone attempts, a “not implemented” message could suggest to work with a bounded type.

Comments of any of these issues are welcome!


sydow:
I forgot one thing: it looks as an enormous amount of change, but this is since I have tried to adopt to a four steps indentation policy...

PR #988, opened 2022-10-21
Add & use actonc --always-build

### LINKED ISSUE #967: test suite should ensure fresh compilation
In our tests we invoke actonc to compile a program and then run it. If we run the test suite multiple times we will also invoke actonc multiple times but as it has an up-to-date check it will not actually recompile the output program if it is newer than the source file.

We should guarantee that we always recompile the program, perhaps by removing the output program first (from the test suite).

plajjan:
--always-build skips the up-to-date check in actonc and always builds, which is necessary for correct tests since we otherwise risk relying on outdated results. The up-to-date check only inspects source and output files on disk. For example, if a .ty file is newer than its source .act file. If actonc itself has changed and updated the format of the .ty file, we won't catch this. --always-build is thus the only safe option for testing.

Fixes #967.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -415,14 +415,14 @@ dist/types/__builtin__.ty: builtin/ty/out/types/__builtin__.ty
 
 builtin/ty/out/types/__builtin__.ty: builtin/ty/src/__builtin__.act $(ACTONC)
 	@mkdir -p $(dir $@)
-	$(ACTC) $<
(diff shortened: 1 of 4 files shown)


plajjan:
@nordlander FYI!! You probably want this in your deactorizer branch!

nordlander:
I sure do — thanks!

PR #990, opened 2022-10-21
Add xml module

plajjan:
Remaining todo items:
- [ ] remove decodeFile
- [ ] remove indent (decided with Björn in call) - we can always add it back in later, but let's make it stupid simple for now
- [ ] use separate type for text to make it a little nicer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -403,7 +403,7 @@ deps/instdir/lib/libuv.a: deps/libuv
 # /deps/libxml2 ------------------------------------------
 LIBXML2_REF=644a89e080bced793295f61f18aac8cfad6bece2
 deps/libxml2:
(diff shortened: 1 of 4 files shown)


plajjan:
@sydow I have fixed the libxml2 installation stuff now, so this branch compiles. I don't remember if you have pulled this branch or not, if you already have, it is best you delete it first and pull a fresh copy. I had to rebase it to make it work with changes on master. So:
- git checkout main
- git branch -D 835-add-xml-module
- git fetch
- git checkout 835-add-xml-module
And you should be able to continue from there!

plajjan:
I'm fairly happy with this as is. We can make further improvements but it's important to get it in, so merging now.

sydow:
Of course you are right.

On the exceptions issue you may have opinions on my bigint push (in from$int)…

Björn

PR #993, opened 2022-10-21
Pull libxml2 from github

plajjan:
Seems to be problems pulling it from gitlab from github actions, so let's try pulling from github instead, where there's a read-only mirror.


plajjan:
This is instead incorporated into #990. Closing this one.

PR #1012, opened 2022-10-27
Add example pidigits program

plajjan:
Lots of debugging crap that should be removed once it's working properly..


plajjan:
Closing in preference of #1013 which actually compiles and works!

PR #1060, opened 2022-11-07
Do -Werror for all, enabling it on MacOS

### LINKED ISSUE #218: Do -Werror for builtin / Stdlib etc compilation on OS X
It seems we (I) only enabled -Werror for Linux. Should do it on OS X too, I suppose!

plajjan:
Fixes #218

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,9 +66,9 @@ endif
 endif # -- END: Apple Mac OS X -------------------------------------------------
 
 
+CFLAGS += -Werror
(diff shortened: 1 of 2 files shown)


plajjan:
Got a segfault  during compilation on Linux, that's weird - never seen that before. I don't think it's related but noting it down here.

plajjan:
Ran another rebuild of CI tests without failures, so merging.

PR #1076, opened 2022-11-08
fixed bug in truediv for ints

sydow:


sydow:
forgot to fix one thing...

Issue #1078, opened 2022-11-08
Sort out UB in backend


plajjan:
As demonstrated by #1082, enabling sanitization of undefined behavior in the backend does not pass CI - it fails on Mac. We should not rely on UB and thus this should be fixed.

@aagapi assigning this to you. I think you can reproduce the problems by running that branch on your mac?

PR #1082, opened 2022-11-08
Sanitize UB in backend DB

plajjan:
Part of #1078.


plajjan:
Bah, I realize I messed up the CFLAGS when I separated them out for the DB in #1066 - they no longer include -target. Not sure if that actually affects this particular problem or not, I don't think so - UB should be UB, but I'll fix that regardless!

plajjan:
Very outdated, we now use for compilation. Closing this.

PR #1089, opened 2022-11-09
Reset comm thread timeout

### LINKED ISSUE #1085: Fix DB client_api select loop resetting timeout
Been looking into GC problems and found at least one thing that's not right in the DB client_api, we only initialize the timeout struct fed into select once. I think it need to be reset on every loop iteration.

From man select on my machine:
>       On Linux, select() modifies timeout to reflect the amount of time not slept; most other implemen‐
>       tations do not do this.  (POSIX.1 permits either behavior.)  This causes problems both when Linux
>       code  which  reads timeout is ported to other operating systems, and when code is ported to Linux
>       that reuses a struct timeval for multiple select()s in a loop without  reinitializing  it.   Con‐
>       sider timeout to be undefined after select() returns.

plajjan:
On Linux, select modifies the timeout value to reflect the amount of time not slept. Since we want to continue sleeping for 3 seconds on every iteration we need to reset the value on every loop iteration.

Fixes #1085.

### DIFF
--- a/backend/client_api.c
+++ b/backend/client_api.c
@@ -283,8 +283,6 @@ void * comm_thread_loop(void * args)
 {
     remote_db_t * db = (remote_db_t *) args;
     struct timeval timeout;
-    timeout.tv_sec = 3;
-    timeout.tv_usec = 0;
     char in_buf[BUFSIZE];
     int msg_len = -1;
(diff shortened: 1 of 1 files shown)


aagapi:
Good find!

PR #1093, opened 2022-11-10
Test on Ubuntu 20.04

plajjan:
We dropped Ubuntu 20.04 when it was clear that libxml2 required a newer version of automake than available on Ubuntu 20.04 and it was just too much to handle at the time. As it turns out, that requirement is probably not a very hard requirement at all, likely the artifact of running autoupdate on a relatively new machine and voila, the required version is bumped to something quite new.

It seems to work just fine compiling and running after we rewrite the required version to what is available on Ubuntu 20.04.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,6 +88,8 @@ jobs:
           - os: "debian"
             version: "11"
           - os: "ubuntu"
+            version: "20.04"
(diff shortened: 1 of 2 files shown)


plajjan:
@aagapi FYI, brought back Ubuntu 20.04 testing and added the necessary changes to our Makefile to get it to build...

PR #1106, opened 2022-11-13
removed incremental operations from int.c

sydow:


sydow:
BSDNT provides incremental (mutating) arithmetic which should not be used in Acton

PR #1107, opened 2022-11-14
improved int-to-str conversion

sydow:


nordlander:
I restarted it, went well this time!

(The restart symbol is normally hidden for some reason, but appears as a circular arrow if you hover over a job name in the job list of the corresponding PR.)

sydow:
> On 14 Nov 2022, at 16:28, Johan Nordlander ***@***.***> wrote:
> 
> 
> I restarted it, went well this time!
> 
> (The restart symbol is normally hidden for some reason, but appears as a circular arrow if you hover over a job name in the job list of the corresponding PR.)

Well, that is a clever trick! Thanks; I have merged.

Björn

plajjan:
FWIW I think restart is hidden until all jobs have run. Once they have run, either succeeding or failing, you get a icon to rerun failed or all jobs for one workflow run.

sydow:
> On 14 Nov 2022, at 19:10, Kristian Larsson ***@***.***> wrote:
> 
> 
> @sydow first off, great job!
> 
> We are currently suffering from a slow udivmod, so I wonder if this is faster because it avoids udivmod or if it's always faster?

Well, I don’t know what udivmod is. Of course, I googled and can see that it is something that is defined in libgcc, but I cannot say more. I do use BSDNT’s divrem function, which gcc may compile to use udivmod under the hood. 
> 
> So, using gcc or clang, is this faster than bsdnts int to str? If so, should we submit this upstream to bsdnt or is it worse in some other way?

I think the function I checked in today is MUCH faster than BSDNT’s zz_get_str. In fact, I find it surprising that BSDNT, which seems well designed and is built by experts in the field, did not put more effort into this. The conversion to a str is done by repeated divmod with the denominator 10, which gives one decimal digit per division. So maybe udivmod is not so slow, but used an excessive number of times? 

My function uses a divide-and-conquer strategy, which divmods by a power of 10 which is roughly the square-root of n, and then recursively converts the quotient and remainder in the same way. It is well known that this is a superior strategy, and I am sure that the BSDNT developers know this. I believe that there has not been much development of the library in the last 9-10 years, but it is there to be useful for those who need a bigint library. I believe that we may continue to use it; after all, the problems that we looked at the last few days, i.e. computing hundreds of thousands of digits of mathematical constants is not our top priority. But if we have arbitrary size ints as the default integer type, then we should be able to do reasonably well, and I think we do. 

I am more worried that we are a little slow on more mundane tasks like looping over a range, which most programmers use every day. Such a loop creates a boxed integer at every step. If we think of nested loops like in the spectral-norm problem, the inner loops run 100’s of times and create 100’s of boxed integers in every run; the same numbers, over and over again. In fact, the spectral-norm program runs much faster if one once, in main, constructs list(range(0,n,1)) and then sends that list as a parameter to all functions, so that they can reuse it for list iteration (which is very fast). Will we be able to unbox this away when the numbers are not of fixed size?
> 
> If not, i.e. if bsdnt int to str is faster under gcc / clang, then this is a temporary workaround until we get a faster udivmod, right?

I think that we now can convert ints to strings as fast as we can reasonably need. Converting a number with 100000 decimal digits takes something like 10 milliseconds, and I think that is OK. For a million digits, times start to go up in the 100’s of milliseconds, but to fix that I believe we would have to implement FFT multiplication. I see no need to improve things here.

Björn

plajjan:
To clarify one of the few things I have learnt recently; libgcc is the C runtime of gcc. In clang it is called compiler_rt (compiler run time) and zig uses the same name although the implementation is different (zig has its own written in zig, not C). So udivmod exists in gcc, clang and zig compiler_rt. When we use `zig cc` (our default) we end up using zigs compiler_rt udivmod. When we use gcc to compile, we end up with libgcc udivmod and when we use clang we get llvm compiler_rt udivmod!

As for BSDNT, it is largely static but @wbhart did merge things in 2021, so I suppose that pull requests are welcome, at least if they are obvious. Could there be other drawbacks to our (your) design? Less portable? Works less well on certain CPU architectures?

I note that with Haskell GHC 9 they have done lots of cleaning so that different implementations of Integers are now supported, either based on ghc or on other libraries. I have seen multiple forum postings mentioning bsdnt as a backend. It feels like it could have an increased interest from Haskell. While Haskell isn't huge, it sure is bigger than Acton and having an academic language isn't bad when working wit numbers. Maybe BSDNT could see an upswing too...

plajjan:
@sydow I agree that we should expect most Acton programs not to be heavy duty integer arithmetic but rather... well, other things but often using for loops and similar constructs, so optimizations there are likely more useful than in pure integer functions...

can we interpret literals as i64 if they fit in i64 and large otherwise and have range() functions etc act differently based on int vs i64 input?

plajjan:
> For a million digits, times start to go up in the 100’s of milliseconds, but to fix that I believe we would have to implement FFT multiplication. I see no need to improve things here.

Couldn't help myself but notice that the BSDNT README mentions FFT multiplication with a link to https://github.com/flintlib/flint2/tree/trunk/fft which hosts FFT code that is BSD licensed although it currently relies on GMP. It could be converted to run on BSDNT data structures instead I guess and then we'd have fast FFT multiplication!?

sydow:
> On 14 Nov 2022, at 23:07, Kristian Larsson ***@***.***> wrote:
> 
> 
> @sydow I agree that we should expect most Acton programs not to be heavy duty integer arithmetic but rather... well, other things but often using for loops and similar constructs, so optimizations there are likely more useful than in pure integer functions...
> 
> can we interpret literals as i64 if they fit in i64 and large otherwise and have range() functions etc act differently based on int vs i64 input?

Literals are already overloaded, so 13 can be in both int and i64, depending on context. 

But parameters to range, and indexing, and slicing, have type int hardcoded. We could think of changing those types to be Integral, but I think we wait until we get rid of witnesses in the most common cases. But this does not solve the problem with unboxing; in nested loops as 

for i in range(0,n,1):
    for j in range(0,n,1):
         use(i,j)

when we come to a new iteration of the inner loop we need a new value for j. The range iterator of course keeps track of the start/step/stop parameters so it does some check whether to stop, updates the internal next value *and allocates a new int node*; the pointer to this node is the new j. This is the same for int and i64, but the overhead is a bit greater for int (since the int node itself contains a pointer to an array of limbs, i.e. 64 bit chunks). So for each iteration of the inner loop, we build again  the same list  [0,1,2,...n-1] of boxed integers. (And of course, the code in use will immediately unbox the value to get the real int inside). This is a challenge for code improvement to simplify this in many (most) cases. But I know that Johan is very good at these things, so I am optimistic. We only need to find time for him…

sydow:
> On 14 Nov 2022, at 23:11, Kristian Larsson ***@***.***> wrote:
> 
> 
> For a million digits, times start to go up in the 100’s of milliseconds, but to fix that I believe we would have to implement FFT multiplication. I see no need to improve things here.
> 
> Couldn't help myself but notice that the BSDNT README mentions FFT multiplication with a link to https://github.com/flintlib/flint2/tree/trunk/fft which hosts FFT code that is BSD licensed although it currently relies on GMP. It could be converted to run on BSDNT data structures instead I guess and then we'd have fast FFT multiplication!?

Yes, this may be a future development of BSDNT. But I don’t think this is for us to do. There is a lot of fine-tuning to do and he also mentioned that the code he refers to should be simplified.

And this is something which will only start to have dramatic effects when we have really HUGE numbers. So I think we leave that as something for the future.

PR #1115, opened 2022-11-20
improved int-to-str conversion

sydow:


sydow:
Sorry for a misleading comment: this pull request improves conversion from a (long) digit string to an int.

PR #1118, opened 2022-11-24
fixed bug in decl of __init__ for i64

sydow:


sydow:
This pull request fixes outdated syntax for the declaration of i64.__init__  in __builtin__.act

PR #1127, opened 2022-12-21
Queue bulk subscribe read

aagapi:
This PR refactors the way queue subscriptions are handled by servers and clients to make them more efficient.
Previously, clients (RTSs) had to individually subscribe to notifications from remote queues, for each of the actor's queue they handled. Particularly, this had 2 inconveniences:
1) Whenever an RTS notices it becomes responsible for an actor (via gossip views installed from servers and running consistent hashing), it needs to individually subscribe to the input queues of each actor it became responsible for, and also un-subscribe to queues of actors it is not anymore responsible for. This adds un-needed complexity in the RTS to handle the distributed RTS case, as well as un-needed extra RPCs between RTSs and servers when the RTS membership changes
2) Whenever there are messages in the remote queues for actor queues, the RTS(s) responsible for that actor receive independent notifications for each queue, which is inefficient

The changes in the current PR are the following:

1) We introduce queue groups. Essentially, this decouples client subscriptions from queue membership to a queue group on the server side. In the new API, the clients (RTSs) subscribe to a queue group rather than individual queues. In the current implementation, this happens exactly once (when the RTS comes up), and there is a single default group for each RTS.
2) Queue membership to groups is automatically updated on the server side from installed gossip views using consistent hashing mapping, the same way that actor membership gets updated on the RTS side. This means that, as RTSs die or re-join the system, actor input queues automatically get assigned to the right group on the server side.
3) To make things efficient on the notification path, current group membership of queues (as well as client subscriptions to groups and corresponding per-client metadata, such as read heads) is cached within each queue's data in the backend (so that inefficient consistent hashing lookups do not need to occur at each enqueue).
4) RTSs receive a single notification per-group whenever there is data in any of the queues it is responsible for, rather than multiple notifications per-actor queue
5) The code base related to this area has been re-factored to some degree. For instance, the consistent hashing implementation has been pulled out into a separate API, that is used on both the RTS/client side (to compute actor-to-RTS mapping) and on the DB server side (to compute queue-to-group mapping).


rts/rts.c
```diff
@@ -2470,7 +2456,15 @@ int main(int argc, char **argv) {
             deserialize_system(start_row);
             log_info("Actor state restored from DDB.");
         } else {
-            log_info("No previous state in DDB; Initializing database...");
+            log_info("No previous state in DDB; Initializing database...\n");
```
plajjan, line 2459: No `\n`

plajjan (approved):
Puh, I've read through quite a bit of the code, skimmed some parts. I can't really say much on the new queue group stuff - too much to take in for me. Looks nice overall. Changes in RTS look OK.

For me it's much better to try the stuff out and rely on testing and empirical use in order to verify correctness.

Had some nitpicks on log messages which I think have gotten back a `\n` which we have previously removed (likely merge artifact).

aagapi:
> Puh, I've read through quite a bit of the code, skimmed some parts. I can't really say much on the new queue group stuff - too much to take in for me. Looks nice overall. Changes in RTS look OK.
> 
> For me it's much better to try the stuff out and rely on testing and empirical use in order to verify correctness.
> 
> Had some nitpicks on log messages which I think have gotten back a `\n` which we have previously removed (likely merge artifact).

Thanks! Correct, the '\n' at the end of some logging messages was re-introduced after the merge with main (in my branch these were printf's which needed a '\n'). Thanks for noticing, will fix the remaining ones.

plajjan:
And to clarify, I haven't done any testing or tried to use this. Seeing how testing fails in CI on Linux I think that needs to be fixed first though :)

aagapi:
> And to clarify, I haven't done any testing or tried to use this. Seeing how testing fails in CI on Linux I think that needs to be fixed first though :)

Yes, I believe that CI failure is also an issue also related to the merge with the changes in the main branch, will fix it before you can test it.

plajjan:
@aagapi ignore the mac build failures, they are not related to your branch. We just haven't had any builds on main or anywhere for a time... something has changed with GHC so that it now fails, I've seen it in other CI runs, see #1139.

plajjan:
I've merged #1145 which is a temporary workaround for the macos build issues. Please rebase this branch on main and you'll get that fix.

PR #1142, opened 2022-12-28
Update to stack LTS 19.33

plajjan:


plajjan:
This is dead. LTS 19.33 uses GHC 9.0 which doesn't' exist in Homebrew. Closing and deleting branch.

PR #1144, opened 2022-12-28
fixed bug in __eq__ for dicts

sydow:


plajjan:
I've rebased (& force pushed) this to get the macos build workaround from #1145.

plajjan:
Got this to pass CI after rebase.

@sydow what does this fix though? That I can check equality on two dicts that are keyed by str?

sydow:
Sorry for not explaining, but this is part of our general lack of documentation/language definition.

The intention is that the code should implement this line from __builtin__.act:

extension dict[A(Hashable),B(Eq)] (Ord):

i.e. any two dictionaries of type dict[A,B] where B implements Eq can be compared (A must implement Hashable for the type dict[A,B] to be well defined).

d1 == d2 is true if the set of keys in the two dicts are equal and for each key k  it holds that d1[k] == d2[k].
d1 < d2 is true if the set of keys of d1 is a subset of the set of keys of d2 and for each key k in d1it holds that d1[k] == d2[k] 
etc for the other comparisons. 

So with the definitions

a = {“a”:1}
b = {“a”:1}
c = {“a”:1, “x”:2}

we have that a==b and a < c are both true. You can compare dicts keyed by any Hashable type as long as values can be compared for equality. As always, testing is minimal; this is a serious problem for us.


Björn

PR #1145, opened 2022-12-29
No concurrent builds

plajjan:


plajjan:
I don't know why, but removing concurrency from the mac builds appears to fix the errors we've seen lately. Will merge this now in order to unblock other stuff and open an issue for further investigation since we obviously want concurrency.

PR #1148, opened 2022-12-29
Statically compile actonc

### LINKED ISSUE #1146: Statically compile actonc
I've been reading a few posts on static compilation of haskell programs... And I think it should basically be possible for us.

So let's try and see if we can statically compile actonc to avoid run time dependencies. This should make it possible to run on somewhat older releases of Linux distros too...

One reason haskell compiles dynamically linked is to avoid licensing issues with gmp and other gnu licensed libraries. This is relevant for closed source projects where GPL requires that users be able to modify the library portions (and reclining to a modified dynamic library is essentially the only way). This should not be a problem for us because Acton itself is an open source repository so anyone can get the source directly.

plajjan:
We add the option to statically compile actonc by setting STATIC_ACTONC=true for the build. We set this variable for our build-debs jobs so the produced .deb file should now run on older platforms too!

Fixes #1146

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -156,7 +156,7 @@ jobs:
           apt-get install -qy libprotobuf-c-dev zlib1g-dev
           apt-get install -qy bash-completion build-essential debhelper devscripts
       - name: "Build Debian packages"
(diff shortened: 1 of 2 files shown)


plajjan:
Wow, it worked. A statically compiled actonc! merging...

PR #1169, opened 2023-02-08
Reorganize builtin

sydow:
Directory builtin reorganized to
- make as much use as possible of __builtin__.h  and .c generated by compiling __builtin__.act
- eliminating list_imppl.c, dict_impl.c and set_impl.c which are legacy file since the time before protocols.


sydow:
Files __builtin__.h and __builtin__.c are now to a large extent the ones generated by actonc when compiling __builtin__.act. Differences are marked in the files; things that will not be used are commented out but left in the file. This is mainly 

- everything having to do with Env, which is defined in __builtin__.act, but has separate env.h and env.c files in builtin (mainly generated by actonc at some earlier stage, so there are some differences compared to what we have commented out now)
- implementations in __builtin__.c of methods for builtin types which use struct fields which are not visible in __builtin__.act.

There are some other small things:
- some _new methods are not generated by actonc. I have manually added  B_HashableD_intG_new, B_ContainerD_listG_new and B_HashableD_strG_new, but there may be others that should be added but are not used in tests.
- B_print has a tuple as argument but the code generator generates code as if it had varargs parameters.
- The method B___init__() in __builtin__.c omits some field initialisations. I have not tracked all manual changes I made here.
- the test test/core_lang_auto/instantiate_abstract_class__bf.act should give a compilation error, but does not. I have introduced a syntax error so that ‘make test’ will succeed and I can merge with main…

plajjan:
Wow, good job! :tada:

PR #1188, opened 2023-03-03
Add DateTime, Timer, Duration & Instant

plajjan:
- [x] rename Timer -> Stopwatch
- [x] add standard, like UTC here on earth. could be something else on Mars...
- [x] remove mention of clock_id mapping to clock_gettime clock_id (it is different between macos & linux)
- [x] add warning about ntp_adjtime wrong info when NTPd not running
- [x] remove UnknownClock, it's really a UTC time but we have no clue about error etc
- [x] only comparable when clock is using same standard and is synced!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -691,6 +691,9 @@ dist/completion/acton.bash-completion: completion/acton.bash-completion
 dist/zig: deps/zig-$(ZIG_OS)-$(ZIG_ARCH)-$(ZIG_VERSION).tar.xz
 	mkdir -p $@
 	cd $@ && tar Jx --strip-components=1 -f ../../$^
+ifeq ($(shell uname -s),Darwin)
+	cp /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/timex.h $@/lib/libc/include/any-macos-any/sys/
+endif
 
 deps/zig-$(ZIG_OS)-$(ZIG_ARCH)-$(ZIG_VERSION).tar.xz:
 	curl -o $@ https://ziglang.org/download/$(ZIG_VERSION)/zig-$(ZIG_OS)-$(ZIG_ARCH)-$(ZIG_VERSION).tar.xz
--- a/rts/io.h
+++ b/rts/io.h
@@ -15,6 +15,7 @@
     #define IS_MACOS
 #endif
 
+extern uv_loop_t *aux_uv_loop;
 uv_loop_t *get_uv_loop();
 
 void alloc_buffer(uv_handle_t *handle, size_t size, uv_buf_t *buf);
--- a/rts/rts.c
+++ b/rts/rts.c
@@ -64,6 +64,7 @@ int return_val = 0;
 char *appname = NULL;
 pid_t pid;
 
+uv_loop_t *aux_uv_loop = NULL;
 uv_loop_t *uv_loops[MAX_WTHREADS];
 uv_async_t stop_ev[MAX_WTHREADS];
 uv_async_t wake_ev[MAX_WTHREADS];
@@ -2488,6 +2489,7 @@ int main(int argc, char **argv) {
             check_uv_fatal(uv_async_send(&wake_ev[i]), "Error sending initial work event: ");
         }
     }
+    aux_uv_loop = uv_loops[0];
 
     for (uint i=0; i <= MAX_WTHREADS; i++) {
         rqs[i].head = NULL;
@@ -2597,11 +2599,10 @@ int main(int argc, char **argv) {
     }
 
     // Run the timer queue and keep track of other periodic tasks
-    uv_loop_t *uv_loop = uv_loops[0];
     timer_ev = malloc(sizeof(uv_timer_t));
-    uv_timer_init(uv_loops[0], timer_ev);
+    uv_timer_init(aux_uv_loop, timer_ev);
     uv_timer_start(timer_ev, main_timer_cb, 0, 0);
-    int r = uv_run(uv_loop, UV_RUN_DEFAULT);
+    int r = uv_run(aux_uv_loop, UV_RUN_DEFAULT);
 
     // -- SHUTDOWN --
 
--- a/stdlib/src/time.act
+++ b/stdlib/src/time.act
@@ -1,19 +1,720 @@
-def monotonic() -> float:
+"""Stopwatch, time, dates and calendars
+
+There are primarily two uses of time
+- measuring elapsed time (e.g. for benchmarking)
+- telling the current date and time (e.g. for logging)
+
+Use a Stopwatch to measure elapsed time, e.g.:
+
+    s = time.Stopwatch()
+    # do something
+    elapsed = s.elapsed()
+    print("Elapsed time:", elapsed.str_us())
+    # Elapsed time: 1.234567 s
+
+Call `now()` to get the current date and time:
+
+    dt = time.now()
+    print("Current time:", dt)
+    # Current time: 2023-03-07T15:33:02.238251863+01
+
+`utcnow()` returns the current date and time in UTC:
+
+    dt = time.utcnow()
+    print("UTC time:", dt)
+    # UTC time: 2023-03-07T14:33:02.238251863Z
+
+Instant and Duration store time using two 64 bit fields for seconds and
+fractional seconds. It amounts to a range of +-292 billion years for Instant and
+an elapsed time of 585 billion years for Duration. The resolution is 1 attosecond.
+It's a lot of zeros, which is why you might want to use str_us(), str_ms() or
+str_ns() to get a more humanly readable representation.
+
+Information about the clock source is kept and used as part of various
+operations to indicate precision and error in resulting values.
+"""
+
+# TODO: localization?
+MONTHS = [
+    "January",
+    "February",
+    "March",
+    "April",
+    "May",
+    "June",
+    "July",
+    "August",
+    "September",
+    "October",
+    "November",
+    "December"
+]
+
+MNTHS = [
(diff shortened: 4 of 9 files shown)


plajjan:
@nordlander @sydow @ddjoh @aagapi I'd very much like feedback on this change. I'm trying to put together a time module that's correct (time is hard, ugh) yet simple to use. There are many things missing, but I'm trying to get the big pieces right here. I realize while writing this that I probably need to write up a proposal document that describes my thought process... I should totally do that, but perhaps there is something to be gleaned from the code? :)

plajjan:
These are some of my thoughts and considerations going into this
- design for correctness but also something that's easy to use
  - I like progression, like simple interface perhaps with implied defaults (reading your locale)!?, for doing common things, but if you need more flexibility, then you add more information (which you then have to understand, like different Calendar or computing between timezones)
  - this way, users are not forced to research all 13 arguments up front but can get a naive implementation right away in a simple way
- big question around split between wall clock and elapsed time, like this means users have to know up front if they are going to use a timestamp for logging or for measuring elapsed time and use the right functions
  - I've chosen a separation between wall clock time and elapsed time
- to correctly measure elapsed time, one should use a monotonic clock, clearly for wall clock one should read from ... the wall clock (real clock)
- while Python has monotonic time available (there is clock_gettime from C wrapped), I think a lot of Python code uses two datetimes (which are read from wall clock) to compute timedeltas. This is wrong and bad but it's how a lot of Python code looks, I think... I'd like to have an interface that encourages users to do the right thing
- most fast languages with high precision timing use monotonic time for measuring elapsed time and make a clear distinction between monotonically measured durations vs wall clock
  - clock_gettime in C with monotonic
  - Rust has Instant & Duration times which are only for measuring elapsed time, then separate datetime stuff for wall clock
  - Zig wraps it up in what they call a Timer, I think this is genius because it naturally leads users to use the right thing
  - Go has an interesting compromise where getting a Time gets you both a monotonic timestamp and wall clock time, for subtraction (i.e. computing duration) it compares the monotonic times only
- I've adopted the Timer idea from Zig. I think it's genius as it naturally leads users into using the right thing. What do you use to measure elapsed time? A timer of course!
- otherwise, the standard way is to get timestamp t1 and t2 then subtract to get the diff, which you can obviously do regardless if you grab those timestamps from a monotonic or real clock, but yeah, I think the Timer interface invites you to do the right thing
- calendars are complex, converting from a seconds-from-unix-epoch to year, month, day etc is a really hard problem and in particular if we want support for more than gregorian calendar
- Swift is very ambitious and gets this seemingly correct, however the interface isn't super inviting (or maybe I'm not used enough to Swift and their documentation - it looks messy)
- we have these more basic classes, like how to represent a point in time / timestamp? with a split between monotonic and real clock, should they be comparable? if we also keep track of which clock we got the timestamp from, we can make them not comparable
- keeping track of clock, not just by name but by some "id", we could also use this for detecting if an actor which has a timestamp is migrated
  - like if an actor takes monotonic timestamp t1 while running on RTS node 1, then gets migrated to RTS node 2 and takes monotonic timestamp t2, those are likely going to be wildly different so computing a duration from t2-t1 is going to be incorrect
  - for such a case, the real clock is actually better to produce the duration since they are going to be reasonable in sync between computers, but suddenly the precision falls from nanosecond level to probably around milliseconds
    - should such a comparison throw an exception? or do we signal the precision in the produced result? say a function was repeatedly called, it means you could get thousands of iterations where the timestamp is nanosecond precision, then there's an actor migration and suddenly you get a timestamp with millisecond precision, good or not?
      - a user that cares about precision could always inspect the result object, or we could perhaps have an argument to Timer saying that we want minimum X precision, so that normally it just gives you the best it can (which would be ~millisecond sometimes) but if you said "I want microsecond precision) then it will throw an exception after an actor migration... or maybe if we get a handle on how good our time source is (maybe all computers are PTP synchronized) then you'd get that precision
- the high level objects are different, like Timer vs DateTime, but the lower level, like Instant (which is what we call our timestamp / point in time) can be reused, but should we?
- Instant holds seconds as int64 and attoseconds as u64, gives us range of +-2800 billion years and attoseconds is 1*10^-18
- actually, all are `int` now, but we should change once i64 is a sub-type of int!?
- the wide range and high precision means these classes can be used for a wide range of computations, like short time scale scientific work as well as things on a very long time scale - pretty cool I think
- obviously comes at a cost in performance - we could offer lower precision classes as well for someone that cares about that, like `CoarseInstance`?
- to convert an Instant to a calendar time, we need to know according to what calendar, so we need calendar information
- Swift seems to do this very correctly and I feel our ambition should be on par, but for now we take shortcuts  (I added a Calendar class but it's empty)

Feel free to comment. If you think the whole thing is rubbish, feel free to start over and make a completely separate proposal :)

We are nowhere near Acton 1.0, so this is by no means frozen. We have plenty of opportunity to make changes, but naturally as we start using this module in other places, changing it is going to hurt, so better get the broad strokes correct at least :)

plajjan:
Also, the deadline to feedback on this proposal isn't really for this pull request. I'd like to merge something fairly soon as I have other modules that depend on this and I want something that works, but we can totally change all of this after this PR is merged too... and there's also a point in merging this because there are some bugs I'd like @nordlander to be able to easily work on which are exposed by this code.

nordlander:
I like the idea, separating durations from wall clock time is spot on. And the complexity of calendars can of course be dealt with in due time.

We might want to ponder a little more on the effect annotations, though. A function/method that measures its execution time can't be pure, so we should turn those into `mut` or `proc`. But a more philosophical issue concerns whether a computation without any links to the outside world should still be able to observe the speed of the platform on which it executes, and thus give different results for different runs despite having no environment dependencies. The pragmatist in me says "why not?", but I also see the risk that we might regret this someday when it stops some desireable compiler optimization from being applied. Anayway, there's nothing wrong in revisiting this issue if and when the need arises, so let's not complicate things in advance.

I can imagine that a `reset()` method could be useful in the Timer class, for the case where the time measurements should start at some specific event that's distinct from when the Timer object was created. And while we're on it, events could be measured even more precisely if we somehow could get hold of the _baseline_ that the rts maintains for each actor message. The baselines for external events are set as close to their actual occurrence as possible, so they are independent of the time the corresponding message has to wait before actually being scheduled. (This "platform independent timing" concept was actually one of the main cornerstones of Timber language I used to work on, together with an accompanying notion of event _deadlines_.)

nordlander:
I'll take a look at the bug that stops the outcommented code, of course.

plajjan:
@nordlander what do you think of Go's Time that stores both a monotonic time and the real clock? I'm sort of leaning towards that since it allows Timer() to work even after actor migration between machines. The result is much lower precision, but that could be OK (needs to be indicated though).

nordlander:
@plajjan That seems like a good idea, although I'm not really aware of the details.

plajjan:
@nordlander Go captures two timestamps (monotonic & real clock) in their Time structure so that when you do subtraction to get elapsed time that happens on the monotonic part etc. Quite clever. However, it doesn't go all the way for Acton since we can migrate actors. The monotonic clock does not have a known start time so it cannot be used for comparison with anything besides itself and only during the same boot. I have now implemented support for this in the Acton time module (I just pushed a new commit). An Instant is still only one measurement, but since we have the Timer class, that is a very natural place to do this, so it now holds both monotonic and real clock time. It will prefer using the monotonic time, when possible, but if that's incomparable it will fall back to real clock. It does this by track which Clock a Instant is from. The Clock has both a type (monotonic vs real and also "unknown" for naive timestamps) and an "instance id" where the instance (maybe it should be called `incarnation`?) is in practice the boot-id, like if we reboot a machine, this gets reset. On Linux we can get the boot_id while on other platforms (macos) we can't, so the monotonic clock will only be useful within the same Acton RTS invocation. I want to somehow expose the precision and accuracy of these clocks too...

I also added a PTimer which only relies on monotonic time for measuring elapsed time. If the monotonic clock is incomparable, it will instead fail, so for anyone writing precision sensitive applications this would be the way to go.

plajjan:
Timezones seems like a real spaghetti though, like all these localtime etc functions read the environment and crap. Rust had a CVE around this if someone was mucking with env vars while running localtime_r or somesuch. What a mess.

klambrec:
I've dealt quite a bit with timezone aware calculations on the wall clock in Python and found this module to be super logical as a user: https://pendulum.eustace.io/docs/

It defaults to UTC for DateTime, except when calling `now()` where it defaults to local time, can do conversions between time zones, TZ aware arithmetic, handy helpers on `Periods`/`Durations` etc..

plajjan:
@klambrec so I'm not really going much into various conversions, shifting etc right now. Although I've considered a lot of it, I've not spent any time on implementation. That'll come later. I did see pendulum and put it on the reading list for future time related work (#1190). I believe the proposal in this PR is essentially compatible with all those ideas and I think we share certain design traits, like having TZ aware everything per default (unlike Python which is naive in most cases) and making it easy to do the right thing. Pendulum really does nothing for measuring elapsed time, where I think we have a very strong story

plajjan:
I think the UnknownClock should be removed. It's really not "unknown", I gave the example of a timestamp read from a text file would be from an unknown clock, but that's not really true. What we mean by the realtime clock is a clock that is synchronized to the UTC standard. If it is not in sync it is as uncomparable as any other timestamp... what the "type" of the clock really means is which time standard it uses and having it be in sync is a criteria for it to be comparable... the question more becomes with what precision / error.

plajjan:
I am going to go ahead and merge this. We can revise things later on, there are some sort of unusual things about this design but I figure it's useful as is, so let's get it merged.

PR #1198, opened 2023-03-22
slightly improved some parser error messages

sydow:
This changes a very confusing error message reported by Kristian in issue 732 to a slightly less bad one...


plajjan:
Nice @sydow! :D

PR #1201, opened 2023-03-23
added builtin type u64

sydow:


plajjan:
Jesus that was fast!

Makes me want to request all the other ones, down to u16 😃 😬 For example we have the port argument to tcp connections, which is a u16 value. Isn't it nice to have that type than to have run time exceptions for out of bounds checks?

sydow:
Hej!

Well, I couldn’t help accepting the challenge, so I implemented i32, u32, i16 and u16. But I can’t merge it since this apparently became too much for arithmetic_divide which now times out after 60 seconds of unfinished compilation.

Should we temporarily remove that test?

Björn

PR #1206, opened 2023-03-26
added list methods clear, extend and pop

sydow:


sydow:
This adds list methods clear, extend and pop.

Methods index, count, remove and sort are *not* added since they place requirements on element type A (to implement Eq for three first, Ord for sort). We need to discuss how to handle this.

DestyNova:
Would it be possible to do a release with this change? `pop` is pretty handy!

plajjan:
Yes of course, we could cut a new release. There are some regressions in main compared to 0.14.2 but overall it's probably better. Still very early in the life of Acton as I'm sure you can appreciate.

I'm deadly curious what you are doing with Acton and how you found it? :)

DestyNova:
> Yes of course, we could cut a new release. There are some regressions in main compared to 0.14.2 but overall it's probably better. Still very early in the life of Acton as I'm sure you can appreciate.
> 
> I'm deadly curious what you are doing with Acton and how you found it? :)

I found it on [this fun benchmarking page](https://programming-language-benchmarks.vercel.app/acton) and thought it looked quite elegant, with a tiny bit of Elixir vibes (the distributed coordination RTS thing, not so much the language). Figured I'd find out more about it by trying to port a Python benchmark implementation to Acton, rather than actually reading the docs first 😃

plajjan:
@DestyNova v0.15.0 is built and released. The Homebrew build has failed for some reason, I guess that semi-automagic CI config has just broken down over time as it tends to do *sigh*

DestyNova:
Thanks @plajjan -- I was able to install the new release from the apt repo and it works fine on the minimised tutorial example I started with:
```python
actor main(env):
    l = ["foo", "foo"]
    print("Pop first item:", l.pop(0))
    await async env.exit(0)
```

plajjan:
@DestyNova note how there is a .Deb automatically built for tip releases as well (from main branch) so on debian you can just install that. We aim for an always working main branch so latest is pretty much the greatest! :)

PR #1211, opened 2023-03-27
added function sorted using timsort implementation from libxml2

sydow:
Don't know about licence issues for builtin/timsort.c

Note that this is not complete timsort; no galloping.

### DIFF
--- a/builtin/builtin.c
+++ b/builtin/builtin.c
@@ -37,6 +37,7 @@
 #include "Iterator.c"
 #include "slice.c"
 #include "hash.c"
+#include "timsort.c"
 #include "list.c"
 #include "dict.c"
 #include "str.c"
--- a/builtin/builtin_functions.c
+++ b/builtin/builtin_functions.c
@@ -237,7 +237,10 @@ $WORD B_min(B_Ord wit, B_Iterable wit2, $WORD iter, $WORD deflt) {
 }
  
 B_list B_sorted(B_Ord wit, B_Iterable wit2, $WORD iter) {
-    return NULL;
+    B_CollectionD_SequenceD_list w = B_CollectionD_SequenceD_listG_witness;
+    B_list res = w->$class->__fromiter__(w, wit2, iter);
+    B_tim_sort(wit, res->data, res->length);
+    return res;
 }
 
 // sum /////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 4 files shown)


plajjan:
Gosh, that was quick. Yer a wizzard Björn

I asked ChatGPT about galloping and it tells me that timsort without galloping still returns correct results, just slower. Cool cool! Nice work :smiley: 

Me:
> Is a timsort without galloping a complete implementation? What does galloping mean? Will it still return correct results and just be slower or will the lack of galloping mean it returns incorrect results?

ChatGPT:
> Timsort is a sorting algorithm that is designed to perform well on many kinds of real-world data, and it is widely used in many programming languages and libraries. Galloping is a technique used in Timsort to accelerate the sorting process by searching for a subsequence of elements in the input array that is already sorted, and then using binary search to find the correct position for the next element to be inserted.
> 
> While a Timsort implementation without galloping is technically complete, it may not perform as well as a version that includes galloping. Without galloping, the algorithm would still return correct results, but it may be slower on certain types of input data, especially when the input array is partially sorted or nearly sorted.
> 
> In general, galloping is a useful optimization for Timsort because it allows the algorithm to take advantage of existing order in the input data, reducing the number of comparisons needed to sort the array. However, it is possible to implement Timsort without galloping and still have correct results. The performance tradeoff between a Timsort implementation with and without galloping will depend on the characteristics of the input data, and it may be worth experimenting with both versions to determine which works best for a given use case.

sydow:
Hej!

ChatGPT is fascinating…

Yes, galloping is just an optimisation. TimSort is a variant of mergesort, so a central subroutine is to merge two already sorted subarrays to a longer subarray. This is simple: you just keep track of a position in each array and compare the values there. The smaller of the two values is moved to the result array and the corresponding position moved forward one step. So you move just one element for each comparison. A not uncommon situation in practice is that e.g. all the elements in the left subarray are smaller than the smallest  element in the right one. Then it is a pity to just move one element at a time. So you keep track of if the moved element comes from the left or the right. After, say, 10 elements in a row from the left you start galloping, i.e. moving the position to the left several steps in one go. If you arrive at an element which is still smaller than the current element to the right, you can now move several elements from the left to the result array. If not, you were unlucky and must go back. As long as galloping is successful you take longer and longer steps...

Björn

PR #1225, opened 2023-03-29
Use atomic mallocs

### LINKED ISSUE #1221: Use GC_MALLOC_ATOMIC
GC_MALLOC_ATOMIC marks the allocated memory as not containing any pointers, which means that libgc can skip scanning the internal part of an allocated block. We should be able to use GC_MALLOC_ATOMIC everywhere we know the contents of something is not a pointer, such as `str->val` and same for bytes etc.

plajjan:
Using GC_MALLOC_ATOMIC instead of GC_MALLOC (the normal malloc is redirected to GC_MALLOC in our repo) means we mark an allocation as atomic. Atomic allocations means the GC will not scan it internally for any pointers, which in turn speeds up the GC mark phase!

Fixes #1221.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,10 +230,10 @@ backend/test/skiplist_test: backend/test/skiplist_test.c backend/skiplist.c
 		$(LDLIBS)
 
 # /builtin ----------------------------------------------
-builtin/builtin_dev.o: builtin/builtin.c $(BUILTIN_HFILES) $(BUILTIN_CFILES) $(DEPSA)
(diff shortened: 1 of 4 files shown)


plajjan:
@sydow I have changed to GC_MALLOC_ATOMIC in a bunch of places, mostly in builtins, which is your turf. The test suite passes and I'm fairly certain this is correct but I think it would be good if you had a look at the code in this PR too and let me know if you think it looks good. I think there are a couple of more places where we can use GC_MALLOC_ATOMIC instead of malloc but I believe I have caught the most common ones in this PR. Feel free to use GC_MALLOC_ATOMIC in your code where suitable! :)

sydow:
This looks perfectly fine to me.

Björn

plajjan:
Cool. I'm merging then...

PR #1231, opened 2023-04-01
Enable incremental GC

plajjan:
This enables the incremental GC. It detects dirty pages and only scans those to shorten the time spent in the mark phase.


plajjan:
We don't build the GC like this anymore and I'm really not sure doing incremental GC, using this GC, is worth it at all... well, it doesn't apply cleanly anyway, so closing.

PR #1234, opened 2023-04-02
Use parallel GC

plajjan:


plajjan:
I've been running this locally now for two days without any problems, so I think it works good. Have also run this PR through CI 3 times without a hitch.

PR #1237, opened 2023-04-04
enabled list shrinking after del and pop when length goes below half …

sydow:
…capacity

Tested on acton-streamer with (very) small positive effect

### DIFF
--- a/builtin/list.c
+++ b/builtin/list.c
@@ -29,7 +29,18 @@ static void expand(B_list lst,int n) {
     }
     lst->data = newptr;
     lst->capacity = newcapacity;
-}  
+}
+
+static void shrink(B_list lst) {
+    if (lst->capacity > 20 && 2*lst->length < lst->capacity) {
+        int newcapacity = lst->length;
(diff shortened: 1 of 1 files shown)


plajjan:
Uhh, realize that the most notable effect should be in reduced memory usage not execution time... but it really does seem faster, I've run a couple of tests now and they're all around 3:06+-1 second. Maybe there are some lists which shrink and has positive effect on GC? Anyway, this is good so we're happy =)

sydow:
On 4 Apr 2023, at 17:11, Kristian Larsson ***@***.***> wrote:
> 
> 
> Uhh, realize that the most notable effect should be in reduced memory usage not execution time... but it really does seem faster, I've run a couple of tests now and they're all around 3:06+-1 second. Maybe there are some lists which shrink and has positive effect on GC? Anyway, this is good so we're happy =)
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you were mentioned.
> 

Yes, I think the gains are only in saved GC time. 

Björn

PR #1247, opened 2023-04-12
SNMP

plajjan:
Add libnetsnmp as a dependency. No code to use it just yet, but this should get us going. Might need to tweak build config later on...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -284,6 +284,9 @@ DEP_LIBS+=deps/instdir/lib/libmd.a
 endif
 
 DEP_LIBS+=deps/instdir/lib/libbsdnt.a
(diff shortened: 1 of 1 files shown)


plajjan:
Interesting, this actually built as-is. I was expecting it to fail on lack of OpenSSL dependency but I guess all systems, in this case our build machine, has OpenSSL libs installed already so it's a non-issue pulling those in. Maybe that's fine for now but going forward I think we want those pinned down too.

plajjan:
We now have support for package management, so we most definitely do not want to place SNMP in stdlib but as a separate package. Also, we rely on zig for C builds, so this is just wildly outdated. Closing.

PR #1249, opened 2023-04-16
Build projects using `zig build`

plajjan:
This switches to using zig build for building Acton projects. There are a number of advantages and unlocked potential advantages:
- zig has a very advanced caching system
  - invoking zig cc as a C compiler, like before, does not use this caching system
  - invoking zig build, like with this change, we make use of the cache!
- zig has awesome cross-compilation support and we can better utilize it when we call zigs high-level build command rather than the lower level cc compatibility commands
  - by directly adding in RTS, builtins etc, we can even skip building libActon libary archives in rel / dev flavors and let it be built per project with the exact compilation flags we want, or for cross-compilation! This could open up for more compilation options that affects builtins, RTS etc.
- build.zig is first compiled into a binary executable whish is then run to build the C parts of the Acton project
  - zig is sort of slow today, mostly due to LLVM, but it should be a lot faster in future versions
  - build.zig should not need to be recompiled unless the project changes, like new modules are added or removed
    - if a source file merely changes, build.zig still does not change

It's fairly simple: Write generated build.zig file and execute `zig build` in order to let zig build the C-parts of the project! We add two extra arguments, syspath and reldev which are passed in from actonc to zig build as to convey where acton is installed and if we are building for dev or release.

To illustrate the caching, including caching of the build.zig executable, the first build is rather slow while rerunning actonc build is rather quick:
```sh
  kll@ThinkYoga:~/terastream/acton/hello$ ../dist/bin/actonc build
  Building project in /home/kll/terastream/acton/hello
    Compiling hello.act for release
     Finished compilation in   0.040 s
    Final compilation step
     Finished final compilation step in   6.593 s
  kll@ThinkYoga:~/terastream/acton/hello$ ../dist/bin/actonc build
  Building project in /home/kll/terastream/acton/hello
    Compiling hello.act for release
     Finished compilation in   0.067 s
    Final compilation step
     Finished final compilation step in   0.059 s
  kll@ThinkYoga:~/terastream/acton/hello$
```
The vast majority of time is spent building build.zig into a program. Once this is done, it's fast. Modifications to individual .act files does not affect build.zig, i.e. we don't need to rebuild the build program, so it is fast. Since it is also caching, it ends up being much faster than gcc or similar both during the individual module compilation and for the final compilation of binary executables, in a small example project, we saved ~50% going from ~1s to ~0.5s (on a slow laptop).

If we add or remove a module in a project, the build.zig program needs to be recompiled, which takes considerable time:
```sh
  kll@ThinkYoga:~/terastream/acton/hello$ ../dist/bin/actonc build
  Building project in /home/kll/terastream/acton/hello
    Compiling hello.act for release
     Finished compilation in   0.066 s
    Compiling foo.act for release
     Finished compilation in   0.023 s
    Final compilation step
     Finished final compilation step in   6.653 s
  kll@ThinkYoga:~/terastream/acton/hello$
```
This should be improved with future versions of zig. Smart people are working hard on the problem.

The long build times doesn't have that much impact on Acton projects since it's mostly just on the initial build. However, for non-project builds, like compilation of individual .act files, the hit of using zig build is considerable. To avoid this, zig build is only used for projects and traditional build is used for non-projects! It is possible to force one or the other using --zigbuild or --no-zigbuild.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@ backend/test/queue_unit_tests
 backend/test/skiplist_test
 backend/test/test_client
 
+builder/zig-cache
+
 compiler/actonc
 compiler/package.yaml
 compiler/tests/*.ty
--- a/Makefile
+++ b/Makefile
@@ -222,6 +222,10 @@ backend/test/skiplist_test: backend/test/skiplist_test.c backend/skiplist.c
 	$(CC) -o$@ $^ $(CFLAGS_DB) -Ibackend \
 		$(LDLIBS)
 
+builder/builder: builder/build.zig $(ZIG)
+	rm -rf builder/zig-cache builder/zig-out
+	cd builder && ../$(ZIG)/zig build && find zig-cache -name build -exec cp {} builder \;
+
 # /builtin ----------------------------------------------
 builtin/__builtin__.c builtin/__builtin__.h: builtin/ty/out/types/__builtin__.ty
 builtin/builtin_dev.o: builtin/builtin.c builtin/__builtin__.h $(BUILTIN_HFILES) $(BUILTIN_CFILES) $(DEPSA) $(LIBGC)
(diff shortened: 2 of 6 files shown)


plajjan:
Woah. Finally working and it's working well. Couple of hacks or things we could clean up in there still but this is a solid start so let's go!

PR #1260, opened 2023-05-02
Auto   builtin  

sydow:
This branch generates the files builtin/__builtin__.* by compiling __builtin__.act.


plajjan:
@sydow build is failing because `dist/builtin/__builtin__.h` does not exist. It is populated by copying `BUILTIN_HFILES`, which in turn is populated by `$(wildcard builtin/*.h)`. At the time that expression is evaluated in the first run of make in an empty repository, we have not yet generated `builtin/__builtin__.h` and so it is not in `BUILTIN_HFILES`.

I would propose simply appending them statically, like so:
```makefile
BUILTIN_HFILES=$(wildcard builtin/*.h) builtin/__builtin__.h
BUILTIN_CFILES=$(wildcard builtin/*.c) builtin/__builtin__.c
```

The make rules around builtin/__builtin__.c are not ideal but given the multi-file output from actonc, there is no perfect way, so this is fine for now.

sydow:
@plajjan After having fixed the Makefile as you suggested, this fails again on all Linux machines while it succeeds on Mac. Since I only have access to a Mac, I don't know what to do. In addition, the error messages from the Linux failures are very unhelpful.

plajjan:
@sydow it's not actually macos vs Linux, it is parallel build (which we do on Linux) vs serial build (which we do on Mac due to some other weird bug somewhere).

In the serial build it just happens so that builtin/__builtin__.h is generated before other files that need it are built. When using a concurrent build, we start building some stuff that depend on __builtin__.h but before it is built. The error message is in there, but since it's a concurrent build, it isn't at the end but embedded a little higher up in the output. The root cause is of course proper declaration of dependencies (as is so often the case with broken Makefiles). I will submit a patch.

plajjan:
@sydow I've fixed the concurrent build stuff and it's passing CI now on all platforms!

sydow:
Tack!

Björn

PR #1261, opened 2023-05-03
Enable concurrent builds in CI for MacOS

### LINKED ISSUE #1147: Investigate problems with concurrent builds on mac
Not sure why we saw these problems pop up in the first place, but removing concurrency from the macos builds makes them pass CI tests, so that's what I did in #1145.

However, we naturally want concurrency, so we need to re-enable it, which means we need to figure out the reason why it failed in the first place... alternatively, we could perhaps just attempt random stack upgrades or similar to "workaround" the problem.

plajjan:
We had some issues with concurrent builds specifically on MacOS that we could never really pinpoint. They appeared out of thin air one day, perhaps they've also gone away by themselves...?

Fixes #1147

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,7 +61,7 @@ jobs:
       - name: "Build Acton"
         run: |
           export ZIG_LOCAL_CACHE_DIR=~/zig-cache
-          make -C ${{ github.workspace }} BUILD_RELEASE=${{ env.BUILD_RELEASE }}
+          make -j2 -C ${{ github.workspace }} BUILD_RELEASE=${{ env.BUILD_RELEASE }}
       - name: "Build a release"
         run: make -C ${{ github.workspace }} release
       - name: "Upload artifact"


plajjan:
I've rerun this a bunch of times and while there have been some failures they have all been unrelated to the concurrent build, like just previously seen known errors.

However, I also cleared the build cache and reran the build, after which all the macos builds fails, thus it is truly broken. There is some part in the build, that is being cached, that does not work great when run concurrently. 

```
2023-05-04T11:01:10.5848390Z   CC       src/libpcre2_posix_la-pcre2posix.lo
2023-05-04T11:01:10.9824310Z   CC       src/pcre2posix_test-pcre2posix_test.o
2023-05-04T11:01:11.1755080Z   CCLD     libpcre2-8.la
2023-05-04T11:01:11.4563780Z ghc-cabal:
2023-05-04T11:01:11.4664320Z '/Users/runner/.stack/programs/x86_64-osx/ghc-8.10.7/lib/ghc-8.10.7/bin/ghc'
2023-05-04T11:01:11.4760570Z exited with an error:
2023-05-04T11:01:11.4862050Z Missing file:
2023-05-04T11:01:11.4948920Z /Users/runner/.stack/programs/x86_64-osx/ghc-8.10.7/lib/ghc-8.10.7/settings
...
```

so ghc-cabal is complaining about missing a file. And we do cache .stack dirs, so it's quite likely that this is what makes a difference here. But what does it fail when there is no cache? Are we trying to run multiple cabal concurrently and it is messing stuff up? cabal should be run implicitly from stack I think, which is run when compiling the compiler. While many other targets depend on the compiler I don't see how they would cause multiple concurrent builds since Make should be clever enough to figure out the dep and only run it once!?

We silence a bunch of the output in the actonc rule, which doesn't make debugging easier either.. not quite sure how to proceed. We could perhaps serialize dependency install, but we don't actually want that, like we want to run stack install concurrently with the rest of the build (building our C deps take a long time).

PR #1274, opened 2023-05-17
Added some more minor code generation improvements

sydow:


plajjan:
This fails on Linux with 
```
```

this comes from when we run `zig cc` using readCreateProcessWithExitCode. LANG is not set in env, so defaults to LANG=C I think. I suppose the output from the process we run might not be utf8. We don't have control over the file handle here, so can't force an encoding anyway.

I wrote a patch (#1276) to change the way we run processes, gaining control over the pipes and it indeed fixes this problem (although I don't quite understand why, because it works even without forcing an encoding, the refactoring alone is enough)... but that introduces other problems, now there are tests that time out. Very strange.

sydow:
Hi,

I saw that. The fault is with builtin_auto/builtin_functions_test.act, which does indeed contain a few non-ASCII  characters in UTF8. I note also that it is actonc which complains when it reads the file? But the most surprising thing is that it appears now. This test has been unchanged for a long time and I don’t think that anything in the pull request has anything to do with this. Could it have anything to do with a new zig? But I take it that the main branch still passes CI.

Björn

plajjan:
@sydow no, it is not when actonc directly reads the file. We have our own readFile function that I wrote awhile back that forces utf8 encoding when we read .act files. For .ty files we have our own readFile which doesn't enforce an encoding, but this has never been a problem, besides the files are rather binary, so perhaps we should do openBinaryFile instead of openFile?

But anyway, the hGetContents that fails is called from within the System.Process library. We call readProcessWithExitCode. It runs a process and sets up pipes to read that process stdout and stderr. It is somewhere there that it fails with this invalid byte sequence.

So whatever process we are running must output something to stdout or stderr that it doesn't like, perhaps a warning and then it includes part of the source file, thus we see the invalid byte sequence and error out. Perhaps your changes to CodeGen now outputs code that cc warns about, includes a line of the source code in the warning (does it ever do that?) and thus the error?

A workaround would be to set a locale in CI, like `LANG=en_US.UTF-8` and defer a more proper fix for later

plajjan:
I pushed a commit to set LANG and LC_ALL but it didn't make a difference, so back to the drawing board...

I did make some progress in my other branch (#1276) to entirely rework how we run subprocesses and got it running locally. It was like my 3rd attempt and the earlier ones failed and deadlocked but while developing this 3rd one I realized that I had a an actual block, like I was actually waiting on zig which in turn was trying to acquire a lock for building a file. The lock was held by another zig process from an earlier version of my code which was blocked on my buggy code in actonc :P so after `killall actonc` this 3rd version worked alright, but my 2nd attempt and perhaps even 1st attempt could also potentially have worked :P

Anyhow, this 3rd version works also in CI except for the brew test where it reproducibly hangs and brew test times out after 5 minutes. I can't figure out why. Not possible to reproduce locally (on macos) and debugging in CI is hell.

My latest attempt is now in #1277 and it appears to pass. It's treating stdout & stderr output as bytestrings as to avoid any encoding issues altogether. Will rerun it a couple of times just to make sure but does look like we have a working path forward there.. now to check if it makes this branch pass...

plajjan:
I did 3 reruns of #1277 and they all passed CI just fine, so I think I've solved the process running stuff there now. It's merged to main and I have now rebased this branch on main. CI running now... will hopefully pass :)

sydow:
Kristian, you’re a phenomenon.

PR #1276, opened 2023-05-18
Run CC process assuming utf8 on stdout/err

plajjan:
With readCreateProcessWithExitCode() as we previously used it, we got an error while compiling test/builtin_auto/builtin_functions_test.act:

  ...

The problem was noticed in CI for a new branch, which does not modify these parts of the compiler, yet fails on an existing test. It is reproducible on a Linux machine when LANG is not set (CI does not set LANG), so the system defaults to LANG=C (I think). The hGetContents is not called directly from our code (we enforce encoding where we do), but it must be called from within readCreateProcessWithExitCode() from the standard System.Process library. We have no way of modifying it. It's puzzling because the output of the commands run do not appear to contain anything but US ASCII, so it's unclear how we can get an invalid byte sequence. And why is this only happening in a branch with seemingly unrelated changes?

I refactored the code to use withCreateProcess context instead, which allows more direct control over the stdout & stderr pipes. I now force utf8 encoding but it should be noted that I got all of this to work *before* forcing encoding. Simply refactoring and running hGetContents ourselves made it work. I do not understand why.

Running CC is also moved to a separate function that can be called from the two places where we run CC. Less code!

Also, while we now enforce encoding, this is just enforcing our assumed encoding when reading. The processes we run could be outputting something else. I presume they honor locale, so it might in fact not be UTF-8 now. But can we influence that in a portable way? I don't see how, setting LANG=en_US.UTF-8 won't work on a machine with sv_SE.UTF-8 and it won't work on another OS?

### DIFF
--- a/compiler/ActonCompiler.hs
+++ b/compiler/ActonCompiler.hs
@@ -36,7 +36,11 @@ import qualified Pretty
 import qualified InterfaceFiles
 
 import Control.Applicative
+import Control.Concurrent.Async
+import Control.Concurrent.MVar
 import Control.Exception (throw,catch,displayException,IOException,ErrorCall)
+import Control.Exception (bracketOnError)
+import Control.Concurrent (forkIO)
 import Control.Monad
 import Data.List.Split
 import Data.Monoid ((<>))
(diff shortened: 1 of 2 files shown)


plajjan:
This does not pass the brew test for some reason. I can't figure it out.

I have another solution in #1277 that does work and I think it's a better solution. Instead of forcing utf8 encoding, we assume the output is bytestrings and just pass it through. It should be safer in all situations as we don't really need to parse it as text in actonc anyway.

Closing as #1277 is better.

PR #1281, opened 2023-05-19
Download deps archive, fallback to building it

plajjan:
Since there is now a cached version of the deps archive libActonDeps uploaded as a release on the main acton repo, we can now attempt downloading it first and only if that fails do we need to build the archive locally.


plajjan:
Closing in preference of #1284 that's already merged!

PR #1289, opened 2023-05-22
rm segfault workaround in example

DestyNova:
The comment in the first actors example mentions that a dummy return was needed to avoid a segfault with "await async". This doesn't seem to be true anymore -- at least, the example doesn't segfault and produces the same output for me either way (running on Pop!_OS 22.04 Linux).


plajjan:
@DestyNova thanks!

CI is failing due to unrelated things so I can't just merge this. We get spurious segfaults, which is from a change I merged yesterday, see #1290. The other is because we currently presume CI is running for this repo, opened #1291 for that. Regardless, nothing you need to think of - I'll fix and thanks again :)

plajjan:
@DestyNova I just merged #1292 which addresses the brew-macos problem in #1291. You should now be able to submit PRs to this repo and have CI run successfully for them. It naturally won't magically update this particular PR since your branch is based before the fix in #1292.

DestyNova:
Oh I need to update the branch, of course.

plajjan:
This is merged through #1303 which is (re)based on main after #1302 to avoid the compilation problems described by #1290. Thanks again @DestyNova

PR #1295, opened 2023-05-23
Cache the debian build dir

plajjan:


plajjan:
This doesn't work. build-debs takes ~12 minutes with caching while without caching it runs in ~10 minutes. The 2 extra minutes are related to cache restore etc.

Not sure why it cannot use a cache, it looks like I get speedup when I build locally but I guess it's the standard mtime of git vs other things problem.

Giving up on this for now...

PR #1299, opened 2023-05-23
Run tar without -v

plajjan:
The acton distribution contains > 15k files so this might take noticeable amount of time to print.


plajjan:
It's pretty much not noticeable, so I'm going to keep the -v for better debug output.

PR #1307, opened 2023-05-30
changed module math to regular stdlib form

sydow:


plajjan:
@sydow nice! I'd like to wait with merging this as I just want to make sure the changes work together with my new branch. Marking this as a draft meanwhile to avoid accidental merge.

plajjan:
Hmm, normally with .ext.c style we don't have a .h file at all, we let actonc generate it. Can we remove it from here too?

sydow:
As I tried to explain in a mail to Kristian, the reason that there is still a math.h in stdlib/src is that numpy.h in the same directory needs to include math.h and I just don’t know what path to write in numpy.h to refer to the generated math.h (which is placed in stdlib/out/types and later also somewhere in dist). As soon as the correct path is added in numpy.h we can and should remove stdlib/src/math.h.

This relates to the problem with paths in our system that Kristian mentioned in today’ meeting and I am sure that Kristian can say immediately which path to use.

plajjan:
@sydow ahh ok, I did read that part in your email but interpreted it differently, thinking it referreed to the includes of numpy that goes to like ../c_src/numpy etc. But OK.

No worries, I have fixed this already as I'm incorporating your fixes in this branch in my branch #1309! We can actually close this PR since #1309 subsumes it.

PR #1328, opened 2023-06-04
Add OpenSSL

plajjan:


plajjan:
This is based on the build.zig file in https://github.com/kassane/openssl-zig. It is incomplete. It only includes a subset of C source files from the openssl library. It has a hardcoded version of the main openssl.h file. Hardcoded, what does that mean? In a "normal" build of OpenSSL, openssl.h is generated from openssl.h.in. It is generated by some custom built code generating tool, written in Perl. It is also called by a custom build system that seems specific to the OpenSSL repo. Very very frustrating trying to rewrite this with the Zig build system, so I've essentially given up. Thus I will close this PR as there is nothing actionable left to do.

My main focus right now is using MbedTLS which seems a) simple and b) good enough. If we want something /better/ later on, I suspect BoringSSL might be it. It's a fork from OpenSSL by Google who use BoringSSL in all their production code. Pretty good testament. BoringSSL seems to use CMake, so it more standardized in that sense.

PR #1369, opened 2023-06-19
Use xz instead of bzip2 compression

plajjan:
xz is faster and better than bzip2. Here we tell it to use lowest level of compression (-0) and thus the fastest mode. --threads=0 enables using multiple threads to compress data.

Despite being the faster level and thus worst compression, it beats bzip2 standard settings. The current release tar ball turns in bz2 & xz:

- bz2: 66061362
- xz:  65009672

On an AMD 5950X (16 cores / 32 threads) with an NVMe drive, it takes ~16 seconds to produce the bz2 while producing the xz takes just over 1 second. It might just be the increased parallelism or a combination with xz being faster overall.

There are plenty of CI runs where it takes just as long to build Acton, like coming in just over 2 minutes (with a zig build cache), as it does to produce the release tar ball (just under 2 minutes). Hopefully this can minimize that time.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -210,7 +210,7 @@ jobs:
           name: acton-macos-11
       - name: "Extract acton"
         run: |
-          tar jxvf $(ls acton-darwin*.tar.bz2 | tail -n1)
+          tar Jxvf $(ls acton-darwin*.tar.xz | tail -n1)
       - name: "Compile acton program"
         run: |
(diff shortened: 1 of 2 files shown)


plajjan:
Going through all the tests it seems xz is indeed faster. There's a hell of a variance between test runs, presumably because we get wildly random test runners assigned and some are very constrained. But again, overall seems faster, so merging.

PR #1403, opened 2023-07-16
Document the --sigs actonc option

mzagozen:
I think it fits in well with the manual type hints example.

### DIFF
--- a/docs/acton-by-example/src/types.md
+++ b/docs/acton-by-example/src/types.md
@@ -2,7 +2,7 @@
 
 Acton is a statically typed language. Unlike traditional languages like C or Java, where the type of every variable must be explicitly stated, we can write most Acton programs without types. The Acton compiler features a powerful type inferencer which will then infer the type.
 
-To see the inferred type signatures after an Acton program is compiled you can use the `--sigs` option of the compiler. As rhe name suggests, this will print out the type signatures for the functions, actors and their attributes and methods found in the compiled file.
+To see the inferred types of an Acton program, use the `--sigs` option of the compiler. As the name suggests, this will print out the type signatures for functions, actors and their attributes and methods in the specified Acton module.
 
 Source:
 ```python
@@ -18,12 +18,9 @@ actor main(env):
     await async env.exit(0)
 ```
 
-Compile and use the `--sigs` option:
+Print type signatures with `--sigs`:
 ```sh
-actonc --sigs types.act
-Building file types.act
-  Compiling types.act for release
-
+actonc types.act --sigs
 
 #################################### sigs:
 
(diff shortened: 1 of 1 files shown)


docs/acton-by-example/src/types.md
```diff
@@ -2,6 +2,40 @@
 
 Acton is a statically typed language. Unlike traditional languages like C or Java, where the type of every variable must be explicitly stated, we can write most Acton programs without types. The Acton compiler features a powerful type inferencer which will then infer the type.
 
+To see the inferred type signatures after an Acton program is compiled you can use the `--sigs` option of the compiler. As rhe name suggests, this will print out the type signatures for the functions, actors and their attributes and methods found in the compiled file.
```
plajjan, line 5: I want to place less emphasis on the built / compiled part. It used to be that actonc would compile a source file no matter what but now we check if alternate output is enabled, such as --sigs, and then exit before doing final compilation. I think we still write a .ty file to disc. The point is that the focus is printing some output, not compiling, so rewording to reflect that...

```suggestion
To see the inferred types of an Acton program, use the `--sigs` option of the compiler. As the name suggests, this will print out the type signatures for functions, actors and their attributes and methods in the specified Acton module.
```

docs/acton-by-example/src/types.md
```diff
@@ -2,6 +2,40 @@
...
+    await async env.exit(0)
+```
+
+Compile and use the `--sigs` option:
+```sh
+actonc --sigs types.act
+Building file types.act
+  Compiling types.act for release
+
+
```
plajjan, line 27:
```suggestion
actonc types.act --sigs

```

Freshest actonc doesn't print the "Building file" cruft for alternate outputs like --sigs...

Also provide file first. Actonc has multiple commands and a file can be thought of as an alias for "just build this file". --sigs is relevant for a single file, but it is not relevant to for example `actonc new my-new-project`. We should encourage users to think of "first command then followed by relevant arguments"

docs/acton-by-example/src/types.md
```diff
@@ -2,6 +2,40 @@
...
+actor main(env):
+    i1 = 1234
+    s1 = "hello"
+    foo(i1, s1)
+
+    await async env.exit(0)
+```
+
+Compile and use the `--sigs` option:
```
plajjan, line 21:
```suggestion
Print type signatures with `--sigs`:
```

plajjan (changes requested):

plajjan (approved):

plajjan:
I was about to merge this when I realized it sort of breaks the format I've tried to keep for AbE thus far. The idea is to have one example per page that explains as few concepts as possible. There is a run-example.py script that can actually run the examples. It relies on simple regex hacks to replace the output. It won't work if there are multiple examples in the same page but I think it's also good that each page is really short. Explain one thing!

Thus, while I think the content is good (after the fixes I merged, heh), I believe it needs to move to its own page. Maybe show types using --sigs is a better example to use first and it is the type hints parts that should be moved to a separate page. Feel free to rearrange to your liking (well, as long as abiding by the one-example-per-page rule)! :)

plajjan:
I've opened #1405 with the changes I described above. Closing this.

PR #1481, opened 2023-09-09
Print unhandled exceptions

### LINKED ISSUE #1475: Print exceptions by default
Unhandled exceptions are currently silently discarded which makes it rather hard to realize something even went wrong. Shouldn't we encourage that all exceptions are caught? For most functions this is prolly not that hard by just having rather wide try/except clauses.

We've talked in the past about being able to install some default exception handler on an actor. We could of course dig deeper into that, but I think the first steps are:
- print exceptions by default
- thus "forcing" users to handle all exceptions (but again, at a high scope is fine)

plajjan:
Rather than hiding them away in the RTS debug output, we print them to stderr per default.

Also added a very simple version of name unmangling to get at least the most basic of examples working with better readability.

With this PR this code:
```python
actor main(env):
    raise ValueError("foo")
    env.exit(0)
```
Looks like:
```
Unhandled exception in actor: -12[foo.main]:
  ValueError: foo
```

The indentation of the exception itself is intentional. I was thinking we would make it easier to parse these tracebacks, which are log multi-line log messages, by indenting the subsequent lines.

Fixes #1475

### DIFF
--- a/base/rts/rts.c
+++ b/base/rts/rts.c
@@ -1448,6 +1448,35 @@ void wt_wake_cb(uv_async_t *ev) {
     // work is run later when wt_work_cb is called as part of the "check" phase.
 }
 
+char *unmangle_name(char *input) {
+    // Check for NULL input
+    if (input == NULL) {
+        return NULL;
+    }
+    char *output = malloc(strlen(input) + 1);
+    // Check for allocation failure
+    if (output == NULL) {
+        return NULL;
+    }
+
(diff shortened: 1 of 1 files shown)


plajjan:
@nordlander WDYT?

plajjan:
I'm going to go ahead an merge this. I think it's the right move. We can discuss in the meeting and adjust if necessary. Partly doing this since I want the unmangle_name() for my logging stuff.

PR #1516, opened 2023-09-25
Build on Debian:12 / bookworm

### LINKED ISSUE #1512: Can we run build-debs on newer debian?
There seems to be a lot of intermittent failures in the build-debs job. Some issue downloading haskell stuff from fpcomplete.com. Some issues on the Internet say it might have to do with an old version of stack. We run on debian:bullseye and it has stack 2.3.3 while 2.7.5 and newer should work better.

Newer versions of debian have a newer stack, so that's a natural upgrade path.

We build actonc statically and this works well on debian:bullseye. I believe I have tried repeating this on newer debian and had some problems, so we need to make sure this part works correctly.

An alternative solution could be to just upgrade stack but that feels rather dirty to do in a .deb build, like is that really reproducible? and surely it breaks debian conventions, right?

plajjan:
Fixes #1512

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -150,7 +150,7 @@ jobs:
   build-debs:
     runs-on: ubuntu-latest
     container:
-      image: debian:bullseye
+      image: debian:12
     steps:
       - name: "Show platform and environment"
         run: |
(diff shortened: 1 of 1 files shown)


plajjan:
Superseded by #1517. Closing.

PR #1517, opened 2023-09-25
Upgrade to Stack LTS 21.13 & GHC 9.4, build on Debian 12 and always do static compilation for Linux

### LINKED ISSUE #1512: Can we run build-debs on newer debian?

plajjan:
It's been awhile since we upgraded and it would be nice to move to a fresher version of ghc, so here we go.

Now doing CI .deb builds using Debian 12 / bookworm. Since our newer Stack LTS requires a newer version of Stack, we have to upgrade stack on Debian 11 and Ubuntu 20.04 and 22.04. It's a little nasty but I think it's ok to say that people who want to develop Acton itself needs to use a relatively recent platform.

The LTS & GHC upgrade gives a good speed increase. Compiling a certain relatively large project now runs ~25% faster or so.

I managed to put together so we can now do static compilation on Linux even when building on other distributions than Debian 11.

Fixes #1512.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -183,7 +183,7 @@ jobs:
       - name: "Build Debian packages"
         run: |
           export ZIG_LOCAL_CACHE_DIR=~/zig-cache
-          make -C ${GITHUB_WORKSPACE} debs BUILD_RELEASE=${{ env.BUILD_RELEASE }} STATIC_ACTONC=true
+          make -C ${GITHUB_WORKSPACE} debs BUILD_RELEASE=${{ env.BUILD_RELEASE }}
       - name: "Compute variables"
         id: vars
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -150,7 +150,7 @@ jobs:
   build-debs:
     runs-on: ubuntu-latest
     container:
-      image: debian:bullseye
+      image: debian:12
     steps:
       - name: "Show platform and environment"
         run: |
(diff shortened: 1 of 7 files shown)


plajjan:
This fails to build with homebrew because it has GHC 9.4.4 while Stack LTS 21.13 requires GHC 9.4.7. There was a PR (https://github.com/Homebrew/homebrew-core/pull/142934) for upgrading brew to 9.4.7 but it was closed. Seemingly because 9.6 will soon be available. I find that a little weird, like why not maintain 9.4 too? 9.2 is seemingly better maintained where it's at 9.2.8. There is a stack LTS for 9.2.8 so that would be an option!

plajjan:
https://github.com/Homebrew/homebrew-core/pull/141617 is the WIP PR which brings in both 9.6.2 and 9.4.7 (also fixing the build for Sonoma)

plajjan:
The PR for GHC 9.4.7 has been merged and I've locally verified that building the Homebrew formula for Acton using this branch now works correctly so going to go ahead and merge this!

PR #1545, opened 2023-10-09
Additional arithmetic and comparison for time.Instant

ddjoh:
- add/sub duration
- comparison

### DIFF
--- a/base/src/time.act
+++ b/base/src/time.act
@@ -455,6 +455,14 @@ class Instant:
 #        else:
 #            return Duration(other.second - self.second, other.attosecond - self.attosecond)
 
+    def add(self, duration: Duration) -> Instant:
+        attosecond_sum = self.attosecond + duration.attosecond
+        return Instant(self.second + duration.second + attosecond_sum // 10**18, attosecond_sum % 10**18, self.clock)
+
+    def sub(self, duration: Duration) -> Instant:
+        attosecond_diff = self.attosecond - duration.attosecond
+        return Instant(self.second - duration.second + attosecond_diff // 10**18, attosecond_diff % 10**18, self.clock)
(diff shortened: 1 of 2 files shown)


plajjan:
Nice, LGTM.

FYI, the idiomatic way to write unit tests is to define test functions in the module like (just like what you've done):

```python
def _test_the_thing() -> None:
    testing.assertEqual(expected, do_something(), "do something did something wrong")
```

The idea is to let the compiler do test discovery, so it will place all functions that matches the name prefix `_test` and has the right signature into `__unit_tests`. Similarly it will create a separate entry point for testing. Since we haven't actual implemented this part yet, I've written it manually like in test_re.act (https://github.com/actonlang/acton/blob/main/test/stdlib_auto/test_re.act#L49)
```python
__unit_tests: dict[str, testing.UnitTest] = {
    "_test_basic": testing.UnitTest(_test_basic, "basic", "Basic regexp matching"),
    "_test_group": testing.UnitTest(_test_group, "group", "Regexp matching with groups"),
    "_test_named_groups": testing.UnitTest(_test_named_groups, "named_groups", "Regexp matching with named groups"),
    "_test_multi_groups": testing.UnitTest(_test_multi_groups, "multi_groups", "Regexp matching with multiple groups"),
    "_test_matches": testing.UnitTest(_test_matches, "matches", "Regexp finding all matches"),
}

actor main(env):
    testing.unit_test_runner(env, __unit_tests)
```

The `testing.unit_test_runner` then runs the tests (in parallel!) and prints pretty output :)

There are more changes, so like this is bound to change as well making it hard to talk about an "idiomatic style" when it is in flux :stuck_out_tongue_closed_eyes:

Issue #1556, opened 2023-10-13
Print source location for NoInfo errors


plajjan:
@sydow I was under the impression that with the improved error message thingy we would always be able to get some source location that would be sort of relevant. Isn't that the case?

I notice it is not currently printed. Can we print it? If there are multiple locations then just print them all?

Here's an example:
```python
actor main(env):
    def test():
        foo: ?list[int] = None
        foo = [1, 2, 3]
        print(foo)
    test()
    env.exit(0)
```
resulting in this error NoInfo 86 without any source code location
```
kll@Boxy:~/terastream/acton/foo$ actonc src/foo.act 
Building project in /home/kll/terastream/acton/foo
  Compiling foo.act for release

ERROR: Error when compiling foo module: Type error
NoInfo 86
kll@Boxy:~/terastream/acton/foo$ 
```

This is what we used to get:
```
kll@Boxy:~/terastream/acton/foo$ /usr/bin/actonc src/foo.act 
Building project in /home/kll/terastream/acton/foo
  Compiling foo.act for release

ERROR: Error when compiling foo module: Type error
 5:13-18
  |
5 |        foo.append(4)
  |            ^^^^^^
Cannot infer W_18 : ?__builtin__.list[__builtin__.int].append < T_19
```

sydow:
Well, what needs to be done for each NoInfo is to replace it with another form of ErrInfo, namely Origin SrcLoc String…

But, of course NoInfo should also take a SrcLoc as an argument; I can’t explain why I didn’t think of that :-(. I will do add that ASAP, but probably not today. 

Unfortunately, I don’t think a location is immediately available for all NoInfo occurrences. We need to have some piece of source code to get the SrcLoc from and that is not always at hand. We will, I think, rewrite code to carry such source code pieces around, but that is a bigger job. So what I will do first is to add a SrcLoc to NoInfo but sometimes the location will be NoLoc…

After trying more examples I think that we should invoke some AI :-) when we get a type error, in particular when we get “Cannot solve" errors. Examples of what I think of is
- if an actor has a parameter called env, guess that it is of type Env.
- in an assignment such as
lst = [],
guess that lst has type [_] (and not a subtype). 

Björn

sydow:
fixed by #1557

PR #1557, opened 2023-10-13
Added SrcLoc to NoInfo constructor of ErrInfo

sydow:


sydow:
Fixes #1556 but with some shortcomings:
- some occurrences have location NoLoc, since no source code location is available in the code as is.
- some SrcLOc assignments are dubious, based on incomplete understanding on my part.

Issue #1558, opened 2023-10-13
Non-exhaustive patterns error when compiling acton-streamer


ciaran2:
### Acton Version

0.17.0.20231012.10.19.40

### Reproduction code

Seems to be triggered by the lines in this diff: https://github.com/dt-terastream/acton-streamer/commit/7cd7c3a273e862015f0f367454eac60da17f605c

### Error output

```
actonc build
Building project in /home/ccain/acton-streamer
  Compiling tsdb/json-err.act for release
   Already up to date, in    0.000 s
  Compiling tsdb/http.act for release
   Already up to date, in    0.002 s
  Compiling tsdb/m3.act for release
   Already up to date, in    0.000 s
  Compiling tsdb/test.act for release
   Already up to date, in    0.000 s
  Compiling acton_streamer/common/utils.act for release
   Already up to date, in    0.003 s
  Compiling acton_streamer/common/mod.act for release
   Already up to date, in    0.005 s
  Compiling acton_streamer/common/io.act for release
   Already up to date, in    0.002 s
  Compiling acton_streamer/net/netconf.act for release
   Already up to date, in    0.000 s
  Compiling acton_streamer/nsoapi/proto.act for release
   Already up to date, in    0.002 s
  Compiling acton_streamer/nsoapi/conf.act for release
   Already up to date, in    0.001 s
  Compiling acton_streamer/nsoapi/channel.act for release
   Already up to date, in    0.000 s
  Compiling acton_streamer/nsoapi/cdb.act for release
   Already up to date, in    0.002 s
  Compiling acton_streamer/nsoapi/schema.act for release
   Already up to date, in    0.007 s
  Compiling acton_streamer/nsoapi/conv.act for release
   Already up to date, in    0.000 s
  Compiling acton_streamer/nso/subscriber.act for release
   Already up to date, in    0.001 s
  Compiling acton_streamer/main/config.act for release
   Already up to date, in    0.001 s
  Compiling acton_streamer/nsoapi/maapi.act for release
   Already up to date, in    0.002 s
  Compiling acton_streamer/nso/writer.act for release
   Already up to date, in    0.000 s
  Compiling acton_streamer/tsdb/writer.act for release
   Already up to date, in    0.000 s
  Compiling acton_streamer/main/main.act for release

ERROR: Error when compiling acton_streamer.main.main module: Type error
actonc: Acton/TypeM.hs:(184,39)-(186,87): Non-exhaustive patterns in case
```

[...]

Issue #1555, opened 2023-10-12
NoInfo 86 Type Error


ciaran2:
### Acton Version

 0.17.0.20231012.10.19.40

### Steps to Reproduce and Observed Behavior

This snippet in the acton-streamer repo https://github.com/dt-terastream/acton-streamer/blob/1aad5ac850ae5903306fa3bce5d6df39bed0b42a/src/acton_streamer/main/main.act#L722 triggers a "NoInfo 86" type error at compile time.

Removing the line `_get_xml.children.append(_curr_node)` allows compilation to proceed without issue.

It's unclear to me what the origin of the type error is in this case because it should be appending an `xml.Node` to a `list[xml.Node]` and more minimal reproduction of the case compiles without issue.

### Expected Behavior

It would be helpful for the compiler to more clearly indicate the nature of the type error in this case.

plajjan:
The error
```
kll@Boxy:~/terastream/acton-streamer$ actonc src/acton_streamer/main/main.act 
Building project in /home/kll/terastream/acton-streamer
  Compiling tsdb/http.act for release
   Finished compilation in   0.422 s
  Compiling tsdb/m3.act for release
   Finished compilation in   0.119 s
  Compiling acton_streamer/common/utils.act for release
   Finished compilation in   1.139 s
  Compiling acton_streamer/common/mod.act for release
   Finished compilation in   0.876 s
  Compiling acton_streamer/common/io.act for release
   Finished compilation in   0.047 s
  Compiling acton_streamer/net/netconf.act for release
   Finished compilation in   0.379 s
  Compiling acton_streamer/nsoapi/proto.act for release
   Finished compilation in   0.645 s
  Compiling acton_streamer/nsoapi/conf.act for release
   Finished compilation in   0.102 s
  Compiling acton_streamer/nsoapi/channel.act for release
   Finished compilation in   0.022 s
  Compiling acton_streamer/nsoapi/cdb.act for release
   Finished compilation in   0.625 s
  Compiling acton_streamer/nsoapi/cdb.act for release
   Already up to date, in    0.000 s
  Compiling acton_streamer/nsoapi/schema.act for release
   Finished compilation in   4.224 s
  Compiling acton_streamer/nsoapi/conv.act for release
   Finished compilation in   0.123 s
  Compiling acton_streamer/nso/subscriber.act for release
   Finished compilation in   0.738 s
  Compiling acton_streamer/main/config.act for release
   Finished compilation in   0.261 s
[...]

PR #1559, opened 2023-10-17
Improve errmsgs 2

sydow:


sydow:
This pull requests makes some minor improvements to error messages for type errors. In particular,
- it gives a good error message for the error in issue #1555. For the minimal example discussed in the comments to that issue, the error message is

```
ERROR: Error when compiling optex module: Type error
 6:9-18
  |
6 |        foo.append(4)
  |        ^^^^^^^^^^
?__builtin__.list[__builtin__.int] does not have an attribute append
Hint: you may need to test if foo is not None
```
- it gives a not very good error message when compiling a branch of acton-streamer discussed in issue #1558. The error message is
```
ERROR: Error when compiling acton_streamer.main.main module: Type error
 586:34-39
    |
586 |actor NetconfGetPollSubscription(source: NetconfSource, dev_sub_key: Keypath, shared_schema: schema.SharedSchema, log_handler: logging.Handler):
    |                                 ^^^^^^
(NetconfSource, acton_streamer.common.mod.Keypath, acton_streamer.nsoapi.schema.SharedSchema, logging.Handler) must be castable to (_, _, _)
```

The first of these examples gives what we want, including a hint. But one must be aware that this was a very low hanging fruit; I do not expect it to be easy to give hints in general.

The second example at least gives a possible  location of the error, and with some careful reading one can see that there seems to be a problem with the number of arguments. But it does not say that, ca 300 lines earlier, an actor instance was created with only three args in the call... I believe that this will be what we end up with in the first round in many, many cases. We have two contradictory constraints, one is first solved and then the other is noted as in error. It is not at all obvious to me how to reorganize the code to give an error message which quotes from lines 317 and 586 and points out that they are incompatible.

plajjan:
@sydow nice, that looks great =)

We got pretty good errors for a mismatch in number of positional arguments to functions so it's a little surprising it's worse for actors but perhaps this is something that could improve in the actor / class unification.

At least this error gives one source code location, so it's possible to guesstimate :)

sydow:
The problem with actors as compared with functions when it comes to positional arguments is as follows (for a smaller example than acton-streamer, look at test/typeerrors/ex13.act):

In the acton-streamer example (and ex13.act) the creation of an actor comes before (in the program text) the actor definition. Compile ex13.act and see that you get the same kind of error message as in acton-streamer. Then edit ex13 to move actor main after actor Act. Compile and see the nicer functional-style error message. This effect does not occur for functions (check ex14.act). I think it has to do with matchActorAssumption in Types.hs, but I gave up…

Björn

PR #1581, opened 2023-11-17
Improve http.Client connection handling

mzagozen:
If the (HTTPS) server closes the socket on us after a response, the stream becomes invalid. Let's communicate that to the user (http.Client) by ~~calling on_error~~. The user (http.Client) can automatically reconnect and resend the failed requests without any intervention from the actors further up the stack.

The HTTP(S) server closing the socket is perfectly valid behavior for HTTP/1.0. Users of http.Client do not have to do anything, they just get the response as expected.

### DIFF
--- a/base/src/__builtin__.act
+++ b/base/src/__builtin__.act
@@ -344,6 +344,9 @@ class MemoryError (Exception):
 class OSError (Exception):
     pass
 
+class ConnectionError (OSError):
+    pass
+
 class RuntimeError (Exception):
     pass
 
--- a/base/src/http.act
+++ b/base/src/http.act
@@ -513,6 +513,10 @@ actor Client(cap: net.TCPConnectCap, scheme: str, address: str, port: ?int, tls_
                 if "connection" in r.headers and r.headers["connection"] == "close":
                     # Is this really the right thing to do here? If the client
                     # does not make a new request we just reconnected for nothing!
+                    # TODO: inspect the _on_response queue to see if there are more requests
+                    # TODO: for HTTP/1.0 do not reconnect right after the first
+                    # request, wait until the 2nd query and then assume there
+                    # will be more.
                     _log.debug("Closing TCP connection due to header: Connection: close", None)
                     _conn_reconnect()
                 if len(_on_response) == 0:
--- a/base/src/http.act
+++ b/base/src/http.act
@@ -557,13 +557,10 @@ actor Client(cap: net.TCPConnectCap, scheme: str, address: str, port: ?int, tls_
                 await async tcp_conn.write(data)
             elif tls_conn is not None:
                 await async tls_conn.write(data)
-        except RuntimeError as exc:
+        except ConnectionError as exc:
             # HTTP/1.0 servers close the connection after each request by default
-            if "bad file descriptor" in str(exc) or "bad stream" in str(exc):
-                _log.debug("TCP connection closed, reconnecting", {"error": str(exc)})
-                _conn_reconnect()
-            else:
-                _on_con_error(str(exc))
+            _log.debug("TCP connection closed, reconnecting", {"error": str(exc)})
+            _conn_reconnect()
 
     # HTTP methods
     def get(path: str, headers: dict[str, str], on_response: action(Client, Response) -> None):
--- a/base/src/http.act
+++ b/base/src/http.act
@@ -447,6 +447,7 @@ actor Listener(cap: net.TCPListenCap, address: str, port: int, on_listen_error: 
 # TODO: default schema="https"
 # TODO: default port=None
 # TODO: default tls_verify=True
+# TODO: add arguments for configuring request buffering (size) and reconnection attempts and timeouts
 actor Client(cap: net.TCPConnectCap, scheme: str, address: str, port: ?int, tls_verify: bool, on_connect: action(Client) -> None, on_error: action(Client, str) -> None, log_handler: ?logging.Handler):
     """HTTP(S) Client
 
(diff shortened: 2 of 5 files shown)


base/src/net.ext.c
```diff
@@ -630,8 +630,14 @@ void tls_write_cb(uv_write_t *wreq, int status) {
 $R netQ_TLSConnectionD_closeG_local (netQ_TLSConnection self, $Cont c$cont, $action on_close) {
     uv_stream_t *stream = (uv_stream_t *)from$int(self->_stream);
     // fd == -1 means invalid FD and can happen after __resume__
-    if (stream == -1)
+    if (stream == -1) {
+        // The TLS connection may have been closed already when the server
+        // closed the socket. In that case just call the on_close callback.
+        char errmsg[] = "TLS TCP socket already closed?";
+        log_debug(errmsg);
+        on_close->$class->__asyn__(on_close, self);
```
mzagozen, line 638: We initially thought about raising an exception here instead, but `http.Client` calls `net.TLSConnection.write()` asynchronously. And I guess others would too?

plajjan: I don't think you should blindly accept async write as the correct thing. It's not a complete design as-is...

plajjan: Actually, for close, I'm not sure we even need to close on_close. Like why do we end up here with a -1 stream?

If you call `TLSConnection.close()` 10 times in a row, would we expect the `on_close()` callback to also be called 10 times? I think it makes more sense to treat it as a declarative thing, so if `TLSConnection.close()` is called on a closed connection, the goal of having the connection closed is already reached. We don't actually transition, i.e. we don't go from established -> closed, and so on_close will not be called!?

The stream must have gone to -1 at some earlier point at which the on_error() callback should be called, right?

mzagozen: Calling `on_close` even though we're already closed (`stream == -1`) allows for chaining the callbacks in `reconnect()`:

```python
    def reconnect():
        close(_connect)
```

base/src/http.act
```diff
@@ -500,10 +511,10 @@ actor Client(cap: net.TCPConnectCap, scheme: str, address: str, port: ?int, tls_
             r, buf = parse_response(buf, _log)
             if r is not None:
                 if "connection" in r.headers and r.headers["connection"] == "close":
-                    close_connection = True
-                    _conn_close()
+                    # Is this really the right thing to do here? If the client
```
plajjan, line 514: Add a TODO entry here. We should inspect our queue, i.e. _on_response list to see if there is more stuff.

I'm also thinking like we can maybe have some basic heuristics, like when we create a http.Client the first time we obviously connect directly then we expect to run at least one query. Maybe we should not reconnect after that (in case of HTTP / 1.0 or not having persistent in later HTTP versions) until there is a second query. Now if we see a second query we can assume "there are multiple requests" and thus reconnect directly after the 2nd and subsequent requests. WDYT?

base/src/http.act
```diff
@@ -524,22 +535,31 @@ actor Client(cap: net.TCPConnectCap, scheme: str, address: str, port: ?int, tls_
...
+            # signaling the socket was closed. Note that this is *not* waiting
+            # for network I/O, just waiting on system I/O for writing to the
+            # local socket buffer.
+            if tcp_conn is not None:
+                await async tcp_conn.write(data)
+            elif tls_conn is not None:
+                await async tls_conn.write(data)
+        except RuntimeError as exc:
+            # HTTP/1.0 servers close the connection after each request by default
+            if "bad file descriptor" in str(exc) or "bad stream" in str(exc):
```
plajjan, line 558: I think we should group this under some reasonable exceptions instead, so we can just do like `except BadStff` instead of doing string matching?

base/src/net.ext.c
```diff
@@ -276,6 +276,12 @@ void on_connect6(uv_connect_t *connect_req, int status) {
         char errmsg[1024] = "Failed to write to TCP socket: ";
         uv_strerror_r(r, errmsg + strlen(errmsg), sizeof(errmsg)-strlen(errmsg));
         log_warn(errmsg);
+        if (strstr(errmsg, "bad file descriptor")) {
```
plajjan, line 279: `r` contains a `UV_something` error, better inspect it than doing string comparison (le expensive). See uv.h

base/src/net.ext.c
```diff
@@ -276,6 +276,12 @@ void on_connect6(uv_connect_t *connect_req, int status) {
...
+            // This can happen if the socket is closed, se we raise an exception
+            // and let the caller retry
+            $RAISE(((B_BaseException)B_RuntimeErrorG_new(to$str(errmsg))));
```
plajjan, line 282: We are generally doing Python-like-exception stuff. It's not a specific design, i.e. we can deviate where we think there is good reason but if we have no idea we might as well follow Python. I think Python uses OSError for most of these, not RunTimeError.

base/src/net.ext.c
```diff
@@ -276,6 +276,12 @@ void on_connect6(uv_connect_t *connect_req, int status) {
         char errmsg[1024] = "Failed to write to TCP socket: ";
         uv_strerror_r(r, errmsg + strlen(errmsg), sizeof(errmsg)-strlen(errmsg));
         log_warn(errmsg);
```
plajjan, line 278: Not sure if we need log_warn here when we raise an exception...

base/src/__builtin__.act
```diff
@@ -344,6 +344,9 @@ class MemoryError (Exception):
 class OSError (Exception):
     pass
 
+class ConnectionError (OSError):
```
plajjan, line 347: I hadn't actually looked much at how Python does things beyond looking at OSError. I see now that it subclasses things quite nicely. Should we not just copy that whole thing verbatim? Like it feels like a better starting point than starting from scratch. In particular, in this case for the error you later raise, I think it's a BrokenPipeError.

We probably want to consider changing the `__init__` method of our OSError so we can take errno and stuff like that, again just like in Python. Without default args it'll be a little fiddly, but that's OK I guess.

PR #1589, opened 2023-11-21
Improve typeerrormsgs 3

sydow:
Some improvements to error messages for type errors. Still very incomplete, but this version should b easier to merge with Johan's new solver.


plajjan:
Very nice @sydow! :)

PR #1593, opened 2023-11-27
Add Windows support

plajjan:


plajjan:
Got some more failures but they're not very frequent. 

- https://github.com/actonlang/acton/actions/runs/7010933665/job/19084648706?pr=1593

We see a GC problem and it also showed up in the first one:

https://github.com/actonlang/acton/actions/runs/7006015835/job/19056958408

GC problem:

```
app ERR: Signals delivery fails constantly at GC #5
app ERR: Signals delivery fails constantly
app ERR: 
testOUT: App exited in state 5 with exit code 0 and termination signal 6
testERR: App exited in state 5 with an unexpected exit code 0 and termination signal 6
testOUT: stopping everything, reason: error encountered
testOUT: ps still alive.. 3 remaining
testOUT: ps still alive.. 3 remaining
```
code:
```c
...
          if (newly_sent != prev_sent) {
            retry = 0; /* restart the counter */
          } else if (++retry >= RESEND_SIGNALS_LIMIT) /* no progress */
            ABORT_ARG1("Signals delivery fails constantly",
                       " at GC #%lu", (unsigned long)GC_gc_no);
...
```
But why? I don't understand what would trigger this. While there are changes to the build.zig in the GC, it shouldn't affect this. The default behavior is still that we build with thread support and that's what it should do on Linux, so that part is the same!?

I haven't really been able to reproduce this locally either, making it very hard to reproduce.

@nordlander if you have time, feel free to stare at this patch and see if you can spot any obvious errors.

plajjan:
I've rebased and rewritten this a bit. We don't use asprintf anymore so that's not needed anymore. Since the first version of this PR, where we encountered those CI failures, I have fixed other segfault bugs, so I can't help but wonder if the errors we saw could be attributed to those...

plajjan:
I have had to rename logging.ERROR to logging.ERR to avoid a conflict with what appears to be a macro named `ERROR` on Windows. Not happy at all over that. We have the same problem with stdout/stderr where I renamed to something else to avoid conflict. Have issue #1744 to track this issue. I don't think sorting that out should block this work though, we'll just accept a slight regression now for the benefit of Windows support and sort it out nicely later.

plajjan:
CI has run 5 complete rounds. The only error was building Telemetrify which had a reference to logging.ERROR that I have now fixed in the acton-next branch so this last CI build that I've now retired should complete successfully. Then we'll merge this, yeeha!

plajjan:
We also need some form of CI testing for this, beyond just compiling an .exe.. but it's the same situation as our musl build, we test compilation but don't run the executable. We should fix that... but separately.

PR #1599, opened 2023-12-08
Relate "self" to Self

nordlander:
Added constraint that forces the type of first parameter of instance methods to be castable to Self.


plajjan:
Rebased on main after removing testing on macos 11.

PR #1612, opened 2024-01-04
Test additions

nordlander:


test/core_lang_auto/defaults.act
```diff
@@ -0,0 +1,54 @@
...
+#    print(out2)
+#    print(out3)
+#    print(out4)
+#    print(out5)
+#    print(out6)
+#    print(out7)
+#    print(out8)
+#    print(out9)
+
+    if out1 != "hello, world":
```
plajjan, line 25: There's `testing.assertEqual()` which is quite a bit nicer from a coding perspective but it would introduce a dependency on the testing module which in turn relies on exceptions etc. Should we care about such dependencies? Maybe it does make sense to have some hierarchy of languages features so that core features only relies on core features and not stdlib etc etc?

Issue #102, opened 2021-08-20
Overload / use defaults for range() function


plajjan:
The `range()` functions currently requires three arguments:
- start
- end
- increment

In Python, it is possible to invoke with one, two or three arguments:
```
>>> list(range(0,3,1))
[0, 1, 2]
>>> list(range(0,3))
[0, 1, 2]
>>> list(range(3))
[0, 1, 2]
>>> 
```

I think we should allow the same in Acton.

nordlander:
@plajjan Absolutely! It's just that the I've deliberately left out the part of the type-checker that would allow function arguments to be absent, for now. Will need some work before it can fly.

plajjan:
Hmm, would it be "leaving out arguments", isn't it about "supporting default values for arguments", which then become optional, no? Maybe it's both, just depends on perspective...

nordlander:
Yep, that's true, but the tricky part during type-inference is matching the actual and expected argument lists in such a way that still allows the actual list to be shorter. It's that part that's not fully implemented, the definition of functions with default argument values is actually already supported.

plajjan:
Now that we have default argument values we support `range(4, 5)` so that the step gets the default value of 1. We also support the single arg version of `range(8)` in which case the only argument the function sees is the "start" argument, which is sort of weird since the user provides the *stop* value. We check which arguments are provided and use the *start* value either for *start* in the 2-val case or as the *stop* value in 1-val case, so this works but I guess can be misleading...?

If someone tries to do `range("foo")` the error could be `Wrong argument type, provided 'str' but 'start' argument takes an 'int'`. The argument is actually used for `stop` but I don't think Python's nor our type system can accept an optional argument before a mandatory argument and sort of fill in the values backwards, right?

plajjan:
@sydow I note our numpy arange() function has a different type signature. Is this intentional or should it also be aligned on range()?

```python
class range (value):
    def __init__(self, start: int, stop: ?int, step: ?int) -> None:
        NotImplemented
```
```python
def arange(start: ?int, stop: int, step: ?int) -> ndarray[int]:
    NotImplemented
```

Looks like arange() is trying to define first start argument as optional, but that can't ever work, can it? also see above comment..

plajjan:
I have updated all our uses of `range()` in the code base to use the 1 or 2 val cases where possible, see #1613.

nordlander:
[...]

PR #1613, opened 2024-01-06
Use defaults args for range()

plajjan:
Since we haven't previously supported default arguments values, all calls to range() have been of the explicit 3 arg kind. Now that we do support default arguments we opt to use the more concise 1 and 2 arg call of range where we rely on the default values, i.e.:
```
  range(0, foo, 1)
```
becomes:
```
  range(foo)
```
and:
```
  range(1, bar, 1)
```
becomes:
```
  range(1, bar)
```


plajjan:
Part of #102.

PR #1637, opened 2024-01-12
Errormsgs for better rows

sydow:
This adapts type errors to the new rows.  Some new type error examples are added.

Almost all golden files have changed, some not much but they had to be updated. I don't think I treat these golden tests in the best way?


plajjan:
I don't see a problem with updating the golden files. The point is to detect when there are changes to this output. It is how I noticed the regression from Johans work. It makes type errors visible through github.

Issue #1658, opened 2024-01-23
lstrip & rstrip with default arg broken


plajjan:
```python
    def lstrip       (self, chars: str) -> str:
        NotImplemented
    def rstrip       (self, chars: ?str) -> str:
        NotImplemented
```

for some reason for lstrip the chars arguments is required. I don't think that was intentional as it seems the function is prepared for accepting a default / NULL argument. However, that does not work!

When I use rstrip or lstrip (after modifying __builtin__.act), the output is just completely blank!

```
actor main(env):
    a = " a b "
    print(a)
    b = a.lstrip()
    print(len(b))
    env.exit(0)
```

output:
```
 a b 
0
```

plajjan:
@sydow I think you know best how to fix this! :)

plajjan:
While we are on the topic, what should the default character set to strip away be? I suppose this was easy in an ASCII world but much trickier with unicode. I have a vague recollection that you've raised this question in the past @sydow. What does Python do?

sydow:
The default in Python is to remove whitespace, which is prepared for in our str.c:

static struct B_str whitespace_struct = {&B_strG_methods,6,6,(unsigned char *)" \t\n\r\x0b\x0c"};

static B_str whitespace_str = &whitespace_struct;

There are some more strange Unicode code points that are classified as whitespace, but I think we ignore them for now.

lstrip seems to be seriously wrong, but should be easy to fix…

PR #1663, opened 2024-01-25
fixes bugs in lstrip, rstrip and setdefault

sydow:


sydow:
In addition to fixing #1658, I did three other things
- fixed a bug in setdefault
- tried to add methods copy and clear to dict. However, a bug in code generation for calls in client code (no Hashable witness supplied) I commented out this. 
- I happened to look at bytearray_test.act in test/builtin_auto. Most of it is commented out with a first line "# TODO: fix all this". I guess this refers to rewriting it to try something, test whether it works and otherwise exit with code 1. In bytearray_test.aact there are only several operations tried with result printed, just as one does during development. Couldn't we keep this and use the concept of golden files rather than rewriting the file?

plajjan:
@sydow nice and I really like to see our test coverage expanded!

Will you open an issue for the problem with code generation when adding copy & clear? I think it's better if you describe it.

As for bytearray_test, I believe I wrote that TODO and I think it refers to both that some things were not actually working (otherwise I would not have commented out all those lines) and that the test should be rewritten to not rely on stuff printed to stdout. I presume we have fixed most of the bugs related to bytarray, so we can uncomment those lines again. If we are to enable golden tests or use conditions in the test itself... hmm, yeah. I guess we could use golden tests more broadly. Need to consider if there is terminal mumbo jumo messing with encoding... and if golden tests should be enabled for everything? It only works for tests having 100% deterministic output obviously, so test using multiple actors are not allowed to print stuff since it would be racy. If you have some suggestion there I'm all ears. The test structure is built a little ad-hoc. Categorized based more on subsystem in Acton, like "builtins" so we can choose to just run the tests of the category that we happen to be working on. It's not perfect by any means. Maybe there is a different category organiation that's better.

PR #1665, opened 2024-01-25
More support for ty files

sydow:
This pull request adds two things:
- actonc  can take one file argument with suffix .ty; if the file exists, its contents is prettyprinted on stdout.
- one more Boolean CompileOption --ty is added; this causes copying of the .ty file from the temporary  project dir to the source file directory before the temporary dir is erased.


plajjan:
Cool, I retried the CI test another time so it passed now. Lots of boring intermittent timeouts unfortunately... 

Changes look good, I'll merge.

PR #1666, opened 2024-01-26
Test discovery

sydow:
This is a first attempt at implementing Kristian's ideas. All the examples in directory test/stdlib_test/src work (and files have been modified to remove what is automatically generated).

The examples are mostly unit tests, so we need more examples of the other categories to be convinced. 

Everything is done in Types.hs; still not sure whether that is a wise move. I will not attempt to merge this now; maybe Kristian can play with it and try more examples.


sydow:
I didn't think of the fact that to pass test/stdlib_auto/test_testing.act it must be compiled with flag --root __test_main. This is as it should be but is (probably) not reflected in the test. Leaving this to Kristian.

PR #1674, opened 2024-02-01
Place data from initialization in non-GC heap

### LINKED ISSUE #1493: Place module constants outside of GC heap
Module initialization code is run during startup. Since module globals are all constants we could except these from GC scanning by placing them outside of the heap.

How?

- Can we somehow dynamically redirect malloc between malloc and GC_malloc_uncollectable?
- Can we except parts of the heap after we have run module initialization?

plajjan:
We now make use of the non-GC malloc, i.e. the normal malloc from libc on startup of Acton up to and including the module initialization. Since Acton only allows for constants at the module top level, we know all such values are immutable and static, thus we do not have to place them in GCed memory. To avoid the GC, since we are doing link time redirection, we dig out the real malloc using dlsym. This will only work on platforms where libc is dynamically loaded. It should be possible to rework this so that we can work in statically linked environments too.

This change makes it a viable path forward to place large but relatively static data as module constants since it can drastically reduce the GC heap and thus improve the performance of the program.

Fixes #1493

### DIFF
--- a/base/builtin/bool.c
+++ b/base/builtin/bool.c
@@ -52,7 +52,7 @@ B_bool B_boolG_new(B_value s) {
 }
 
 B_bool toB_bool(long b) {
-    B_bool res = malloc(sizeof(struct B_bool));
+    B_bool res = acton_malloc(sizeof(struct B_bool));
     res->$class = &B_boolG_methods;
     res->val = b;
     return res;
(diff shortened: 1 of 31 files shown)


plajjan:
This breaks on the DB resume tests. @nordlander I need to sit down and talk with you to figure out this out I think. I think this malloc / GC-malloc scheme, as per this PR, should work for deserialization as well but.. well, it doesn't, so must be missing something. Need to reason about it :)

PR #1699, opened 2024-02-22
Unboxing 1

sydow:
This pr does unboxing of variables of bounded numeric type and restores arithmetic and Boolean expressions to their natural form. Most of the work is done in a new boxing pass between lambda lifting and code generation. Some clean-up will be needed, but since I advertised it I thought it prudent to make it available.

We also remove all unnecessary witnesses from the final C code. In particular, my dct program has no witnesses at all in the C code. In telemetrify, around 800 unnecessary witnesses were removed. But the effect on time in telemetrify is marginal; ca 3-4%. But in a program where 75% of the time is spent on GC, this is perhaps not so marginal after all...

Two tests in the test suite were changed in order to pass, but I think the test programs were to blame.


plajjan:
@sydow very cool!

PR #1722, opened 2024-03-06
Fix stdin_install type, str -> bytes

plajjan:


plajjan:
Closing in preference of #1723.

PR #1782, opened 2024-04-10
Test mem

plajjan:


plajjan:
@sydow I fixed $FORMAT by switching asprintf for snprintf and having it first figure out how long of a string is necessary, then doing acton_malloc and finally doing the actual string formatting. It's faster than before so I guess asprintf does something similar internally. I've also added a to_str_noc for no-copy and use it where necessary. There might be further optimization potential here.

But more importantly, with the help of $FORMAT, I have now replaced all asprintf calls. I think this looks really good, like we turn 3 lines into a single $FORMAT call which is just much easier to read and work with. In our last meeting I asked you to ponder the design for string formatting but I feel so pleased about this design (and it fixes our current issue with memory leaks) that I am going to go ahead and merge this. I still want your feedback on it (we can always revert or make further changes) and there might be optimization potential among the to$str, toB_str and to_str_noc functions which you are better positioned to do.

PR #1794, opened 2024-04-23
Better scoping rules

nordlander:
+ correct types of defaultable parameters
+ type-correct assignment of default values

### DIFF
--- a/compiler/Acton/Deactorizer.hs
+++ b/compiler/Acton/Deactorizer.hs
@@ -148,7 +148,7 @@ instance Deact Decl where
 
             props (VarAssign _ p _) = [ (n, t) | PVar _ n (Just t) <- p ]
             props (Assign _ p _)    = [ (n, t) | PVar _ n (Just t) <- p, n `elem` locals ]
-            props (If _ bs els)     = restrict (concat $ map props els) (foldr1 intersect $ map bound bs)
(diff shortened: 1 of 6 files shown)


plajjan:
Weehoo, so this unlocks the unboxing-3 branch, right? @nordlander

nordlander:
Yep, it should, although Björn must verify whether that's actually the case. I'm also noticing some failures compiling telemetrify with this branch, some of which seem to require just trivial updates to telemetrify but there's also one that smells like a lurking compiler bug. I'll rest when I've fixed that!

plajjan:
@nordlander oh, I see. You managed to merge despite `test-external-project` test failing (that test compiles Telemetrify). We should enforrce... I think we did but must have disabled at some point I guess.

nordlander:
@plajjan Yes, but I didn't notice that the merge was successful until your fanfare! To me it looked like no merge because of the faiure notification I got.

Anyway, I've fixed the bug I mentioned now. But for test-external-projects to succeed we need to slightly polish telemetrify as well, by updating the argument order for the Process constructor (net/netconf.act) and replacing ERROR by ERR in main/config.act. Not sure in which branch to do that, though.

plajjan:
@nordlander right, that part is my fault. It's because I made that release of Acton 0.22 and I haven't yet telemetrify, by merging in acton-next into main. I will take careof that.

PR #1812, opened 2024-05-30
Fix str slice outside of range

plajjan:
The problem is really that slice is 0 long and allocating an array that is 0 long and then doing operations on it becomes UB so we SIGILL in testing since we now run tests in --dev mode (which comes with all sorts of UBsan).


plajjan:
@sydow can you please look at this. Do you think this is the correct fix? I am going to put this on auto-merge since this fixes my current problem, but I'm raising this with you in case you think there's a better or more appropriate fix.

Issue #1871, opened 2024-08-26
Only include deps as sources, do not bundle built libs in distribution


plajjan:
- [x] argp
- [x] bsdnt
- [x] bdwgc
- [x] mbedtls
- [x] netstring
- [x] pcre2
- [x] protobuf-c
- [x] tlsuv
- [x] utf8proc
- [x] uuid
- [x] uv
- [x] xml2
- [x] yyjson
- [x] snappy-c

PR #1872, opened 2024-08-26
Stop building utf8proc for distribution

plajjan:
Just ship the source code and instead let zigbuilder build it when needed.

Also moves the dependency from the main build.zig to base's build.zig, which we can do now that we are relying entirely on the Zig style build and no manual linking of .a files, like we did before.

Bump version of utf8proc to one with a newer build.zig that installs the header correctly and also has option for shared build (which we don't use).


plajjan:
Part of #1871.

PR #1877, opened 2024-08-27
Allow concurrency for cross-compile tests

plajjan:
This is rather crude but as we have a project lock, we cannot get concurrency when doing the cross-compilation tests to various platforms when we are using the same project. The simple workaround is to compile different projects!


plajjan:
We had other concurrency issues at the time. Seeing how cross-compilation tests are primarily CPU constrained, there is no need to add more parallelism, so I'm scrapping this.

PR #1888, opened 2024-08-28
Clean .build/sys symlink when done

plajjan:


plajjan:
I've removed the whole .build dir altogether so this is no longer relevant.

PR #1896, opened 2024-08-30
Add remote_close callback to TCPConnection / TCPListenConnection / TLSConnection

ddjoh:
Callback to indicate when the remote side has shutdown / initiated close of its side of the TCP/TLS connection


plajjan:
LGTM. Not quite sure about overall API but we still have time to change that before 1.0, so merging this now :)

PR #1923, opened 2024-09-18
Added forgotten scope extension in the normalizer.

### LINKED ISSUE #1913: Try-scope local not generally available in else branch
### Acton Version

0.23.0.20240910.15.8.1

```
import xml

actor main(env):
    try:
        root = xml.decode("<test/>")
    except Exception as ex:
        pass
    else:
        print(root.tag)
    env.exit(0)
```

### Expected Behavior

Success!

nordlander:
And removed one defunct file. Fixes #1913.


plajjan:
I'd like to have a test that covers this. @nordlander can you write one that properly covers this or can we just take the reproduction case from #1913 ? (I can add that in)

nordlander:
Just reuse the repro case!

PR #1937, opened 2024-10-08
Re-enable any() & all()

### LINKED ISSUE #1936: Re-enable / reimplement `any` & `all`
`any` and `all` were commented out some time ago. @sydow do you remember why? The related commit is https://github.com/actonlang/acton/commit/0e53227c3183ea4763bf1ab4c8a9e93c45129456 

Just uncommenting the code again leads to a compilation error for $NOT... but I think that's something more general about boolean-something-blabla-problems in the compiler. I think it's possible to rewrite the all / any functions to work right now without fixing anything else.

plajjan:
These were commented out, presumably due to a compilation error. I do get an error when I commented them out but I have reshaped the functions somewhat, forcing the bool function and checking it's result, which seems to be working.

Also added tests.

Fixes #1936

### DIFF
--- a/base/src/__builtin__.act
+++ b/base/src/__builtin__.act
@@ -840,17 +840,17 @@ extension bytes (Hashable):
 def abs(x : Number) -> Real:
     return x.__abs__()
 
-#def all [A(value)] (it : Iterable[A]) -> bool:
-#    for x in it:
-#        if not x:
-#            return False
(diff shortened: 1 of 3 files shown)


plajjan:
@sydow @nordlander besides this being a little uglier, it compiles and works, so I'm merging...

PR #1964, opened 2024-10-30
Fix conversion of u16 to int

ciaran2:


plajjan:
Closing, superseded by #2047. The fix here is correct but I wanted tests to show the problem before the fix and that the fix indeed fixes the issue. I added test cases but couldn't push to the ciaran2 branch, so pushed to new branch, which is the PR in 2047. Closing this.

PR #1986, opened 2024-11-13
added method pop to protocol Mapping and dict implementation

sydow:
This PR adds method pop to protocol Mapping and the dict implementation. Also a minor change to method get. The signatures are 
    get         : (A,?B) -> ?B
    pop         : mut(A,?B) -> ?B


sydow:
The required check test-app-yang fails and prevents the merging of this PR which would fix issue #1989. I don't know what the problem is.

plajjan:
Agreed in todays meeting that this is the way forward, having separate get() and getdef() functions. I will split this PR into two parts, the first introducing getdef() / popdef(), then we can merge that and change the various applications we have after which we can do the second change, which is modifying the get() / pop() definitions in this backwards incompatible way. I will fix!

plajjan:
This is now implemented via #2040 etc. Closing.

PR #2009, opened 2024-11-29
Iterator exceptions

sydow:
Changes iterators so that they throw exception StopIteration at termination (rather than returning None). 

As a consequence, the Iterator class in __builtin__.act his now defined by

class Iterator[A] (object):
    __next__      : () -> A

(previously the return type was ?A)


plajjan:
This is failing CI testing because of the changes to `max()` I think. It's the same problem as for pop etc in #1986. I said I would work on that to split the change in two parts. I think I can do the same for this PR. As things currently look, I will have some spare cycles early next week.

plajjan:
This PR has some bug, not sure where. After I have fixed max / max_def, the functionality of Acton programs should be identical before and after this branch, yet, when I do `make gen` in the respnet repo, it generates entirely different code, indicating that somewhere we have an actual difference in behavior. I suppose it's best if I can find a smaller reproduction.. easier said than done.

sydow:
I don’t know what to do about this. If I do git pull in the main branch and then build the system and do make test, all tests pass. I also don’t know what the respnet repo is. Could you explain more?

Björn

plajjan:
Closing this, see #2023 instead!

PR #2011, opened 2024-12-02
Improve reading build config errors handling

plajjan:
We unify a bit across the various commands that we have.


plajjan:
This is merged in via #2024.

PR #2023, opened 2024-12-16
Replace C NULL for iteration stop with StopIteration

plajjan:


plajjan:
@sydow I have found a bug which is reproducible just in the Acton repository. I'm not sure if this is because of the same issue as in the respnet repo or not, but it's at least a difference in behavior that seems to come with this PR!

commands to reproduce:
```
cd test/compiler/hello
acton build
```

output:
```
kll@Boxy:~/dt/acton/test/compiler/hello$ acton build
INFO: Acton global cache is empty: rebuilding common libraries (e.g. libc) which will take some time...
INFO: Acton local cache is empty: rebuilding Acton base which will take some time...
Building project in /home/kll/dt/acton/test/compiler/hello
  Compiling hello.act for release
   Finished compilation in   0.010 s
  Final compilation step
   Finished final compilation step in  21.741 s
kll@Boxy:~/dt/acton/test/compiler/hello$
```

The messages `INFO: Acton global cache is empty: rebuilding common libraries (e.g. libc) which will take some time...` etc are always printed out. That's not supposed to happen. It should only happen when the cache dir is actually empty. This logic is in cli/src/acton.act and / or base/src/file.act I suppose...

nordlander:
I've located the bug, it's the CPS pass that contains a leak. Fix is coming up.

sydow:
I don’t want to compete about being the source of the bug, but I also found a bug that may be involved. This bug is easy to fix and I did so locally, but it does not repair the hello example, so maybe both bugs are involved? I have not yet pushed this, waiting for comments about further things I may need to do (see below).

Here is an example where my bug is involved:

``` 
def f(res):
   try:
       res = 1
       raise ValueError()
   except ValueError:
       pass
   return res         # should return 1

actor main(env):
   print(f(0))
   env.exit(0)
``` 

The interesting variable res is now a parameter to f. The above code erroneously prints 0, since the parameter is not declared volatile by the naive algorithm I used in my fix the other day. This algorithm only looks at the statement sequence in which the try/except statements occur, scanning it backwards from end to start looking for variable declarations where we need a volatile annotation. We need to continue to do this to the parameters of the fun (or actor) declaration of which the statement sequence is the body. I need to think about other cases (state vars in classes?). 

The reason I think that this bug may affect the hello example is the following fragment from cli/acton.act:

``` 
def warn_on_large_zig_global_cache(cap: file.FileCap):
    < beginning of code omitted for space reasons>
    total_size: int = 0
    for f in fs.walk(cache_dir):
        print ("f=",f,"size=",f.size)
        total_size += int(f.size)
    gb = 1024 * 1024 * 1024
    if total_size > warn_level * gb:
        <code for this case also omitted>
    if total_size == 0:
        print("INFO: Acton global cache is empty: rebuilding common libraries (e.g. libc) which will take some time...")
``` 

Here it is obvious that total_size must be declared volatile, which it would be using my naive algorithm, but CPS conversion enters and makes total_size be involved as a parameter to (several) C procedures and as a field  in C structures, so I get lost in the code and cannot really see what the fix may be in the C code,  even less how to achieve this in CodeGen.hs. But then maybe something went wrong in CPS conversion...

nordlander:
And fixing the overly eager CPS converter immediately exposes the problem Björn described. Intricate!

nordlander:
I've pushed my small fix that avoid unneeded CPS transation. It doesn't resolve the issue by itself, but will most likely do so in combination with Björn's pending commit. But the fundamental problem with CPS/lambda-lifting of exception handlers remains unsolved for now :-(

plajjan:
@sydow @nordlander FYI, it is still broken. CI testing shows almost all test are passing, but it's actually not. If we look in test-macos or test-linux, we can see that `acton test` doesn't actually run the tests properly. I think they immediately return an OK result instead of reporting actual progress - again, indicating improper control flow.

To reproduce, run `cd test/stdlib_tests && acton test` - compare on main with results on this branch - it just prints yellow stars for each test and exits instead of actually running the test.

plajjan:
I'd really like to have some test cases for the various failing cases that we are seeing now, so we don't just rely on whatever the acton cli happens to look like right now

plajjan:
Okay, so I added a new test program that includes @sydow example from above as well as what I believe is the failing patter we saw in the `acton` program, that @nordlander and I looked at. This new `cps1` test is still failing, which is good. It shows we have a problem :)

Feel free to rename the test program or whatnot, I don't know what this is called.

plajjan:
Whops, I just noticed I wasn't using the latest version of this branch. The `cps1` test program I just wrote does work with the latest fixes :stuck_out_tongue_closed_eyes: 

But okay, so there's got to be some other interesting interaction of features going on...

plajjan:
I have rebased this work on the latest main, pushed to a new branch and opened PR #2054 for   the new branch, which supersedes this one and thus I'm closing this PR.

PR #2054, opened 2025-01-02
Replace C NULL for iter stop with StopIteration

plajjan:


plajjan:
@nordlander I added a new test case which demonstrates some failure in the compiler to mark variables volatile. The new test case is in test/core_lang_auto/exceptions_volatile.act

First we can look at the working case, although note the two lines I commented out:
```python
def fun2():
    a = "FOOBAR2" # easy to find in C source
    my_list = [1]
    complete = True
    #if 1 == 1:
    #    complete = False
    for r in my_list:
        complete = False
    return complete
```

```c
kll@Kristians-MacBook-Air core_lang_auto % actonc --cgen exceptions_volatile.act | grep -A20 FOOBA
    B_str a = to$str("FOOBAR2");
    B_list my_list = B_mk_list(1, to$int(1));
    volatile B_bool complete = B_True;
    B_Iterator N_1iter = ((B_Iterator (*) (B_Iterable, B_list))W_27->$class->__iter__)(W_27, my_list);
    if ($PUSH()) {
        while ((B_True)->val) {
            B_int N_2val = ((B_int (*) (B_Iterator))N_1iter->$class->__next__)(N_1iter);
            B_int r = N_2val;
            complete = B_False;
        }
        $DROP();
    }
    else {
        B_BaseException N_3x = $POP();
        if ($ISINSTANCE0(N_3x, B_StopIteration)) {
        }
        else {
            $RAISE(N_3x);
        }
    }
    return complete;
```

most importantly we see `volatile B_bool complete = B_True;`. Now uncomment the lines, so we have:
```python
def fun2():
    a = "FOOBAR2" # easy to find in C source
    my_list = [1]
    complete = True
    if 1 == 1:
        complete = False
    for r in my_list:
        complete = False
    return complete

```

and notice how `volatile` disappears. That doesn't look right...
```c
kll@Kristians-MacBook-Air core_lang_auto % actonc --cgen exceptions_volatile.act | grep -A20 FOOBA
    B_str a = to$str("FOOBAR2");
    B_list my_list = B_mk_list(1, to$int(1));
    B_bool complete = B_True;
    if (ORD_B_int__eq__(to$int(1), to$int(1))) {
        complete = B_False;
    }
    B_Iterator N_1iter = ((B_Iterator (*) (B_Iterable, B_list))W_37->$class->__iter__)(W_37, my_list);
    if ($PUSH()) {
        while ((B_True)->val) {
            B_int N_2val = ((B_int (*) (B_Iterator))N_1iter->$class->__next__)(N_1iter);
            B_int r = N_2val;
```

nordlander:
I've spotted the error. It's in Björn's recent additions to CodeGen, duplicating an error I also made in my recent  mod to CPS (but was lucky enough to find and iron out!). Essentially, its root is the dubious nature of our utility function `bound`, which cannot tell a binding occurrence from a reassignment. There are alternatives in the compiler, but the honest truth is that the overloading of `bound` on Statements is a mess. :-(

I'll commit a fix to CodeGen tomorrow. But ideally we should also straighten up all code that deals with scopes in the compiler, for the benefit of our future selves! :-/

plajjan:
@nordlander the telemetrify test is failing. Compilation and the (very small) test suite passes but then the final somewhat larger test of loading in a schema from a replay file fails:
```
/__w/acton/acton/telemetrify/telemetrify-core/out/bin/telemetrify.test maapi-schema
Current time: 2025-01-03T16:06:37.627548212Z
Starting up...
SMAAPI connected!!!! 
*d�lmadminhaaaadsystemajMAAPI user session started!!!! 
��MAAPI loaded schema!!!! 
<$Actor -160 telemetrifyQ_nsoapiQ_schemaQ_SharedSchema at 0x7fef659a0f50>
MAAPI end user session failed: ValueError: Invalid `EObject` tag: 46
make: *** [Makefile:15: test] Error 1
```

I can only presume that control flow is messed up somehow. This is normally a very stable & deterministic test - I . Nonetheless, I have just cleared all CI caches and started a rerun just on the off chance there was an intermittent failure here.

plajjan:
Same failure after clean rebuild... @nordlander

plajjan:
:fireworks:

PR #2064, opened 2025-01-09
Add complex.from_real_imag()

plajjan:
This staticmethod constructor allows us to create a complex instance
from two Reals.

Add a test suite for complex.

Fix division to raise exception on zero-divison

The acton.guide has been updated with a page on complex numbers.

Add new RAISE macro to simplify raising exceptions from C code.

### DIFF
--- a/base/builtin/common.h
+++ b/base/builtin/common.h
@@ -58,4 +58,6 @@ void $default__init__($WORD);
     to_str_noc($b);                                      \
 })
 
+#define RAISE($T, ...)      $RAISE((B_BaseException)$NEW($T, ##__VA_ARGS__))
+
 char *unmangle_name(char *input);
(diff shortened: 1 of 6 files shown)


plajjan:
@sydow @nordlander FYI! I got complex working by adding this from_real_imag() constructor. Added a test suite and found the division-by-zero not raising exception, so I fixed that and also realized our $RAISE is very ugly so added a nicer macro around that. One test is disabled as it depends on __fromatom__, which is not implemented and I'm not sure how to proceed there. With a little guidance, maybe I could give that a shot.

Also added a bit of documentation. I have used the Claude AI both for producing the test suite and docs and while I strive to proof read and guide the LLM, I am not proficient enough in the subject to feel entirely confident about the result.  I will merge this now, because @mzagozen needs it for his advent-of-code and given the test suite I am quite confident this works well enough, but please feel free to read so we can fix any potential issues :)

sydow:
Great work, Kristian!

I think we will soon change everything with atom, so I suggest you fix the problem by changing the problematic test from

d = c ** 2

to 

d = c ** complex.from_real_imag(2,0)

plajjan:
@sydow ok, I've re-enabled the test according to your suggestion in #2070! Thanks!

PR #2067, opened 2025-01-09
Add support for complex numbers as dictionary keys

mzagozen:
This fixes segfaults when more than 1 complex key is added to a dict.


plajjan:
LGTM.

PR #2069, opened 2025-01-11
Fix hash function, no signed overflow UB

plajjan:
Bit shifting overflow is UB for signed integers. We now use unsigned (which are well defined for bit shift overflow) to avoid UB. Since we compile rather strictly (Zig's default to clang), any UB results in SIGILL.

### DIFF
--- a/base/builtin/hash.c
+++ b/base/builtin/hash.c
@@ -84,8 +84,8 @@ long B_i16D_hash (B_i16 n) {
 static long double_hash(double d) {
     int e, sign;
     double m;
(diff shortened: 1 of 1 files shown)


plajjan:
I am fairly confident this is correct and it fixes a crash that we ran into https://github.com/mzagozen/aoc2024/pull/2#pullrequestreview-2543743235 so I am going to merge. Nonetheless, I think it is good if you @sydow or @nordlander read the changes just to make sure they are sane, thanks! :)

PR #2073, opened 2025-01-12
Restructure compiler & add example LSP server

plajjan:


plajjan:
@sydow @nordlander Now that this is merged, I would suggest that you immediately rebase any existing branches you might have that modifies the compiler. I think the rebase should mostly be clean since the files have just moved, but it's probably a good idea to get the rebase done sooner rather than later. Let me know if you run into any issues.

PR #2105, opened 2025-01-21
Add CI job to run on Ubuntu 24.04 on arm64

plajjan:


plajjan:
I saw a few failures here but I don't think they are directly related to arm64, there is some problem with our test runner or something, see #2109. I'll merge this. We don't require this test to pass in order to merge.

PR #2107, opened 2025-01-21
Allow dep path OR url in build config

### LINKED ISSUE #2106: Prevent "path" overriding "url" in build.act.json
### Proposal

In `build.act.json`, where we list the project dependencies, a local intra-project dependency may use `"path"` to get the dependency from a local path. This is mutually exclusive with the `"url"` for specifying a remote (fetched) inter-project dependency.

Valid inter-project package dependency in a nested project:
```json
    "dependencies": {
        "yang": {
            "path": "../"
        }
    },
```

Valid intra-project dependency:
```json
    "dependencies": {
        "yang": {
            "url": "https://github.com/orchestron-orchestrator/acton-yang/archive/refs/heads/main.zip",
            "hash": "1220cd97e7833d7ea81d38a55240b8fc836dda21e6e9ed496d5713198a51ad68e6a0",
        },
```

1. Enforce the XOR in the compiler and guide the user to use `acton build --dep foo=../foo` if they intended to override an inter-project dependency.
2. Update the documentation (https://www.acton-lang.org/learn/acton-by-example/pkg/local_dependencies.html) to reflect the new recommendations for intra-project dependencies and stop suggesting to use `"path"` override for local inter-project dependencies.
3. Add a new documentation section to explain how to build a project with inter-project dependencies but override to local path (`acton build --dep foo=../foo`).

### Motivation

This guides the users to the only supported method of overriding dependencies.

### Alternatives

N/A

plajjan:
We used to allow path and url, in which case the path would override the url and this was useful to override with like a local fork of a dependency when doing development across multiple projects simultaneously. However, it's bad changing checked in files like build.act.json and so we added acton build --dep to override deps. It also has the benefit of being propagated to the build of dependencies, so if we have a main project that depends on foo and bar. In turn, foo depends on bar, so if we now override the path to bar, the same override is passed to the build of foo as to the main project.

Given that we have --dep, the old way of manually overriding the path in the configuration file is deprecated. We now block and require that only path OR url is set. Otherwise the user gets an error. We hint about the --dep argument and this is also based on the config the user has in their configuration file, e.g.:

  kll@Kristians-MacBook-Air test_deps % cat build.act.json
  {
      "dependencies": {
          "a_b": {
              "path": "deps/a-b",
              "url": "https://github.com/asdf/fnafna/refs/heads/main.zip"
          }
      }
  }
  kll@Kristians-MacBook-Air test_deps % acton build
  ERROR: Dependency 'a_b' has both path and url set. Set only path or url. You can override path for a dep, e.g.: acton build --dep a_b=deps/a-b
  kll@Kristians-MacBook-Air test_deps %

Fixes #2106

### DIFF
--- a/docs/acton-by-example/src/pkg/local_dependencies.md
+++ b/docs/acton-by-example/src/pkg/local_dependencies.md
@@ -16,7 +16,7 @@ It is possible to use dependencies available via a local file system path by set
 }
 ```
 
-These are best used for dependencies located within the same git repository or similar. All users need to have the same relative path to the dependency so if the paths stretch over multiple repositories, the used needs to keep the paths aligned.
+These are best used for dependencies located within the same git repository or similar. All users need to have the same relative path to the dependency so if the paths stretch over multiple repositories, the user needs to keep the paths aligned.
 
 ```admonish
 You can temporarily override the path to a dependency through the `--dep` argument, e.g. `acton build --dep foo=../foo`. This can be useful to fork a library and make local modifications to it before submitting them back upstream.
--- a/docs/acton-by-example/src/pkg/local_dependencies.md
+++ b/docs/acton-by-example/src/pkg/local_dependencies.md
@@ -1,13 +1,11 @@
 # Local Dependencies
 
-It is possible to use dependencies available via a local file system path by setting the `path` attribute. Edit `build.act.json` and add or modify an existing dependency. Set the `path` attribute to a relative path on the local file system, e.g.:
+It is possible to use dependencies available via a local file system path by setting the `path` attribute. Edit `build.act.json` and add or modify an existing dependency. Set the `path` attribute to a relative path, e.g.:
 
 ```json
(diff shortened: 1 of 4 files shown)


docs/acton-by-example/src/pkg/local_dependencies.md
```diff
@@ -18,6 +16,8 @@ It is possible to use dependencies available via a local file system path by set
 }
 ```
 
+These are best used for dependencies located within the same git repository or similar. All users need to have the same relative path to the dependency so if the paths stretch over multiple repositories, the used needs to keep the paths aligned.
```
mzagozen, line 19:
```suggestion
These are best used for dependencies located within the same git repository or similar. All users need to have the same relative path to the dependency so if the paths stretch over multiple repositories, the user needs to keep the paths aligned.
```

mzagozen (approved):
Nice!

PR #2133, opened 2025-01-28
Improve type error hint for is not None

plajjan:
'is' checks identity. None is C NULL so equal comparison is conceptually more correct in Acton since None shouldn't be thought of as a singleton as in Python.


plajjan:
@nordlander I just want to double check that you agree here. Pretty sure you've told me in the past that this "is not None" is idiomatic Python given singleton but since our None is not a singleton, the more natural way to think of this on a conceptual level is just a plain equal or not equal... right? and if that's the case, we should a) encourage users through type error hints to write idiomatic Acton and b) change a lot of our own invocations in stdlib etc to use `!= None` instead of `is not None`

plajjan:
The last papal bull was wrong or at the very least misinterpreted. `if x is not None` is the correct, idiomatic and preferred way. `if x != None` is to be avoided (oh my, I have some code to change back). Closing this.

PR #2148, opened 2025-02-01
Skip XML comments when building Node

mzagozen:


plajjan:
What is the current behavior? Isn't it nice we can include comments? So make the behavior optional?

mzagozen:
Right now a comment node raises an `Unexpected nodetype 8, content is ...` exception from xml.ext.c, making it impossible to read in an XML document with comments. Of course it would be nicer to preserve, but this fixes the immediate problem of not allowing comments at all.

PR #2149, opened 2025-02-03
Fix transitive deps

mzagozen:
When adding transitive dependencies with a recursive search, do not override the main project dependencies. These can either be listed explicitly in the main project build.act.json or passed on the command line with "--dep foo=../path/to/foo".


plajjan:
Very good!

We don't really have test cases for package management. I don't think we should have automatic ones (I don't want CI to depend on Internet), but perhaps manual ones that one we can run when we actually modify this code. It is fairly self-contained I think so that would acceptable.

PR #2150, opened 2025-02-03
Dont hide fixed constraints

nordlander:


compiler/lib/src/Acton/TypeM.hs
```diff
@@ -295,6 +295,8 @@ typeReport (IncompatError info msg) filename src    =
                                          "Incompatible types"
                                          [(locToPosition (loc info) filename src, This msg)]
                                          []
+typeReport (SurplusRow p) filename src =
+                                    Err Nothing "Too many arguments supplied" [(locToPosition NoLoc filename src, This (prstr (label p)))] []
```
plajjan, line 299: Buuu for NoLoc ;)

nordlander: Indeed! But soon this will change...

PR #2152, opened 2025-02-04
build-debs arm64

mzagozen:


plajjan:
Seems to work, let's see if the apt repos also work...

PR #2157, opened 2025-02-07
Fix arm64 release artifacts

mzagozen:


mzagozen:
Won't see the fix working until this lands in *main*, but I hope this is the last of it ...

PR #2160, opened 2025-02-07
Use --dev to build deps in test cmd

mzagozen:
The main project is built with --dev for test cmd, let's do the same for the deps.


plajjan:
LGTM.

PR #2161, opened 2025-02-07
Make type error location "clickable"

mzagozen:
Use the real filesystem path to the module where a type error was raised.  This means replacing "." with path separator and appending the ".act" suffix. So foo.bar@123 becomes foo/bar.act@123

This makes it possible for VS Code to create a clickable link to the file. Well, almost, because VS Code does not recognize "@" as a file - location separator. Not sure if the diagnostic package can be tweaked to use ":" instead of "@"?

### DIFF
--- a/compiler/actonc/Main.hs
+++ b/compiler/actonc/Main.hs
@@ -876,8 +876,13 @@ handle errKind f src paths mn ex = do
     putStrLn (Acton.Parser.makeReport (f ex) src)
     handleCleanup paths mn
 
+modNameToFilename :: A.ModName -> String
+modNameToFilename mn = joinPath (map nameToString names) ++ ".act"
+  where
+    A.ModName names = mn
+
 handleTypeError opts errKind f src paths mn ex = do
-    printDiag opts $ mkErrorDiagnostic (modNameToString mn) src (typeReport ex (modNameToString mn) src)
+    printDiag opts $ mkErrorDiagnostic (modNameToFilename mn) src (typeReport ex (modNameToFilename mn) src)
     handleCleanup paths mn
 
 handleCleanup paths mn = do
--- a/compiler/actonc/test/typeerrors/ex1.golden
+++ b/compiler/actonc/test/typeerrors/ex1.golden
@@ -1,7 +1,7 @@
 Building file test/typeerrors/ex1.act using temporary scratch directory
   Compiling ex1.act for development
 [error]: Incompatible types
-     +--> ex1@4:5-4:9
+     +--> ex1.act@4:5-4:9
      |
    1 | def f(x,y):
      :     ^ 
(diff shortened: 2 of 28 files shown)


plajjan:
Sounds like an issue to be reported upstream https://hackage.haskell.org/package/diagnose

Diagnose is inspired by Ariadne, a rust library to print pretty error messages. Ariadne seems to be using : to separate file name from location. Not sure why diagnose pick @. Seems like colon would be better. Perhaps a patch could be accepted upstream to just change or offer customizability.

compiler/actonc/Main.hs
```diff
@@ -876,8 +876,13 @@ handle errKind f src paths mn ex = do
...
     handleCleanup paths mn
 
+modNameToErrorString :: A.ModName -> String
+modNameToErrorString mn = joinPath (map nameToString names) ++ ".act"
+  where
+    A.ModName names = mn
+
 handleTypeError opts errKind f src paths mn ex = do
-    printDiag opts $ mkErrorDiagnostic (modNameToString mn) src (typeReport ex (modNameToString mn) src)
+    printDiag opts $ mkErrorDiagnostic (modNameToErrorString mn) src (typeReport ex (modNameToErrorString mn) src)
```
plajjan, line 885:
```suggestion
    printDiag opts $ mkErrorDiagnostic (modNameToFilename mn) src (typeReport ex (modNameToFilename mn) src)
```

plajjan:
I think the change from the internal module path to filenames is valid regardless if it makes clickable links or not, so we can proceed with this particular PR. I made a suggestion to your code.

We have golden tests for typeerrors and syntax errors that are changed because of this. You need to run `make test-typeerrors-accept` to accept the new output as golden, commit and push anew.

The other piece is of course work in upstream. We could vendor the Diagnose library and make local changes if we want to move fast and/or test this out to see how it feels.

A challenge might be that our error format includes multiple locations. One is picked as the primary point where we think the error is, but sometimes it is just hard to pin point the exact error, like it is often the non-alignment between location A and B. Is A or B then the error? I think we tend to point to the use / call site and consider the definition the correct one, but it's probably not 100%... so just saying, that's a challenge. In gcc style errors (from which the clickable link format comes from), stuff just gets rendered quite differently, so clickable links are easier there.

plajjan:
@mzagozen WIP or not? first commit says WIP?

mzagozen:
Ah, that was from before, when I still thought I could change the file path and range separator somewhere in Acton code. Since that is hardcoded in the Diagnose library, we can either:
1. fork it and modify it ourselves like you suggested,
2. or suggest a change to recognize `@` separator in https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/terminalContrib/links/browser/terminalLinkParsing.ts#L75. 

Since you're planning on making other changes to the error message output, option 1?

mzagozen:
We will also need to address the range format itself. If I understand correctly the current range format allows for the error message to reference multiple lines, and is also not something VS Code understands.

PR #2169, opened 2025-02-17
Ensure thread stack size is at least 8 MB

mzagozen:
On Mac OS the default is 524288, which is too low for the recursion depth required by the acton-yang generators. On Linux it is 1 MB, where the current generators work, but it may just come back to haunt us when the model depth increases 👯 We now set the secondary threads stack size to at least 8 MB, or more if the primary thread stack size is larger.

### DIFF
--- a/base/rts/rts.c
+++ b/base/rts/rts.c
@@ -2794,7 +2794,14 @@ int main(int argc, char **argv) {
 
     size_t primary_thread_stack_size;
     size_t target_thread_stack_size;
+#if defined(IS_MACOS)
     primary_thread_stack_size = pthread_get_stacksize_np(pthread_self());
+#else
+    pthread_attr_t attr;
+    pthread_getattr_np(pthread_self(), &attr);
+    pthread_attr_getstacksize(&attr, &primary_thread_stack_size);
+    pthread_attr_destroy(&attr);
+#endif
     target_thread_stack_size = REQUIRED_STACK_SIZE > primary_thread_stack_size ? (size_t)REQUIRED_STACK_SIZE : primary_thread_stack_size;
 
     if (primary_thread_stack_size < target_thread_stack_size)
--- a/base/rts/rts.c
+++ b/base/rts/rts.c
@@ -2792,6 +2792,14 @@ int main(int argc, char **argv) {
 #ifdef ACTON_THREADS
     cpu_set_t cpu_set;
 
+    size_t primary_thread_stack_size;
+    size_t target_thread_stack_size;
+    primary_thread_stack_size = pthread_get_stacksize_np(pthread_self());
+    target_thread_stack_size = REQUIRED_STACK_SIZE > primary_thread_stack_size ? (size_t)REQUIRED_STACK_SIZE : primary_thread_stack_size;
+
+    if (primary_thread_stack_size < target_thread_stack_size)
+        log_warn("Current primary thread stack size: %u, required thread stack size: %u", primary_thread_stack_size, target_thread_stack_size);
+
     // RTS Monitor Log
     pthread_t mon_log_thread;
     if (mon_log_path) {
@@ -2820,7 +2828,18 @@ int main(int argc, char **argv) {
     // thus we need to start 1..num_wthreads. Only need to keep track of
     // branches we start.
     for (int idx = 1; idx <= num_wthreads; idx++) {
-        pthread_create(&threads[idx-1], NULL, main_loop, (void*)(intptr_t)idx);
+        pthread_attr_t stackSizeAttribute;
+        size_t stackSize = 0;
+        pthread_attr_init(&stackSizeAttribute);
+        pthread_attr_getstacksize(&stackSizeAttribute, &stackSize);
+        if (stackSize < target_thread_stack_size)
+        {
+            log_debug("Current thread stack size: %d, required thread stack size: %d", stackSize, target_thread_stack_size);
+            int err = pthread_attr_setstacksize(&stackSizeAttribute, target_thread_stack_size);
+            if (err)
+                log_error("pthread_attr_setstacksize failed: %s", strerror(err));
+        }
+        pthread_create(&threads[idx-1], &stackSizeAttribute, main_loop, (void*)(intptr_t)idx);
         // Index start at 1 and we pin wthreads to CPU 1...n
         // We use CPU 0 for misc threads, like IO / mon etc
         if (cpu_pin) {
--- a/base/rts/rts.c
+++ b/base/rts/rts.c
@@ -2807,10 +2807,22 @@ int main(int argc, char **argv) {
     if (primary_thread_stack_size < target_thread_stack_size)
         log_warn("Current primary thread stack size: %u, required thread stack size: %u", primary_thread_stack_size, target_thread_stack_size);
 
+    pthread_attr_t ss_attr;
+    size_t secondary_thread_stack_size = 0;
+    pthread_attr_init(&ss_attr);
+    pthread_attr_getstacksize(&ss_attr, &secondary_thread_stack_size);
+    if (secondary_thread_stack_size < target_thread_stack_size)
+    {
+        log_debug("Secondary thread stack size: %d, required thread stack size: %d", secondary_thread_stack_size, target_thread_stack_size);
+        int err = pthread_attr_setstacksize(&ss_attr, target_thread_stack_size);
+        if (err)
+            log_error("pthread_attr_setstacksize failed: %s", strerror(err));
+    }
+
     // RTS Monitor Log
     pthread_t mon_log_thread;
     if (mon_log_path) {
-        pthread_create(&mon_log_thread, NULL, $mon_log_loop, (void *)(intptr_t)mon_log_period);
+        pthread_create(&mon_log_thread, &ss_attr, $mon_log_loop, (void *)(intptr_t)mon_log_period);
         if (cpu_pin) {
             CPU_ZERO(&cpu_set);
             CPU_SET(0, &cpu_set);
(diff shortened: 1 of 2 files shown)


mzagozen:
Specifically, this avoids "bus error" on MacOS 15.3.1 when running `make gen` in *respnet*. Before the change we get the following stack trace:

```
* thread #15, name = 'Worker 1', stop reason = EXC_BAD_ACCESS (code=2, address=0x17051fb18)
  * frame #0: 0x0000000100782d38 mplsnet_gen`GC_clear_stack(arg=<unavailable>) at misc.c:365
    frame #1: 0x0000000100774124 mplsnet_gen`GC_generic_malloc_many(lb_adjusted=32, k=1, result=0x000000010159fb80) at mallocx.c:353:17
    frame #2: 0x000000010079d0b0 mplsnet_gen`GC_malloc_kind(lb=24, k=1) at thread_local_alloc.c:198:3
    frame #3: 0x00000001007715c8 mplsnet_gen`GC_malloc(lb=24) at malloc.c:393:10
    frame #4: 0x0000000100101674 mplsnet_gen`GC_calloc(count=1, size=24) at common.c:53:12
    frame #5: 0x00000001000446a8 mplsnet_gen`acton_malloc(size=24) at common.c:110:12
    frame #6: 0x0000000100045b00 mplsnet_gen`to$str(str="_") at str.c:83:9
    frame #7: 0x0000000100523b0c mplsnet_gen`yangQ_schemaQ_get_path_name(dnode=0x00000001018df6e0) at schema.c:4909:207
    frame #8: 0x0000000100533080 mplsnet_gen`yangQ_schemaQ_DLeafD_prdaclass(self=0x0000000101620dc0, loose=0x00000001010dd9a8, top=0x00000001010dd9a8, set_ns=0x00000001010dd9a8, gen_json=0x00000001010dd998, gen_xml=0x00000001010dd998) at schema.c:2686:21
    frame #9: 0x00000001004cd618 mplsnet_gen`yangQ_schemaQ_DNodeInnerD_prdaclass(self=0x00000001016fd300, loose=0x00000001010dd9a8, top=0x00000001010dd9a8, set_ns=0x00000001010dd9a8, gen_json=0x00000001010dd998, gen_xml=0x00000001010dd998) at schema.c:1089:28
    frame #10: 0x00000001004cd618 mplsnet_gen`yangQ_schemaQ_DNodeInnerD_prdaclass(self=0x00000001016fd280, loose=0x00000001010dd9a8, top=0x00000001010dd9a8, set_ns=0x00000001010dd9a8, gen_json=0x00000001010dd998, gen_xml=0x00000001010dd998) at schema.c:1089:28
    frame #11: 0x00000001004cd618 mplsnet_gen`yangQ_schemaQ_DNodeInnerD_prdaclass(self=0x00000001018df790, loose=0x00000001010dd9a8, top=0x00000001010dd9a8, set_ns=0x00000001010dd9a8, gen_json=0x00000001010dd998, gen_xml=0x00000001010dd998) at schema.c:1089:28
    frame #12: 0x00000001004cd618 mplsnet_gen`yangQ_schemaQ_DNodeInnerD_prdaclass(self=0x00000001018df6e0, loose=0x00000001010dd9a8, top=0x00000001010dd9a8, set_ns=0x00000001010dd9a8, gen_json=0x00000001010dd998, gen_xml=0x00000001010dd998) at schema.c:1089:28
    frame #13: 0x00000001004cd618 mplsnet_gen`yangQ_schemaQ_DNodeInnerD_prdaclass(self=0x00000001018df580, loose=0x00000001010dd9a8, top=0x00000001010dd9a8, set_ns=0x00000001010dd9a8, gen_json=0x00000001010dd998, gen_xml=0x00000001010dd998) at schema.c:1089:28
    frame #14: 0x00000001004cd618 mplsnet_gen`yangQ_schemaQ_DNodeInnerD_prdaclass(self=0x00000001016fd100, loose=0x00000001010dd9a8, top=0x00000001010dd9a8, set_ns=0x00000001010dd998, gen_json=0x00000001010dd998, gen_xml=0x00000001010dd998) at schema.c:1089:28
    frame #15: 0x00000001004cd618 mplsnet_gen`yangQ_schemaQ_DNodeInnerD_prdaclass(self=0x0000000101626900, loose=0x00000001010dd9a8, top=0x00000001010dd998, set_ns=0x00000001010dd998, gen_json=0x00000001010dd998, gen_xml=0x00000001010dd998) at schema.c:1089:28
    frame #16: 0x000000010037acc0 mplsnet_gen`orchestronQ_buildQ_L_25C_31cont(idx=0x00000001010db060, layer=0x00000001016aa0e0, tttsrc=0x00000001016f4f40, W_33=0x00000001010de4f0, self=0x00000001015993f0, W_162=0x00000001010de600, W_182=0x00000001010de918, modname=0x00000001016f4bc0, output_dir=0x00000001016f4f80, tttsrc_getlayers=0x00000001016f4d20, wfc=0x000000010158eed0, N_iter=0x00000001016f4ce0, W_383=0x00000001016f3f80, W_385=0x00000001010de4e8, imports=0x00000001016f4d60, C_cont=0x00000001016f4fa0, W_329=0x00000001010de528, C_32res=0x000000010159de00) at build.c:1907:17
    frame #17: 0x000000010037c010 mplsnet_gen`orchestronQ_buildQ_L_48ContD___call__(L_self=0x000000010159cdc0, G_1=0x000000010159de00) at build.c:1949:12
    frame #18: 0x000000010010e410 mplsnet_gen`wt_work_cb(ev=0x0000000101108748) at rts.c:1558:17
    frame #19: 0x0000000100aa57a8 mplsnet_gen`uv__run_check(loop=0x0000000101590540) at loop-watcher.c:67:1
    frame #20: 0x0000000100a7be00 mplsnet_gen`uv_run(loop=0x0000000101590540, mode=UV_RUN_DEFAULT) at core.c:462:5
    frame #21: 0x0000000100113afc mplsnet_gen`main_loop(idx=0x0000000000000001) at rts.c:1753:13
    frame #22: 0x0000000100793efc mplsnet_gen`GC_pthread_start_inner(sb=0x00000001705a2f90, arg=0x000000016fdfd7d8) at pthread_start.c:55:12
    frame #23: 0x0000000100788250 mplsnet_gen`GC_call_with_stack_base(fn=(mplsnet_gen`GC_pthread_start_inner at pthread_start.c:45), arg=0x000000016fdfd7d8) at misc.c:2420:12
    frame #24: 0x0000000100798820 mplsnet_gen`GC_pthread_start(arg=0x000000016fdfd7d8) at pthread_support.c:2675:10
    frame #25: 0x00000001954582e4 libsystem_pthread.dylib`_pthread_start + 136
```

plajjan:
Wow, good find!

base/rts/rts.c
```diff
@@ -2820,7 +2820,18 @@ int main(int argc, char **argv) {
     // thus we need to start 1..num_wthreads. Only need to keep track of
     // branches we start.
     for (int idx = 1; idx <= num_wthreads; idx++) {
-        pthread_create(&threads[idx-1], NULL, main_loop, (void*)(intptr_t)idx);
+        pthread_attr_t stackSizeAttribute;
```
plajjan, line 2823: How is the stack size managed for the main thread? Is it always big enough?

We normally don't use the programs main thread for running the program, main thread runs timer queue and stuff like that. But if you `acton build --no-threads` you should get a program without thread support and then you are bound to the stack size of the main thread. What happens then for the respnet get use case?

plajjan:
I read that macos uses 512kb on secondary threads whereas the main thread has 8MB per default. Feels quite outdated to do 512 but above all, for us all threads should be equal. Maybe the better solution is to check stack size of main thread, ensuring that it is above our minimum (but could be more if OS default is higher) then copy the stack size for the main thread to secondary threads.

base/rts/rts.c
mzagozen: Seems so. It works and is faster without threads too ;)

plajjan: How much faster? :)


Did you see my comment below on resizing based on main thread stack size?

mzagozen: The result was influenced by using a *release* build too. Release with `--no-threads` runs through the same set of YANG modules in 0.5s, release with threads in 0.8, and the *development* builds in 3s (no threads) and 4s (threads) respectively.

I did see your comment, but I think I may be getting the main thread size incorrectly, because the value I read with `getrlimit(RLIMIT_STACK, &rl)` is 67092480.

plajjan: getrlimit is the limit, i.e. max stack size, not the actual stack size.

`pthread_attr_getstacksize` perhaps? - https://pubs.opengroup.org/onlinepubs/009696699/functions/pthread_attr_getstacksize.html

mzagozen: Yes, no GC, release build, threaded and non-threaded are equivalent.

plajjan: Ok, very good. We know the GC is slow. Thanks!

base/rts/rts.c
```diff
@@ -2792,10 +2792,37 @@ int main(int argc, char **argv) {
...
+        int err = pthread_attr_setstacksize(&ss_attr, target_thread_stack_size);
+        if (err)
+            log_error("pthread_attr_setstacksize failed: %s", strerror(err));
+    }
+
     // RTS Monitor Log
     pthread_t mon_log_thread;
     if (mon_log_path) {
-        pthread_create(&mon_log_thread, NULL, $mon_log_loop, (void *)(intptr_t)mon_log_period);
+        pthread_create(&mon_log_thread, &ss_attr, $mon_log_loop, (void *)(intptr_t)mon_log_period);
```
plajjan, line 2825: Just FYI, the RTS monitor thread do not run Acton continuations, only some functions written in C, so they really don't require a large stack space. Though it certainly doesn't hurt either.

I'd actually like to remove the RTS mon thread overall...

PR #2172, opened 2025-02-18
Add glob pattern to file.listdir()

mzagozen:
Internally uses fnmatch(), which supports shell-style wildcards.

### DIFF
--- a/base/src/file.act
+++ b/base/src/file.act
@@ -256,7 +256,7 @@ actor FS(cap: FileCap):
         mkdir(new_tmp_dir)
         return new_tmp_dir
 
-    action def listdir(path: str) -> list[str]:
+    action def listdir(path: str, glob: ?str=None) -> list[str]:
         """List directory contents"""
         NotImplemented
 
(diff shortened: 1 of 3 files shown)


plajjan:
A guess on why CI testing is failing: fnmatch is not supported on Windows.

Since the matching is done after the listing, there is no inherent benefit to doing this in C. We might as well use an Acton function to glob filter the result. I think that's better, since it also automatically becomes more portable and it also becomes useful in other contexts - maybe you got a list of file names and now you want to glob match.

If we had pushed the glob matching down to some syscall or whatever, then it would have been different as we could gain quite a bit of efficiency, but seeing how that is not the case, I'd say write it in Acton. Glob matching isn't that hard so I expect no more than a page of code to implement it, no?

mzagozen:
> A guess on why CI testing is failing: fnmatch is not supported on Windows.

You are right, there is no fnmatch.h for Windows.

> [...] I'd say write it in Acton. Glob matching isn't that hard so I expect no more than a page of code to implement it, no?

True. I used fnmatch.h because it was convenient and it required the least amount of code. But now that we have no choice but to roll our own, it gives a chance to make it even more useful. For example the [glob pattern](https://docs.python.org/3/library/pathlib.html#pathlib-pattern-language) in Python supports the equivalent glob pattern matching with the additional ability to recursively descent into subdirectories.

PR #2176, opened 2025-02-20
Fix 2140

sydow:
Dubious (temporary) fix. See comment to zip in builtin/builtin_functions.c


mzagozen:
Can confirm it works for me too!

Issue #2135, opened 2025-01-29
Missing error location for constraint


mzagozen:
### Acton Version

0.24.1.20250129.8.25.26

### Steps to Reproduce and Observed Behavior

```
class Foo(object):
    thing: ?bool

    def __init__(self, thing):
        self.thing = thing

mut def remove_thing(foos):
    for f in foos:
        f.thing = None

def process():
    p = []
    foos = [Foo(True)]
    # This is obviously wrong, remove_thing() does not return Iterable
    p.extend(remove_thing(foos))
```

```
[error]: Constraint violation
     ╭──▶ @0:0-0:0
     │
   0 │ <no line>
     •
     • ╰╸ None must be a subclass of __builtin__.list[_]
─────╯
```

### Expected Behavior

Error points to somewhere in the line `p.extend(remove_thing(foos))`

sydow:
Fixed by #2181 .

plajjan:
Now:
```
kll@Kristians-MacBook-Air acton % actonc compiler/actonc/test/typeerrors/ex26.act
Building file compiler/actonc/test/typeerrors/ex26.act using temporary scratch directory
  Compiling ex26.act for release
[error]: Constraint violation
     ╭──▶ ex26.act@15:5-15:33
     │
  15 │     p.extend(remove_thing(foos))
     •     ┬───────────────────────────
     •     ╰╸ None must be a subclass of __builtin__.list[_]
─────╯
kll@Kristians-MacBook-Air acton %    
```

PR #2181, opened 2025-02-26
fixed #2132 and #2135

### LINKED ISSUE #2132: Missing error location for "keyword component(s) ... is missing in tuple"
### Acton Version

0.24.1.20250127.23.53.9

```
actor Bob():
    def foo(cb):
        pass

actor Rob():
    def bar():
        b = Bob()
        b.foo()
```

```
[error]: Incompatible types
     ╭──▶ @0:0-0:0
     │
   0 │ <no line>
     • 
     • ╰╸ keyword component(s) 'cb' is missing in tuple
─────╯
```

### Expected Behavior

Error message points to the call of `Bob.foo()` with the missing argument

sydow:


plajjan:
It gives me such great joy seeing these fixed, thanks Björn @sydow! =)

PR #2189, opened 2025-03-05
Fix isinstance on optional types

plajjan:


plajjan:
@nordlander here is a quick fix to not segfault on a NULL value when isinstance checking.

I initially attempted to also handle `if isinstance(foo, None)` but the compiler thinks this is illegal, so I skipped it for now.

nordlander:
👍

PR #2194, opened 2025-03-12
Fix base64.decode buffer allocation, UTF-8 handling, change input/output to bytes

mzagozen:
We must use decoder.calcSizeForSlice() instead of calcSizeUpperBound() to get an exact number of bytes we will decode, without padding. If we allocate the maximum size including padding, we may decode less bytes and then our decoded buffer may end up containing garbage.

We also align with other languages (including Python) on the input and output types for encode and decode - the functions now work with Acton bytes instead of str.

### DIFF
--- a/base/__root.zig
+++ b/base/__root.zig
@@ -3,7 +3,7 @@ const std = @import("std");
 const acton = @import("acton.zig");
 const gc = @import("rts/gc.zig");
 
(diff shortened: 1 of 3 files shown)


mzagozen:
The `perf-vs-main` job runs the tests from this branch using Acton from main (tip), right? Then it is expected to fail here ...

plajjan:
Awesome =)

PR #2203, opened 2025-03-18
Fix method calls on empty string

mzagozen:


plajjan:
LGTM. I'll merge this now. @sydow feel free to read the changes and let us know if anything looks broken.

PR #2207, opened 2025-03-20
Fix aarch64 linux cross compile

mzagozen:
When cross-compiling on amd64 host for arm64, Zig picks the *baseline* default CPU. But that is not sufficient to compile Mbed-TLS, it needs AES. For other cross-compilation targets we already provide a default CPU model with the required instructions. We now use *[cortex_a72](https://en.wikipedia.org/wiki/ARM_Cortex-A72)* CPU for cross-compiling for `aarch64-linux-*`. This is the first AWS Graviton from 2018. Not sure if we should target something more modern? I guess it does make a difference in the Zig compiler output?

I also found that using *generic+crytpo* works for all cross-compile targets, but again, what does that mean in terms of other optimizations?

### DIFF
--- a/cli/src/acton.act
+++ b/cli/src/acton.act
@@ -1425,6 +1425,7 @@ actor main(env):
         p.add_option("tempdir", "str", "?", "", "Directory for temporary build files")
         p.add_option("syspath", "str", "?", "", "syspath")
         p.add_option("target", "str", "?", "", "Target, e.g. x86_64-linux-gnu.2.28")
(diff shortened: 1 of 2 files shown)


plajjan:
TIL about `generic+crypto` - sounds like exactly what we want... so let's do that instead? I'm fine with this PR, so feel free to merge if you just want to move forward right now, but I think `generic+crypto` sounds like the better solution (since we don't need to care about x86 vs aarch64, right?)

PR #2220, opened 2025-03-30
Support f-strings

### LINKED ISSUE #2050: Support f-string formatting
We should add support for using f-strings for formatting! It's the modern format and in fact so popular (which in this case I think is an indicator for really good & intuitive) that other languages have mimicked Python, like the C++ fmt library etc.

```
name = "world"
message = f"Hello {name}!"

# Big integers
big_num = 12345678901234567890
message = f"The value is {big_num}"  # Beyond i64 range

# Expressions
base = 2
exp = 10
result = f"{base}^{exp} = {base**exp}"  # "2^10 = 1024"
```

plajjan:
This adds support for f-strings. Under the hood it just rewrites it to use % operator style. We largely treat everything as a string, which greatly simplifies things since we are doing all this as a rewrite in the Parser where we do not yet have type information.

Fixes #2050

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -202,7 +202,8 @@ jobs:
         run: |
           export DEBIAN_FRONTEND=noninteractive
           apt-get update
-          apt-get install -qy bzip2 curl g++ gdb haskell-stack make procps zlib1g-dev
+          apt-get install -qy bzip2 curl g++ haskell-stack libtinfo-dev make procps zlib1g-dev
+          apt-get install -qy gdb
       - name: "locale en_US.UTF-8"
         run: |
           apt-get install -qy locales
@@ -264,7 +265,7 @@ jobs:
       - name: "Install build prerequisites"
         run: |
           apt-get update
-          apt-get install -qy bzip2 curl g++ haskell-stack make procps zlib1g-dev
+          apt-get install -qy bzip2 curl g++ haskell-stack libtinfo-dev make procps zlib1g-dev
           apt-get install -qy bash-completion build-essential debhelper devscripts
       - name: "Check out repository code"
         uses: actions/checkout@v4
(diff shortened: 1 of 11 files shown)


plajjan:
@nordlander PTAL

plajjan:
@nordlander I also added my new test suite for the compiler to this branch. It only has test cases focused on the new f-strings for now, but I think it is suitable for being expanded upon with all sorts of tests. It doesn't show the idea we talked about doing golden checking for input+output of later passes in the compiler, but it does show just based Haskell-based unit testing of our parser!

nordlander:
Looks good overall, and appears to work. But the use of an accumulating parameter that's extended by `acc ++ [c]` isn't ideal in Haskell as it can be highly inefficient, especially if the accumulator grows large. Better then to build the accumulator in reverse by writing `c : acc`, and undoing the reversal in one sweep by writing `reverse acc` when the accumulator is returned.

I also suspect that position info inside the embedded expressions ends up wrong since a completely new parser is started for each such expression. This means that the file offset and line number counters are locally reset, which might lead to confusing error messages. Maybe it would be worthwhile to try running the parser `expr` directly on the current input each time a complete `TextPart` fragment has been consumed? That ought to open up for simplifications to code that currently matches braces as well...

plajjan:
@nordlander agree with ++ being bad and we should of course keep track of our location, so using new parser is bad. In parallel to this PR I have also experimented quite a bit with getting good error messages out of this parser and it requires substantive changes. Writing a correct parser is one thing, writing a correct parser with good error messages for invalid data is an entirely different thing. Thus, my view is that this whole f-string parser needs to be rewritten pretty much from scratch.

I will merge this PR as-is, with it's potential slowness (not too bad on short f-strings) and close-to-non-existent-error-message and not good source location. But we will get f-strings. We also get a unit test suite for the parser which we can expand upon. I will open an issue for these improvements.

I have rebased and force pushed, will set to merge.

PR #2221, opened 2025-03-30
Don't require newline at end of file

plajjan:
There's lots of tools that strive to not write newlines at the end of a file. I have seen editors do this and lately I noticed AI agents produce files without a newline. The Acton parser throws an error at this, which seems excessive.. so now change that. The fix here is to just add a newline if one isn't present. That's a little hacky but simpler than changing parsing rules. I can't really see any drawbacks to this.


plajjan:
@nordlander PTAL

PR #2225, opened 2025-04-03
Expand acton unit test suite

plajjan:
I have expanded the unit test suite for the acton compiler library. The purpose at this point is to show what the concept can look like. We use plain Acton program source as input and then run that through the compiler steps up to the and including the compiler we want to test, in this case the CPS pass. We golden test both the input to the CPS pass as well as the output. This way, it is easy to notice if there are changes in previous passes that affect the output or if there are changes in the particular pass we want to test.

The cps_volatiles test is probably quite a poor test to apply like this, but it's really just a placeholder. We should add in other tests.

I have opted to use a single large test file, ActonSpec.hs. We don't have enough tests at the moment or in the near future to warrant any sort of split and I don't think it is obvious what such a split should look like. I can imagine cases when we want to group related tests and those could be grouped along a number of imaginable lines.


plajjan:
@nordlander ok, I'm pretty happy with this now. PTAL, what do you think?

PR #2255, opened 2025-05-07
Hashing

plajjan:


plajjan:
When I define a default implementation of the `hash_bytes` method in the Hashable protocol, I'm getting compilation errors:

```
install
+- install Acton
   +- zig build-lib Acton Debug aarch64-macos-none 6 errors
/Users/kll/dt/acton/dist/base/out/types/acton/rts.c:878:31: error: call to undeclared function 'B_HashableD_intG_new'; ISO C99 and later do not support implicit function declarations
/Users/kll/dt/acton/dist/base/out/types/acton/rts.c:878:19: error: cast to 'B_Hashable' (aka 'struct B_Hashable *') from smaller integer type 'int'
/Users/kll/dt/acton/dist/base/out/types/http.c:8026:33: error: call to undeclared function 'B_HashableD_strG_new'; ISO C99 and later do not support implicit function declarations
/Users/kll/dt/acton/dist/base/out/types/http.c:8026:21: error: cast to 'B_Hashable' (aka 'struct B_Hashable *') from smaller integer type 'int'
/Users/kll/dt/acton/dist/base/out/types/http.c:8094:33: error: call to undeclared function 'B_HashableD_strG_new'; ISO C99 and later do not support implicit function declarations
/Users/kll/dt/acton/dist/base/out/types/http.c:8094:21: error: cast to 'B_Hashable' (aka 'struct B_Hashable *') from smaller integer type 'int'
error: the following command failed with 6 compilation errors:
```

@nordlander do you see why that is? I have pushed some commits, so just uncomment the hash_bytes method in there and recompile whole acton and you should also get this error.

I can see how the generated C code in dist/base/out/types/__builtin__.c does change depending on this and are indeed missing the mentioned functions... but why are they missing, how does a default implementation of hash_bytes influence this?

plajjan:
I created this branch / PR from @sydow newhash branch. The differences are:
- @sydow had removed the wyhash module from stdlib - I removed that, so we keep the wyhash module
- the hashing.c files are kept, not removed like @sydow did, we believe some functions are used in telemetrify - should be cleaned up later though
- I renamed putBytes to hash_bytes after a conversation with @nordlander 
- I fixed passing bytes to Zig code - now doing proper slice based on the explicit nbytes len in our bytes, not C-string NULL sentinel terminated.

@sydow FYI - hope you are fine with these changes..

sydow:
My comment above was badly expressed. Adding a default for hash_bytes does not hinder a user from defining __hash__ in some non-recommended way, even if we discourage it in some way.  But it prevents other types from defining hash_bytes for their types which has a component of the present type. So we should not provide a default do-nothing version of hash_bytes. The non-conforming user may still define __hash__ in some stupid way and hash_bytes as doing nothing...

nordlander:
@plajjan: The compilation errors in the presence of a default `hash_bytes` implementation were caused by a decision long ago to replace a `NotImplemented` extension body by `NotImplemented` definitions of the abstract methods. A good idea as such, but not sufficient to inform the code generator about C dependencies when all methods of the current class actually _have_ valid default implementations. I fixed this by simply leaving `NotImplemented` in the translated class body, in conjunction with any added `NotImplemented` methods. An more fundamental solution would have been to generate method definitions for missing _inherited_ methods as well, but that will have to wait (they just result in C function headers anyway). And the preserved annotation in the class body is nevertheless a correct piece of info.

plajjan:
@sydow ok, thanks for quick and good feedback. Dangerous to let two hash-amateurs on the loose ;)

Agreed, we should not let hash_bytes have a default implementation since that would break havoc since everyone needs to rely on hash_bytes of others is correctly implemented. However, as we are really saying this is the one and true way of hashing, it seems that we should also remove the possibility of customizing that. @nordlander and I talked about it and agree that we should really just turn the global `hash()` function into setting up the hasher and those bits currently in `__hash__`, then we can remove `__hash__` completely and `hash_bytes` can be named `hash`. That also aligns perfectly with the Rust Hash trait and Swift Hashable protocol which I already believe were sources of inspiration for your proposal, which surely is a benefit, right?

PR #2260, opened 2025-05-29
Include diff in testing.assertEqual(a, b) exception

mzagozen:
If the `testing.assertEqual(a, b)` assertion fails, we now include the diff output in the exception message. That is in addition to the A, B values which are also printed in full.

I find that in this case the full output is also useful if one wants to copy & paste the output and update the expected value in the test.

### DIFF
--- a/base/src/testing.act
+++ b/base/src/testing.act
@@ -16,17 +16,21 @@ class NotEqualError[T](AssertionError):
     a: ?T
     b: ?T
 
-    def __init__(self, a, b, msg: ?str=None, print_vals: bool=True):
+    def __init__(self, a, b, msg: ?str=None, print_vals: bool=True, diff_str: ?str=None):
         self.a = a
         self.b = b
         self.error_message = msg if msg != None else "Expected equal values but they are non-equal."
         self.print_vals = print_vals
+        self.diff_str = diff_str
 
     def __str__(self):
         msg = "%s: %s" % (self._name(), self.error_message)
         if self.print_vals:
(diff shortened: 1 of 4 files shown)


plajjan:
We have some pretty large output (golden testing) where the full output is just too much, it's pages and pages long and so we just print the diff. That's handled more directly in the test runner code though, as you've probably noted (clearly, otherwise it would have been in assertEqual which you are now extending).

I did think about putting it in assertEqual but couldn't decide on whether to default to diffs or full A & B values. One option would be to have a heuristic, like if there are more than X lines, then print diff only. But I also feel like I'm colouring the generic assertEqual a little too much compared to what really belongs in the `acton test` functionality... so yeah, no real guidance here, I'm just rambling on about some of my thoughts in this area.

I'm OK with this PR, feel free to merge

mzagozen:
> I did think about putting it in assertEqual but couldn't decide on whether to default to diffs or full A & B values. One option would be to have a heuristic, like if there are more than X lines, then print diff only. But I also feel like I'm colouring the generic assertEqual a little too much compared to what really belongs in the acton test functionality... so yeah, no real guidance here, I'm just rambling on about some of my thoughts in this area.

That was my train of thought as well ... If I do use strings for `testing.assertEqual()`, then the length is typically limited to tens of lines, not hundreds like the golden outputs. If the test fails it is convenient to have the full output available for copy pasting to update the test code.

I've been sitting on this branch for a while now, time to merge :)

PR #2271, opened 2025-06-09
Add docstring support to compiler type system

plajjan:
NameInfo was recently expanded to contain docstring fields for functions, classes, actors, protocols, extensions, and modules. Docstrings are now:
- Extracted from AST during type checking and placed in docstring attribute
- Persisted in .ty interface files
- Displayed when viewing .ty files with actonc
- The .ty file content has changed from just a TEnv to a complet NModule to naturally make room for storing the module docstring

Added test coverage in syntax1.act demonstrating docstrings for all supported language constructs.

### DIFF
--- a/compiler/lib/src/Acton/Env.hs
+++ b/compiler/lib/src/Acton/Env.hs
@@ -164,7 +164,7 @@ prettyOrPass te
 
 prettyDocstring :: Maybe String -> Doc
 prettyDocstring Nothing         = empty
-prettyDocstring (Just docstring) = text "\"\"\"" <+> text docstring <+> text "\"\"\""
+prettyDocstring (Just docstring) = text "\"\"\"" <> text docstring <> text "\"\"\""
 
 instance Pretty WTCon where
     pretty (ws,u)               = --dotCat prettyW ws <+> colon <+>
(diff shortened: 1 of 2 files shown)


nordlander:
Looks good overall. One recurring pattern can be simplified, though:
```
    case nmod of
        NModule te mdoc -> ... do something with te and mdoc ...
         _ -> error $ "Invalid module info in ..."
```
Since the potential error here "should not happen", I think it's cleaner to just write
```
    let NModule te mdoc = nmod
    ... do something with te and mdoc ...
```
That is, no need to supply any explicit error handling in these cases, because the Haskell rts will just halt with a message similar to `error` should `nmod` not match the left-hand pattern at run-time.

I'd also suggest swapping the pattern order in local function `stripQuotes` in `Env`, to avoid having the single-quote patterns match even the triple-quotes.

plajjan:
@nordlander hmm, I've been looking into the \n thing. prettyPrint uses show which I guess does the rather raw representation. What do you expect for newlines in docstrings when looking at a .ty file ?

plajjan:
I've addressed all the feedback, removed superfluous error handling (relying on RTS unwind), fixed the pattern match order. Had a call with @nordlander and realized that we store strings in an escaped form, so now doing unescape. This happens in Types.hs when lifting over the docstring from a AST statement to NameInfo docstring field. This way TEnv / NameInfo is ready for pretty printing.

PR #2280, opened 2025-06-16
Fix string 'in' operator for substrings found at position 0

mzagozen:
The __contains__ method was using bmh() > 0 which incorrectly returned false when a substring was found at the beginning of the string (position 0). Changed condition to bmh() >= 0 to include position 0 matches.


plajjan:
LGTM, merging.

PR #2288, opened 2025-06-18
Add more meat to ecolift

mzagozen:
Add more projects to the ecosystem. Explain how to discover new feature syntax. Do not modify outputs of code generators.


ecolift/CLAUDE.md
```diff
@@ -12,21 +12,34 @@ The following repositories contain syntax highlighting and similar for editors /
...
+- git@github.com:orchestron-orchestrator/netclics.git
+
+The Telemetrify application stack consists of the Telemetrify application
+(telemetrify) and its supporting protocol libraries.
 
 - git@github.com:telemetrify-collector/telemetrify.git
 - git@github.com:actonlang/acton-grpc.git
 - git@github.com:actonlang/acton-http2.git
+- git@github.com:actonlang/acton-snappy.git
+- git@github.com:actonlang/acton-zlib.git
```
plajjan, line 38: I placed the telemetrify related apps here more because they aren't as mature yet, so might not work and so it's better to start with the other ones when ecolifting.

zlib is pretty much feature complete and stable. Same for snappy, so I think they can go higher on the list.

PR #2290, opened 2025-06-18
Use f-strings in stdlib

plajjan:
Convert % formatting and string concatenation to f-strings across stdlib.


plajjan:
Noticed this isn't complete, but let's merge this for now.

PR #2294, opened 2025-06-19
process: inherit PATH when custom env provided

mzagozen:
When spawning processes with custom environment variables, automatically inherit PATH from parent process unless explicitly overridden. This ensures execvp() can resolve executables beyond default system paths while maintaining backward compatibility.

Fixes issues where processes with custom environments fail to find executables installed outside /bin:/usr/bin (e.g., /usr/local/bin).


mzagozen:
It would be more elegant to implement this in process.act, but env vars are only available via the `env` cap - `env.getenv("PATH")`. The Process actor gets `ProcessCap` with no environment info and no `env` cap, so we instead bypass the restriction in C 🤷

plajjan:
I get what you are aiming for; to only override PATH in the one situation where it matters, but it's quite hard to understand reading the text. The user has to learn that there is this one situation (when we have a bare command) where PATH is copied over into the new environment and so if the user cares about PATH, they need to pay attention.

ISTM that it's much easier to always inherit PATH into the new environment unless PATH is explicitly set by the user. Simpler logic, easier to explain. Just simpler. Anyone who cares about cleaning up PATH merely needs to remember to explicitly set it.

plajjan:
Also, we shouldn't be modifying arguments...

Issue #2278, opened 2025-06-16
Improve error message for actor method or init argument named self


mzagozen:
# `self` in actor

## `self` as actor method argument
Say we define an actor with a method where one of the arguments is named `self` - either the first one:

```acton
actor test():
    # self could also be first
    def test_method(first, self):
        pass
```

Then the compiler will print a very cryptic error message:

```
Building file test_impl.act using temporary scratch directory
  Compiling test_impl.act for release
actonc: ### qSchema Dot unexpected self.test_methodG_local  ::  None
CallStack (from HasCallStack):
  error, called at src/Acton/QuickType.hs:95:46 in acton-0.26.0-FQsbyVAusTe2kJMcb06rDv:Acton.QuickType
actonc exited with code:  1  terminated with signal: 0
```

## `self` as actor init argument
Say we define an actor with one of the init arguments named `self`:

```acton
actor test(first, self):
    def test_method():
        pass
```

Then the generated C code will include a duplicate `self` parameter in generated `__init__` function:

```c
$R test_implQ_testD___init__ (test_implQ_test self, $Cont C_cont, B_NoneType first, B_NoneType self) {
    return $R_CONT(C_cont, B_None);
}
```

# Solution

Not sure what the correct course of action is, but either:
1. The compiler message should be improved, saying that `self` is not allowed for actor methods or init arguments. 
2. The `self` argument in actor methods and init arguments should be allowed and not conflict with internals. For init args, maybe use the same kind of namespacing technique used for protecting `class` -> `$class`?

I am leaning towards 1 to avoid creating confusion with "object" `self.X` method and attribute access vs. actor where it's just `X`.

plajjan:
@nordlander if we are to improve the error message, do you think that should be done from the parser or type checking pass?

I did a first attempt to modify the parser but it doesn't feel quite right, so I'm pondering whether doing it in types is more natural

plajjan:
@mzagozen I added some docs at least - https://acton.guide/actors/self.html - WDYT?

PR #2302, opened 2025-06-20
Document "self" in actor

mzagozen:
Part of #2278.


docs/acton-by-example/src/functions/actor_methods.md
```diff
@@ -31,6 +35,47 @@ actor main(env):
...
-Multiplying 3 with 42
+Calculator multiplying 3 with 42
 Result: 126
 ```
+
+## The implicit `self` in actors
+
+Unlike classes where methods must explicitly declare `self` as their first parameter, actors have an implicit `self` that is automatically available. This fundamental difference means that actor methods are defined without a `self` parameter. In fact, using `self` as a parameter name in actor methods or in the actor's initialization arguments will result in an error.
+
+The implicit `self` in an actor refers to the actor instance itself. This allows an actor to pass a reference to itself to another actor by simply passing `self`. This is particularly useful when implementing callback patterns or when an actor needs to register itself with another actor.
```
plajjan, line 46: I don't quite like this wording. It focuses on `self` is implicit, but `self` is just not needed to refer to attributes on the own actor.

The sole purpose of `self` in actor is to get an external view of the actor that we can pass to other actors that need to interact with us. I think we can express this much more concisely.

PR #2304, opened 2025-06-21
drop __hash__  

plajjan:


plajjan:
I really meant to merge this when I merged #2255. We don't just want the new hashing, we actively want to remove the old one too.

PR #2317, opened 2025-06-23
Add type error for 'self' parameter in actors

### LINKED ISSUE #2278: Improve error message for actor method or init argument named self

plajjan:
Actors and methods within actors cannot have 'self' as a parameter name since it's implicitly available. Add checks in the type checking phase with a helpful error message explaining this restriction.

- Check actor parameters and method parameters in checkEnv
- Use Control.Exception.throw instead of throwError because the error is thrown inside runTypeM which catches all TypeM exceptions and converts them to crashes. By using Control.Exception.throw, we bypass the TypeM monad error handling and let the exception propagate to Main.hs where it's properly caught and formatted with Diagnose.
- Add test cases err43 and err44 in typeerrors/

Fixes #2278


plajjan:
I'm going to go ahead and merge this. I think it's reasonably simple and non-intrusive. We can always revisit if we think it can be improved and made even less intrusive.

PR #2323, opened 2025-06-26
Fixed "2320"

sydow:
fixes#2320


sydow:
Easy fix: one call to "norm" (in the body of an actor definition) should be to "normSuite"...

PR #2325, opened 2025-06-29
Normalize nested primFORMAT calls

plajjan:
String interpolation is handled by the $FORMAT macro in C, which translates to like snprintf and some other statements. Macro expansion isn't recursive, so we cannot have recursive $FORMAT calls and thus must expand it first.

The normalizer now extracts nested interpolated string calls (primFORMAT) into temporary variables. This fixes code generation for nested interpolated strings like:

  "User: {"Name: "{name}}"

which now becomes:

  N_fmt = $FORMAT("Name: %s", (name,))
  msg = $FORMAT("User: %s", (N_fmt,))

The solution works recursively for arbitrary nesting depth and handles formats in all expression contexts (assignments, returns, function args, etc).

- Add extractFormats function to recursively find nested primFORMAT calls
- Extract nested formats into temporary variables with sequential statements
- Handle format extraction uniformly through normStmt entry point
  - Refactored flat pattern match into new normStmt function that makes it easier to call extraction / normalization through fewer code paths
- Support all statement contexts where expressions can appear:
  - Return statements
  - Assignments (regular, mutation, variable)
  - Expression statements
  - Control flow conditions (if, while)
  - For loop iterators
  - After statement expressions
  - Comprehension conditions

### DIFF
--- a/compiler/lib/src/Acton/Normalizer.hs
+++ b/compiler/lib/src/Acton/Normalizer.hs
@@ -279,30 +279,37 @@ instance Norm Stmt where
     norm' env s@(Break l)           = return $ exitContext env s
     norm' env s@(Continue l)        = return $ exitContext env s
     norm' env (Return l Nothing)    = return $ exitContext env (Return l $ Just eNone)
-    norm' env (Return l (Just e))   = do e <- norm env e
-                                         case isVal e of
-                                             True -> return $ retContext e
+    norm' env (Return l (Just e))   = do e' <- norm env e
+                                         (stmts, e'') <- extractFormats env e'
+                                         case isVal e'' of
+                                             True -> return $ stmts ++ retContext e''
                                              False -> do
                                                  n <- newName "tmp"
-                                                 return $ sAssign (pVar n $ conv t) e : retContext (eVar n)
+                                                 return $ stmts ++ (sAssign (pVar n $ conv t) e'' : retContext (eVar n))
       where retContext e            = exitContext env $ Return l $ Just e
             t                       = typeOf env e
 
     norm' env (Assign l ps e)       = do e' <- norm env e
+                                         (formatStmts, e'') <- extractFormats env e'
                                          (ps1,stmts) <- unzip <$> mapM (normPat env) ps
                                          ps2 <- norm env ps1
                                          let p'@(PVar _ n _) : ps' = ps2
-                                         return $ Assign l [p'] e' : [ Assign l [p] (eVar n) | p <- ps' ] ++ concat stmts
+                                         return $ formatStmts ++ (Assign l [p'] e'' : [ Assign l [p] (eVar n) | p <- ps' ] ++ concat stmts)
       where t                       = typeOf env e
     norm' env s@(For l p e b els)   = do i <- newName "iter"
                                          v <- newName "val"
-                                         norm env [sAssign (pVar i $ conv t) e,
-                                                   handleStop (While l (eBool True) (body v i) []) els]
+                                         e' <- norm env e
+                                         (stmts, e'') <- extractFormats env e'
+                                         rest <- norm env [sAssign (pVar i $ conv t) e'',
+                                                         handleStop (While l (eBool True) (body v i) []) els]
+                                         return (stmts ++ rest)
       where t@(TCon _ (TC c [t']))  = typeOf env e
             next i                  = eCall (eDot (eVar i) nextKW) []
             handleStop loop els     = Try l [loop] [Handler (Except l0 qnStopIteration) (mkBody els)] [] []
             body v i                = sAssign (pVar v t') (next i) : sAssign p (eVar v) : b
-    {-
+
+    norm' env (With l [] b)         = norm env b
+{-
     with EXPRESSION as PATTERN:
         SUITE
     ===>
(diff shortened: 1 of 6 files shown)


nordlander:
Hmm. I don't think it's sufficient to just check a subset of the statements, for a subset of the expression forms, to capture all nested calls to $FORMAT. Nearly all statements can contain expressions, and most expressions also contain sub-expressions. So basically the whole Acton syntax needs to be scanned for a generic pattern like this, which essentially is what the Normalizer is all about.

Here's what I think is needed:
1. A flag in the Normalizer environment that tells whether it's traversing an argument of $FORMAT or not.
2. An extra case where the Normalizer matches a Call expression, which checks whether the function head is $FORMAT (and if so, updates the flag before normalizing its arguments).
3. A sub-case to the above pattern, that captures a call to $FORMAT inside an argument to another call to $FORMAT. This is the case that needs to be transformed, by binding the inner call to a fresh variable and lifting out that binding to the closest surrounding statement sequence preceding the current location.

The lifting in bullet 3 is the core challenge, but we already have a very similar mechanism in place that lifts comprehensions. So what's needed is just a pair of functions that mirror the current `addComp` and `getComps`; perhaps called `addBinding` and `getBindings` instead. So `addBinding` would be called when bullet 3 above matches, with the generated name and the normalized $FORMAT call as arguments. And like `getComps`, `getBindings` would be called inside `normSuite` to insert the lifted statements in the right positions.

However, as I write this I realize that there might be unfortunate interactions between lifted comprehensions and lifted $FORMAT calls, if these expression forms contain each other. Such cases will require that the lifted statements are interleaved to match their dependency order. My conclusion is that the best way to handle the whole thing is therefore to generalize `addComp` and call it `liftStmt` instead, taking just a Stmt as its argument. In the comprehension case that statement would be what `mkCompFun` produces, and in the $FORMAT case it would be the assignment of a $FORMAT call to the generated variable. That way all lifted statements will be accumulated in their dependency order, so when a corresponding `getLiftedStmts` is invoked in `normSuite`, the found statement list is already processed and ready to be inserted right before the returned normalized suite.

A bit tricky, but fully doable. Maybe I should put this on my list of things?

plajjan:
@nordlander I think you said something along the lines of "it's just a single line to add in Normalizer" (I'm going to frame that and put it on the wall, next to the finished code ;)), so I felt compelled to avoid larger changes to norm norm', which is why I opted to create somewhat larger extractBla helper functions.

Hmm, but yeah, perhaps you are right that we need a flag. While I have developed this for quite complex cases, including comprehensions, I think I can glimpse how this breaks down, although I'm struggling to create a failing case right now.

Just to give an idea of how this translates in general and perhaps in particular for comprehensions, behold:
```python
a = "aaa>{', '.join(["{a if a != 'z' else a:<10}" for a in ['a', 'b', 'c']])}<bbb"
```

which works and turns into this, which I think looks quite reasonable
```c
B_Eq barQ_W_17;
B_Iterable barQ_W_11;
B_list barQ_N_compfun () {
    B_Sequence N_2w = ((B_Sequence)B_SequenceD_listG_new());
    B_list N_3res = B_mk_list(0);
    B_Iterator N_4iter = ((B_Iterator (*) (B_Iterable, B_list))barQ_W_11->$class->__iter__)(barQ_W_11, B_mk_list(3, to$str("a") , to$str("b") , to$str("c")));
    if ($PUSH()) {
        while ((B_True)->val) {
            B_str N_5val = ((B_str (*) (B_Iterator))N_4iter->$class->__next__)(N_4iter);
            B_str a = N_5val;
            B_str N_6fmt = $FORMAT("%-10s", ((B_str)B_strG_new(((B_value)((((B_bool (*) (B_Eq, B_str, B_str))barQ_W_17->$class->__ne__)(barQ_W_17, a, to$str("z"))->val) ? a : a))))->str);
            ((B_NoneType (*) (B_Sequence, B_list, B_str))N_2w->$class->append)(N_2w, N_3res, N_6fmt);
        }
        $DROP();
    }
    else {
        B_BaseException N_7x = $POP();
        if ($ISINSTANCE0(N_7x, B_StopIteration)) {
        }
        else {
            $RAISE(N_7x);
        }
    }
    return N_3res;
}
B_str barQ_N_1fmt;
B_str barQ_a;
int barQ_done$ = 0;
void barQ___init__ () {
    if (barQ_done$) return;
    barQ_done$ = 1;
    B_Eq W_17 = (B_Eq)B_OrdD_strG_witness;
    barQ_W_17 = W_17;
    B_Iterable W_11 = (B_Iterable)B_SequenceD_listG_witness->W_Collection;
    barQ_W_11 = W_11;
    B_str N_1fmt = $FORMAT("aaa>%s<bbb", ((B_str)B_strG_new(((B_value)({ B_str $tmp = to$str(", ");
                                                                         ((B_str (*) (B_str, B_Iterable, B_list))$tmp->$class->join)($tmp, barQ_W_11, barQ_N_compfun()); }))))->str);
    barQ_N_1fmt = N_1fmt;
    B_str a = barQ_N_1fmt;
    barQ_a = a;
}
```

plajjan:
In the light of the complexity required here, I opted to rewrite $FORMAT as a C function instead which achieves the same thing. Closing this in preference of #2327.

PR #2332, opened 2025-07-02
Add missing lub cases for skolemized tyvars

nordlander:


plajjan:
@nordlander fixed CI required checks (I was matching wrong names, so it was waiting for jobs to complete which were never going to be run), so this was now green and could be merged!

PR #2342, opened 2025-07-06
Allow anything after after

### LINKED ISSUE #1458: Allow calling actor & object methods in `after` statement
`after n: x()` currently only supports plain functions in the local scope, like we can define a method in the local actor and call that, but if we have an object or another actor we can't call the methods on those. Instead we have to wrap it up in a local function first. Seems a bit verbose? Can we allow `after` to accept object methods directly?

Like so:
```python
actor Foo():
    def foo():
        print("foo")

actor main(env):
    f = Foo()
    after 1.0: f.foo()
```

If `lambda` was accepted that would also make things a little more flexible, but alas, it is not...

plajjan:
`after X:` used to be constrained to just local functions but we now allow any expression so that we can call methods on objects or remote actor methods etc.

Fixes #1458

### DIFF
--- a/compiler/lib/src/Acton/Parser.hs
+++ b/compiler/lib/src/Acton/Parser.hs
@@ -1174,10 +1174,7 @@ after_stmt = addLoc $ do
                 assertDefAct l "after"
                 e <- expr
                 colon
-                e' <- addLoc $ do
-                    n <- name
-                    (ps,ks) <- parens funargs
-                    return $ S.Call NoLoc (S.Var (S.nloc n) (S.NoQ n)) ps ks
+                e' <- expr
                 return $ S.After NoLoc e e'
 
 var_stmt :: Parser S.Stmt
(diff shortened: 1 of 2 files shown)


nordlander:
Yes, the previous restriction to local procs only was just an attempt to "simplify" things for the programmer -- if `self` is always the implied target of timed messages there will be no non-determinism to handle if we'd like to add the possibility to cancel them, for example. But allowing any message to be timed is arguably a much more general feature (like it was in Timber, btw).

However, we should then also take the opportunity to optimize the timed message handling at run-time. As it currently stand, it's the _sender_ that's woken up when the timer triggers, just to place the message in the message queue of the receiver. Ideally `after` should result in a fully evaluated `Msg` that's just placed in the timer queue instead of a particular in-box because of the baseline parameter. Requires some fiddling with the type-checker, the deactorizer and the rts; I assign myself to do that (issue #2346).

PR #2365, opened 2025-07-17
Fix actor alias lookup

### LINKED ISSUE #2353: Why is it not possible to create actors in `__builtin__`?
I tried adding actors to the `__builtin__` module which does compile in itself but I am unable to use them. For example, in `__builtin__.act`:

```python
actor Foo():
    a = 1
```

and then in a test program:
```python
actor main(env):
    f = Foo():
    env.exit(0)
```

and I get:
```
[error Compilation error]: Name FooG_newact is not in scope
     ╭──▶ helloworld.act@1:1-1:2
     │
   1 │ actor main(env):
     • ┬
     • ╰╸ Name FooG_newact is not in scope
─────╯
```

@nordlander I suppose we do some magic for the builtin modules, I presume for Env since we bootstrap it with special C code and might not want codegen to build conflicting functions for it!? or what is going on here? why isn't this working?

plajjan:
The deactorizer transforms actor constructors like `Actor()` into calls
to generated functions like `ActorG_newact()`. This transformation was
broken for unqualified actor names that were aliases to qualified names.

Actors from imported modules are added to the environment as aliases:
- `import worker` → `worker.Worker()` works (qualified name)
- `from worker import Worker` → `Worker` becomes `NoQ Worker` with
  `NAlias (worker.Worker)`
- `__builtin__` is auto-imported → `StringDecoder` becomes
  `NoQ StringDecoder` with `NAlias (__builtin__.StringDecoder)` in the
  names environment

The old deactorizer handled unqualified names (`NoQ n`) but didn't
resolve aliases:

    deact env (Var l (NoQ n))
      | n `elem` sampled env = return $ Var l (NoQ n)
      | n `elem` locals env  = return $ Dot l (Var l (NoQ selfKW)) n'

When `Worker()` or `StringDecoder()` (parsed as `NoQ` names) weren't in
`sampled` or `locals`, they fell through to the general case which
called `newactQName (NoQ name)`.  This generated `WorkerG_newact` and
`StringDecoderG_newact` instead of the correct `worker_WorkerG_newact`
and `__builtin___StringDecoderG_newact`.

The fix adds alias resolution to the `NoQ` case:

    deact env (Var l n@(NoQ n'))
      | isActor env n = case lookup (qnName n) (names env) of
                           Just (NAlias qn) -> return $ Var l $ newactQName qn
                           _ -> return $ Var l $ newactQName n

Now unqualified actor names look up their aliases and generate the
correct module-prefixed constructor functions.

This affected builtin actors and explicit aliased imports like `from
worker import Worker`, but not qualified actors like `worker.Worker()`
which didn't need alias resolution.

I added a test case in the first which shows the generated code before the fix and then the second commit that actually fixes the alias resolution in the deactorizer also includes the fix to the test output so we can see how it is now doing the right thing.

Fixes #2353

### DIFF
--- a/compiler/lib/src/Acton/Deactorizer.hs
+++ b/compiler/lib/src/Acton/Deactorizer.hs
@@ -196,6 +196,11 @@ newactQName (QName m n)             = QName m (newactName n)
 newactQName (NoQ n)                 = NoQ (newactName n)
 newactQName (GName m n)             = GName m (newactName n)
 
+qnName :: QName -> Name
+qnName (NoQ n) = n
+qnName (QName _ n) = n
+qnName (GName _ n) = n
+
 addSelfPar p                        = PosPar selfKW (Just tSelf) Nothing p
 
 selfRef n                           = Dot l0 (Var l0 (NoQ selfKW)) n
@@ -227,7 +232,7 @@ instance Deact Expr where
       | n `elem` locals env         = return $ Dot l (Var l (NoQ selfKW)) n'
       where n'                      = if n `elem` wrapped env then localName n else n
     deact env (Var l n)
-      | isActor env n               = return $ Var l $ newactQName n
+      | isActor env n               = return $ Var l $ newactQName (unalias env n)
       | otherwise                   = return $ Var l n
     deact env (Async l e)           = Async l <$> deact env e
     deact env (Await l e)           = do e' <- deact env e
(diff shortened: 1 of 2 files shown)


nordlander:
This analysis is entirely correct and the error well spotted! I'd like to propose an even simpler solution, though: just change the deactorizer line
```
  | isActor env n               = return $ Var l $ newactQName n
```
into
```
  | isActor env n               = return $ Var l $ newactQName (unalias env n)
```
This will be the only change needed, and has the benefit of keeping all unaliasing in one place.

plajjan:
@nordlander ok, that is simpler indeed - thanks! :)

Issue #2061, opened 2025-01-08
Codegen for Times[ZZ] does not output zQ_TimesD_ZZG_new function


mzagozen:
### Acton Version

0.24.1.20250108.8.28.39

### Steps to Reproduce and Observed Behavior

I wanted to implement the `Times[ZZ]` protocol to allow multiplication of ZZ.

```acton
import testing

class ZZ(value):
    re: float
    im: float

    def __init__(self, re, im):
        self.re = re
        self.im = im

    def __str__(self):
        if self.re == 0 and self.im == 0:
            return "0"
        elif self.re == 0:
            return "%fj" % self.im
        elif self.im == 0:
            return str(self.re)
        else:
            return "%f + %fj" % (self.re, self.im)

    def __repr__(self):
        return "ZZ(%f, %f)" % (self.re, self.im)

extension ZZ(Eq):
    def __eq__(self, other):
        return self.re == other.re and self.im == other.im

extension ZZ(Plus):
    def __add__(self, other):
        return ZZ(self.re + other.re, self.im + other.im)

    @staticmethod
    def __zero__():
        return ZZ(0, 0)

extension ZZ(Times[ZZ]):
    def __mul__(self, other: ZZ) -> ZZ:
        return ZZ(self.re * other.re - self.im * other.im,
                  self.re * other.im + self.im * other.re)

def _testZZ():
    c0 = ZZ(0, 0)
    c1 = c0 + ZZ(1, 0)
    c11 = c0 + ZZ(1, 1)    
    testing.assertEqual(c0, ZZ(0, 0))
    testing.assertEqual(c0 + c0, ZZ(0, 0))
    testing.assertEqual(c1, ZZ(1, 0))
    testing.assertEqual(c11, ZZ(1, 1))
    testing.assertEqual(c1 * c1, ZZ(1, 0))

```

It looks like the generated C code is missing the zQ_TimesD_ZZG_new function?

```
❯ acton test
Building project tests...
Failed to build project tests
actonc exited with code 1 / 0
stderr: Building project in /home/mzagozen/aoc/2024
  Compiling z.act for development
   Finished compilation in   0.097 s
  Compiling tf2.act for development
   Already up to date, in    0.000 s
  Compiling tf.act for development
   Already up to date, in    0.000 s
  Compiling d02.act for development
   Already up to date, in    0.000 s
  Compiling d01.act for development
   Already up to date, in    0.000 s
  Compiling aoc.act for development
   Already up to date, in    0.000 s
[...]

Issue #1126, opened 2022-12-13
Compilation error for Hashable extension


ddjoh:
### Acton Version

0.14.2.20221201.12.18.21

### Steps to Reproduce

```
class Key:
    @property
    x: int

    def __init__(self, x):
        self.x = x

extension Key(Eq):
    def __eq__(a, b) -> bool:
        return a.x == b.x

extension Key(Hashable):
    def __hash__(self):
        return hash(self.x)

actor main(env):
    var dct: dict[Key, int] = {}
```

### Expected Behavior

Success!

### Actual Behavior

```
Building project in /home/ddjoh/workspaces/act-test-1
  Compiling mod1.act for development
   Finished compilation in   0.111 s
Building executable out/dev/bin/mod1
ERROR: internal compiler error: compilation of generated C code of the root actor failed
NOTE: this is likely a bug in actonc, please report this at:
NOTE: https://github.com/actonlang/acton/issues/new?template=ice.yaml
NOTE: acton 0.14.2.20221201.12.18.21 compiled by ghc 8.10 on linux x86_64
NOTE: cc: 0.10.0
cc stdout:

cc stderr:
ld.lld: error: undefined symbol: mod1$$Hashable$Key$new
>>> referenced by mod1.c:155 (act-test-1/out/types/mod1.c:155)
>>>               mod1.o:(mod1$$main$__init__) in archive /home/ddjoh/workspaces/act-test-1/out/dev/lib/libActonProject.a
```

nordlander:
@plajjan See my comment in #2380

PR #2380, opened 2025-07-18
Fix protocol extension constructor generation

plajjan:
Protocol extensions with inherited methods weren't getting constructor
functions generated, causing C compilation errors. This is a incomplete patch that fixes the symptoms for single-inheriting extensions.

When a protocol extends another (e.g., Hashable extends Eq), extensions
implementing the child protocol were incorrectly treated as abstract:
- `extension Key(Eq)` implements `__eq__`
- `extension Key(Hashable)` implements `hash`
- But `HashableD_Key` inherits from `Eq[Key]`, so `abstractAttrs` sees the inherited `__eq__` as abstract
- `declCon` then skips constructor generation for the abstract class `HashableD_Key`

This caused undefined symbol errors:

    error: call to undeclared function 'HashableD_KeyG_new'

Earlier compiler passes already verify that extensions provide all
required methods. The issue is purely in code generation - extensions
shouldn't be treated as abstract due to inherited methods from parent
protocols.

Extensions are converted to classes by `convExtension` by the Converter.
By the time CodeGen runs, `declCon` handles all class declarations:
user-defined classes, protocol extensions (converted to classes), and
actors. Protocol extensions get `Derived` names from
`extensionName` (e.g., `Derived Hashable Key` → `HashableD_Key`), while
regular classes keep their simple names.

The fix detects protocol extensions by their `Derived` naming pattern.
Since only protocol extensions have `Derived` names when `declCon`
processes class declarations, this reliably identifies them.

This allows protocol extensions to get constructors regardless of
inherited abstract methods while preserving the abstract class check
that prevents regular user-defined abstract classes from getting
constructors.

In practice, a lot of Acton program that used a workaround, like so:
```python
class Key:
    def __init__(self, x):
        self.x = x

extension Key(Hashable):
    def __eq__(a, b) -> bool:
        return a.x == b.x

    def hash(self, hasher):
        self.x.hash(hasher)
```

can now instead be written in the more proper way:
```python
class Key:
    def __init__(self, x):
        self.x = x

extension Key(Eq):
    def __eq__(a, b) -> bool:
        return a.x == b.x

extension Key(Hashable):
    def hash(self, hasher):
        self.x.hash(hasher)
```

### DIFF
--- a/compiler/lib/src/Acton/CodeGen.hs
+++ b/compiler/lib/src/Acton/CodeGen.hs
@@ -711,7 +711,8 @@ classCast env ts x q n              = parens . (parens (gen env t) <>)
 genNew env n p                      = newcon' env n <> parens (gen env p)
 
 declCon env n q b
-  | null abstr || hasNotImpl b      = (gen env tRes <+> newcon env n <> parens (gen env pars) <+> char '{') $+$
+  | null abstr || hasNotImpl b || isProtocolExtension b
+                                    = (gen env tRes <+> newcon env n <> parens (gen env pars) <+> char '{') $+$
                                       nest 4 (gen env tObj <+> gen env tmpV <+> equals <+> acton_malloc env (gname env n) <> semi $+$
                                               gen env tmpV <> text "->" <> gen env1 classKW <+> equals <+> char '&' <> methodtable env1 n <> semi $+$
                                               initcall env1) $+$
@@ -728,6 +729,13 @@ declCon env n q b
         retobj (PosArg e p)         = PosArg (eCall (tApp (eQVar primCONSTCONT) [tObj]) [eVar tmpV, e]) p
         env1                        = ldefine ((tmpV, NVar tObj) : envOf pars) env
         abstr                       = abstractAttrs env (NoQ n)
+        -- Check if this is a protocol extension class
+        -- Protocol extensions are created by convExtension with names like:
+        -- Derived protocol_name class_name (e.g., Derived Hashable Key -> HashableD_Key)
+        -- By the time we reach declCon in CodeGen, only protocol extensions have Derived names
+        isProtocolExtension b       = case n of
+                                        Derived _ _ -> True
+                                        _           -> False
 
 acton_malloc env n                  = text "acton_malloc" <> parens (text "sizeof" <> parens (text "struct" <+> gen env n))
 
--- a/compiler/lib/test/9-codegen/protofoo.c
+++ b/compiler/lib/test/9-codegen/protofoo.c
@@ -102,13 +102,37 @@ protofooQ_Key protofooQ_KeyG_new(B_int G_1) {
     return $tmp;
 }
 struct protofooQ_KeyG_class protofooQ_KeyG_methods;
+B_NoneType protofooQ_EqD_KeyD___init__ (protofooQ_EqD_Key W_self) {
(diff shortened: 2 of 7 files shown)


plajjan:
@nordlander PTAL. Trying to identify protocol extensions in codegen where they have already been converted to classes feels like "too late" but I found Derived name and I think that protocol extensions is the only class-thing that goes through declCon that has a Derived name. Other Derived names for witnesses or whatnot never end up here.. I think!? So it's a little hacky but should be robust and does match the correct thing.

And code works, both existing code with workaround should work as well as the correct way of doing things, so we can do a smooth transition for our ecosystem.

Is there a simpler, more robust and elegant way? Or perhaps you have a better broader solution in mind?

nordlander:
No, this isn't a solution to the problem, merely a removal of the immediate symptom (which just happens to make the simple case of issue #2061 compile).

Protocols and extensions are indeed converted to classes after type-checking, with the idea that extensions are concrete classes of witnesses while protocols form abstract base classes to those witnesses (although they can contain as many inheritable method implementations as one wishes). But unlike regular classes the protocols support multiple inheritance and Python's Method Resolution Order (MRO), which calls for some rather intricate code transformations. For example, the numeric protocol hierarchy of `__builtin__.act`(graphically depicted on line 552) has two branches that go beyond single inheritance (`Minus` and `Logical`), which leads to the generation of extra witness parameters and several extra protocol classes after type-checking (see classes `MinusD_Number`, `MinusD_Real`, `MinusD_Rational`, `MinusD_Integral` and `LogicalD_Integral` in the `--types` output) together with extra classes for every extension that implements these protocols (`MinusD_IntegralD_int` and `LogicalD_IntegralD_int` in addition to `IntegralD_int`, for example).

These transformations take place in module `Converter`, with the MRO algorithm in module `Env`. The code is tricky but I believe it to be correct -- with the exception of the big fatal flaw illuminated already in issue #1126. The core problem is that the `Converter` algorithm fails to honor that methods required by an extension can in general be found in multiple _previous_ extension classes as well, a problem that has no counterpart in Python since it doesn't support the extension of a class as an afterthought. Still, the type-checker does take previous extensions into account when it checks for extension completeness, so the problem only shows up as internal incompleteness in the classes generated by `Converter`. But they are really incomplete, their method tables will have some empty slots that should have been filled with function pointers from some other method table.

Forcing the code-generator to generate constructor functions for such incomplete classes is therefore not a proper solution, even though it might work in cases where the compiler hasn't found the need to look up an inherited method in one of the incomplete witness. A proper solution must involve modification of the algorithm in `Converter`, and the obstacle there is that it's quite complicated as it is, with an annoyingly large overlap between `convProtocol` and `convExtension`. The good thing, though, is that there's no question about how this conversion should work in principle -- I only wish there was a more straightforward way to express it!

Maybe it's time for me to prioritize this issue (too)?

plajjan:
@nordlander I suspected as much but wrote the solution anyway. I wanted to understand the problem - and then found what looked like a simple hack to fix it ;) I didn't know about multiple inheritance for extensions and thus didn't consider that angle. While implementing this I also realized I still don't understand which implementation is finally picked for the witness - like if Hashable extension for a class support the inherited Eq and there is also a Eq extension, how we do choose one - and just finished the solution rather than figuring out how that works - but I'll look into that myself!

You can't prioritize all issues we have ;) but you and I definitely need to sit down and make a priority list so we know what gets done when. I did start by doing a very coarse sorting of almost all issues on this repo in the "Acton project" - it's ready for us to talk about.

PR #2391, opened 2025-07-29
Revamp class attribute init check

### LINKED ISSUE #1116: Compilation error for class declaration - bad class attr init check
### Acton Version

acton 0.14.0.20221120.14.27.9

### Steps to Reproduce

    class rational(value):
        def __init__(self, num : int, denom : int):
            if denom == 0:
                raise ValueError("Denominator in rational is zero")
            else:
                if denom > 0:
                    self.num = num
                    self.denom = denom
                else:
                    self.num = -num
                    self.denom = -denom

### Expected Behavior

Successful compilation

### Actual Behavior
```
rational> actonc rational.act 
Building file rational.act
  Compiling rational.act for release

ERROR: Error when compiling rational module: Type error
 10:22-24
   |
10 |                self.num = -num
   |                     ^^^
Attribute not found: num

rational> 
```

It seems that both if statements are unacceptable; only unconditional assignments to self.num and self.denom pass the compilation.

### LINKED ISSUE #1425: Missing checks for initialization of base class properties
### Acton Version

0.16.0.20230807.13.20.3

### Steps to Reproduce

Compile and run

```
class Foo:
    @property
    x: str

    def __init__(self, x):
        self.x = x

    def __str__(self):
        return self.x

class Bar(Foo):
    def __init__(self):
        pass

actor main(env):
    b = Bar()
    print(b)
    await async env.exit(0)
```

### Expected Behavior

Compiler error due to uninitialized properties inherited from class Foo.

### Actual Behavior

```
ERROR: segmentation fault
NOTE: this is likely a bug in acton, please report this at:
NOTE: https://github.com/actonlang/acton/issues/new
NOTE: include the backtrace printed below between -- 8< -- lines

-- 8< --------- BACKTRACE --------------------
[New LWP 575478]
[New LWP 575479]
[New LWP 575480]
...
```

### LINKED ISSUE #2158: Detect uninitialized class attributes
### Acton Version

0.24.1.20250205.13.28.10

Example:
```
class Test():
    results: dict[str,str]
    def __init__(self):
        self.result = {}
```
This will currently result in a class with _two_ attributes: `results` and the presumed typo `result`. And only the typo will get initialized!!!

### Expected Behavior

The missing initialization of `results` being detected by the compiler (so that the typo can be found and fixed).

plajjan:
This is a rewrite of the class attribute initialization check. In the
__init__ method we have a constructor part which must unconditionally
initialize all attributes, so that we get a valid object. In the
constructor, references to `self` must not escape until after the
construction of the new object is complete (i.e. all attributes have
been initialized) nor can we call self.methods() since those methods
typically assume the self object is valid.

The inferrence of properties now scans through the entire __init__ method
looking for self.x assignments. Previously it would stop at some
particular statements, but that limited our view, so that's been removed
and we are now scanning the entirety of __init__.

We dynamically determine where the constructor ends by scanning through
the __init__ method up to the point where a reference to `self` is made
- at which point the constructor must be complete since we cannot pass a
reference to `self` before the own object has been properly constructed.
We make an exception for call self methods with NotImplemented bodies -
since those are implemented in C, we reckon the user also knows not to
touch the wrong self attributes. Loops are skipped past, since we cannot
statically determine if assignments in loops will happen or not,
however, we do scan to ensure that no references to self escape from
within the loop.

The analysis is more complete around handling of try / except as well as
if/else branches. Previously, we would only support:
```
def __init__(self, foo):
  self.x = 1 if foo is not None else 3
```
Now we support:
```
def __init__(self, foo):
  if foo is not None:
      self.x = 1
  else:
      self.x = 3
```
Similarly, we can mix in try / except

```
def __init__(self, num : int, denom : int):
    if denom == 0:
        raise ValueError("Denominator in rational is zero")
    else:
        if denom > 0:
            self.num = num
            self.denom = denom
        else:
            self.num = -num
            self.denom = -denom
```
Added a page to acton.guide explaining most of this as well.

The error printing has also been improved, so rather than a simple "attribute x is not initialized" we now have:

```
[error Type error]: Attribute 'y' is not initialized in TestClass.__init__
     +--> uninit_grandparent2.act@14:5-17:1
     |
   6 |    class BaseB(BaseA):
     :          ^------------
     :          `- Attribute inherited from BaseB
   7 |        y: int
     :        ^
     :        `- Attribute 'y' is defined here
     :
  13 |    class TestClass(BaseB):
     :    ^----------------------
     :    `- In class TestClass
  14 | +>     def __init__(self):
  15 | |          BaseA.__init__(self)
  16 | |          # Should get a failure since y is not initialized
  17 | |>
     : |
     : `- Attribute 'y' is not initialized in __init__
-----+
```

Fixes #1116.

Fixes #1425.

Fixes #2158.

### DIFF
--- a/docs/acton-guide/src/classes/initialization.md
+++ b/docs/acton-guide/src/classes/initialization.md
@@ -1,36 +1,55 @@
 # Class Initialization
 
-All class attributes must be initialized in the `__init__` method. This ensures that objects are always in a valid state after construction.
+All class attributes must be initialized in the `__init__` method to ensure objects are always in a valid state after construction.
 
 We can think of `__init__` as having two parts:
-1. **Constructor part**: Must fully initialize the object through unconditional simple logic
-2. **Post-constructor part**: Can perform other logic once the object is fully initialized, calling methods on `self` or passing references to `self`
+1. **Constructor part**: Must fully initialize the object. Ends when `self` escapes (is passed externally)
+2. **Post-constructor part**: Can perform other logic once the object is fully initialized
+
+## The Constructor Part
+
+The constructor part continues until `self` escapes - that is, until we pass a reference to `self` externally. At that point, the object must be fully initialized. Examples of `self` escaping:
+- Passing `self` to a function: `register(self)`
+- Calling a method on `self`: `self.validate()`
+- Passing a method reference: `callback(self.on_event)`
+- Returning `self` or raising with `self`
+
+During the constructor part, you can access already-initialized attributes:
+
+```python
+class Config(object):
+    def __init__(self, base_value: int):
+        self.base = base_value
+        self.doubled = self.base * 2        # OK: self.base is initialized
+        self.quadrupled = self.doubled * 2  # OK: self.doubled is initialized
+```
+
+Note that accessing `self.attribute` where the attribute is uninitialized will stop the constructor part, as it indicates an error in initialization order.
+
+## The Post-Constructor Part
+
+Once `self` escapes, we enter the post-constructor part where all attributes must already be initialized:
 
 ```python
 class BankAccount(object):
-    owner: str
-    balance: float
-    transaction_log: list[str]
-
     def __init__(self, owner: str, initial_deposit: float):
-        # Constructor part: Initialize all attributes first
+        # Constructor part: Initialize all attributes
         self.owner = owner
         self.balance = initial_deposit
         self.transaction_log = []
 
-        # Post-constructor part: Now we can use methods and pass self
-        self.log_transaction("Account opened")
-        register_account(self)  # OK to pass self - object is fully initialized
+        # Post-constructor: self escapes here, all attributes must be initialized
+        self.log_transaction("Account opened")  # Calls method on self
+        register_account(self)                  # Passes self to function
         if initial_deposit > 10000:
             flag_for_review(self)
-
-    def log_transaction(self, msg: str):
-        self.transaction_log.append(msg)
 ```
 
-## Conditional branches and `raise`
+## Control Flow in the Constructor Part
 
-if / elif / else are valid as long as all branches initialize all attributes. `raise` will not create the new object and is thus seen as en exception to the unconditional initialization check.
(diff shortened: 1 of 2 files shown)


plajjan:
@nordlander PTAL! I'm not entirely done yet but I think the main areas in Types.hs are coming in to shape so it would be useful if you could look at those!

plajjan:
It's nice (or scray? 😝) to see that it is uncovering latent bugs in almost all of our projects :P

plajjan:
@nordlander I improved the error message. I noticed while using this on other Acton repos, like acton-yang, that the error message just pointed to where the attribute is originally defined, so for the Node class in acton-yang, that's not very helpful since it's the __init__ message of a derived class that is missing initializing it. New error message shows it all:

```
[error Type error]: Attribute 'y' is not initialized in TestClass.__init__
     +--> uninit_grandparent2.act@14:5-17:1
     |
   6 |    class BaseB(BaseA):
     :          ^------------
     :          `- Attribute inherited from BaseB
   7 |        y: int
     :        ^ 
     :        `- Attribute 'y' is defined here
     :
  13 |    class TestClass(BaseB):
     :    ^----------------------
     :    `- In class TestClass
  14 | +>     def __init__(self):
  15 | |          BaseA.__init__(self)
  16 | |          # Should get a failure since y is not initialized
  17 | |> 
     : | 
     : `- Attribute 'y' is not initialized in __init__
-----+
```

I placed this in a second commit and I think it might make sense to review the commits individually, the first focusing on the core functionality and the second on adding the improved error message.

plajjan:
We don't allow self.method-calls - but I realized, while reading our acton-gnmi module, I think we can make an exception when the method is NotImplemented since then the programmer is really taking control of things. @nordlander and I discussed it today as well and agree that we should allow it.

there are some warnings around overlapping pattern matches from ghc - I will look.

And scan for self in function defs.

plajjan:
And implement the same branch early-exit check we have for if/ elif etc for try / except!

Issue #2411, opened 2025-08-08
Remove numpy


plajjan:
We've been a little back and forth over this for years now. Numpy is cool but our implementation of it is quite messy, in the sense that it sprawls, we have special support in the compiler for numpy. numpy is the only module still using the stub-compilation mode (which we otherwise could remove).

We are not using numpy anywhere other than in some example code (which no one looks at). We constantly do more special work just for numpy.

I noticed @sydow removed it in a recent branch for revamping use of int in builtins. I guess it's the last straw and I feel it would be relief to remove it. It's still there in the git log and we could plausible bring back a revamped version if and when we need it. The longer we wait, likely the better since if we have native support for unboxed types etc it could make implementation of a numpy module fit better and be less hacky.

- [ ] remove numpy module
- [ ] remove tests
- [ ] remove numpy related hacks in compiler
- [ ] remove stub compilation mode

PR #2413, opened 2025-08-09
Remove numpy

plajjan:
Part of #2411


plajjan:
Ok, let's go! bye bye numpy

PR #2420, opened 2025-08-12
Improve XML parse error reporting

mzagozen:
Use xmlGetLastError() to get the last error from libxml2. Strip trailing whitespace from the message if present.

We already get more verbose output from libxml2 on stderr, but this allows us to pass at least some hints about what went wrong via an exception.

### DIFF
--- a/base/src/xml.ext.c
+++ b/base/src/xml.ext.c
@@ -20,7 +20,7 @@ xmlQ_Node $NodePtr2Node(xmlNodePtr node) {
     }
     if (node->type != XML_ELEMENT_NODE) {
         char *errmsg = NULL;
-        $RAISE(((B_BaseException)B_RuntimeErrorG_new($FORMAT("Unexpected nodetype %d, content is %s", node->type, node->content))));
+        RAISE(xmlQ_XmlParseError, $FORMAT("Unexpected nodetype %d, content is %s", node->type, node->content), NULL, NULL);
     }
 
     B_list nsdefs = B_listG_new(NULL, NULL);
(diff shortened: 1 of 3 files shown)


base/src/xml.ext.c
```diff
@@ -71,8 +71,33 @@ xmlQ_Node $NodePtr2Node(xmlNodePtr node) {
 xmlQ_Node xmlQ_decode(B_str data) {
     xmlDocPtr doc = xmlReadMemory((char *)data->str, data->nbytes, NULL, NULL, 0);
     if (!doc) {
-        // xmlErrorPtr err = xmlGetLastError();
-        $RAISE(((B_BaseException)B_RuntimeErrorG_new(to$str("xml parse error"))));
+        xmlErrorPtr err = xmlGetLastError();
+        // Note: libxml2 also prints additional context to stderr including the
```
plajjan, line 75: watawat, libxml2 prints directly to stderr? surely we want to avoid that!?

mzagozen: Yes, like this (last line is from our Acton program):

```
Entity: line 1: parser error : Premature end of data in tag asd line 1
<asd>
     ^
{'exc':'RuntimeError: XML parse error: Premature end of data in tag asd line 1'}
```

I now added a flag to suppress this output to stderr.

base/src/xml.ext.c
```diff
@@ -69,10 +69,33 @@ xmlQ_Node $NodePtr2Node(xmlNodePtr node) {
...
+                msg_clean[len] = '\0';
+                errmsg = $FORMAT("XML parse error: %s", msg_clean);
+            } else {
+                // Use original message as-is
+                errmsg = $FORMAT("XML parse error: %s", err->message);
+            }
+        } else {
+            errmsg = to$str("XML parse error");
+        }
+        $RAISE(((B_BaseException)B_RuntimeErrorG_new(errmsg)));
```
plajjan, line 98: It's a little weird to throw a RuntimeError, no? I see that it was already here before your change.. but maybe we should take this opportunity to change it?

plajjan (approved):
I trust that this works. Consider taking the opportunity to change the error type to something better and perhaps add a test case or two (that would go in tests/stdlib_tests/src/test_xml.xml).

Feel free to merge. If you want to merge as-is, that's OK. If you want to add the test and / or change exception type, that's OK too.

PR #2422, opened 2025-08-13
Fix XML encoding, decoding for special chars, CDATA

### LINKED ISSUE #2419: Fix XML decoding / encoding with entities and CDATA
### Acton Version

0.26.0.20250812.8.59.41

# Decoding XML

## Special character escapes

Special characters in XML are `&`, `<`, `>`, `"`, `'`. There are [five valid entities for encoding special characters in XML](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#Standard_public_entity_sets_for_characters). We can always transform the entities to literal characters when decoding node content or attributes, but leave them untouched in CDATA.

```xml
<root tag="&quot;">
  <escape>&lt;&gt;</escape>
</root>
```

## CDATA

CDATA is parsed as a separate node type in libxml2 and we currently do not handle it (throw a runtime error): 

```xml
<root>
  <comment><![CDATA[hello&lt;]]></comment>
</root>
```

The text inside of CDATA should be appended to the rest of the text in the node as-is, without the CDATA wrapper.

# Encoding XML

We must use the same valid entities for encoding special characters in XML as needed. For example, if the attribute is quoted with a single quote, then double quote doesn't need an escape, and vice versa. IMHO we should just stick with entities for encoding, not use CDATA.

### Expected Behavior

To be able to parse and produce valid XML. But do we extend the existing artisanal encoder, or offload to libxml2?

mzagozen:
Add XML special char encoding that only escapes (replaces with entity)
when strictly necessary per XML spec:
- Text content: escapes & and < only
- Attribute values: escapes & < and " only (since we quote with ")
- The > character is never escaped as it's not required by XML spec

When we encounter a CDATA tag in the XML, we collect and combine the
contents of the tag with consecutive text and other CDATA nodes
verbatim.

Fixes #2419

### DIFF
--- a/test/stdlib_tests/src/test_xml.act
+++ b/test/stdlib_tests/src/test_xml.act
@@ -19,7 +19,8 @@ def _test_xml_roundtrip():
         """<a>\n    <b1>hej</b1>\n    <b2>test</b2>\n</a>""",
         """<a xmlns="http://foo"><b>2</b></a>""",
         """<a xmlns:ns="http://foo"><ns:b>2</ns:b></a>""",
-        """<data><l>čaw</l><baw></baw></data>"""
+        """<data><l>čaw</l><baw></baw></data>""",
+        """<a>100 ㏀</a>""",
     ]
     for s in test_xml:
         #print(s)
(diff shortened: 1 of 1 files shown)


base/src/xml.ext.c
```diff
@@ -12,6 +12,92 @@
     (nm)->str = acton_malloc_atomic((nm)->nbytes + 1);       \
     (nm)->str[(nm)->nbytes] = 0
 
+// Helper function to count extra bytes needed for XML escaping
+// Returns the number of extra bytes needed to replace the special character
+// with encoded entity (not total bytes)
+static int count_xml_escape_extra(unsigned char *str, int nbytes, int escape_quotes) {
+    int extra = 0;
+    for (int i = 0; i < nbytes; i++) {
+        switch (str[i]) {
```
plajjan, line 21: I think this is wrong. You are iterating over this byte-by-byte. XML takes strings as input which allow utf-8 chars, so you could be inspecting the second byte (or third or four) in a utf-8 char and think that it's a special char when it's really not.

The comparison is basically correct, I just think you need to iterate over (utf-8) chars. Look at `utf8proc_iterate` - it's used in `str.c`. You could also consider writing such escape functions in Acton instead of in C.

base/src/xml.ext.c
```diff
@@ -169,22 +259,28 @@ static B_str xmlQ_encode_attrs(B_list attrs) {
...
         B_tuple attr = attrs->data[i];
-        res_bytes += ((B_str)attr->components[0])->nbytes + ((B_str)attr->components[1])->nbytes + 4; // 4 = ' ','=,'"', '"'
-        res_chars += ((B_str)attr->components[0])->nchars + ((B_str)attr->components[1])->nchars + 4;
+        B_str key = (B_str)attr->components[0];
+        B_str value = (B_str)attr->components[1];
+
+        // Count extra bytes needed for escaping
+        int extra_bytes = count_xml_escape_extra(value->str, value->nbytes, 1);
+
+        res_bytes += key->nbytes + value->nbytes + extra_bytes + 4; // 4 = ' ','=,'"', '"'
```
plajjan, line 268: What is the comment at the end of the line trying to explain, 4 = watawhat?

plajjan (changes requested):

PR #2457, opened 2025-09-24
Handle malformed HTTP request gracefully

mzagozen:
The HTTP Server actor decodes parts of the HTTP request to str. If the
part is not a valid UTF-8 string, it is not a valid HTTP request, so we
stop processing.


mzagozen:
*perf-vs-main* fails as expected with `ValueError: to$str: Unicode decode error`

PR #2460, opened 2025-09-26
Fix unboxing of u64

### LINKED ISSUE #2452: Unboxing should include u64
From Boxing.hs:
```hs
integralTypes                      = [tInt, tI64, tI32, tI16, tI64, tU32, tU16]
```
so `tI64` is in there twice and `tU64` is missing!

plajjan:
Typo, this should be U64, not I64!

Fixes #2452


sydow:
This is handled in #2465.

PR #2466, opened 2025-09-30
Add test of effectful loop condition

nordlander:


plajjan:
@nordlander isn't it better to just have this as a golden test on codegen?

Overall, every test we add that's a full binary adds 1-2 seconds to our testing time whereas the tests we have in ActonSpec.hs run in milliseconds - 3 orders faster!! isn't codegen golden output better here?

PR #2490, opened 2025-10-24
Make Acton CLI responsible for build.zig, build.zig.zon

mzagozen:
Acton CLI now replaces the SYSPATH placeholder in build.zig.zon (from dist/builder/build.zig.zon) with the correct relative path to the Acton distribution. Previously this was only done in actonc, which made way for a race condition when building projects with a diamond-shaped dependency tree.

actonc now *does not* touch build.zig and build.zig.zon when started as a sub-compiler from Acton CLI (--sub flag). When started as a standalone compiler, for example when building Acton CLI itself, it still ensures that both files exist.

### DIFF
--- a/cli/src/acton.act
+++ b/cli/src/acton.act
@@ -110,6 +110,7 @@ actor CacheSizeChecker(cap: file.FileCap):
     check_caches()
 
 def write_buildzig(file_cap, build_config: BuildConfig):
+    """Unconditionally write build.zig and build.zig.zon files for the project"""
     fs = file.FS(file_cap)
 
     def get_build_zig_templates():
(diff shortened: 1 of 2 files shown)


cli/src/acton.act
```diff
@@ -120,6 +121,11 @@ def write_buildzig(file_cap, build_config: BuildConfig):
         build_zig_zon_template = file.ReadFile(
             file.ReadFileCap(file_cap),
             file.join_path([base_path(file_cap), "builder", "build.zig.zon"])).read().decode()
+
+        # Replace SYSPATH with the relative path to the Acton system installation (Zig requires relative paths)
```
plajjan: We have the gen_buildzigzon function that does a bit of templating, so I'd rather keep it all in there rather than spread it out like this.

PR #2492, opened 2025-10-24
Extend static B_int allocation to -128

mzagozen:
Extended the B_int_strs cache to include pre-allocated B_int instances for negative integers from -1 to -128 (8-bit signed). This avoids dynamic allocation, and by extension the GC, with the aim of improving performance.


mzagozen:
Perhaps the range -1 .. -128 is too broad as the most common negative number returned from C is -1 (like from str.c, net.c). @sydow what do you think?

sydow:
I see no problems and Marko's PR looks perfect. I would recommend merging it as it is.

plajjan:
> Compared to the static B_bool change, this is statistically insignificant. Marking as draft, should reduce the range to -5.

but is this insignificant when you do `.contains()` or with `.index()`??

PR #2564, opened 2025-12-23
Solver next

nordlander:
Fixes a string of mostly independent bugs. But one bug remains that stops telemetrify-core from compiling. The fix (remove a logically redundant call to simplify) is trivial but has the unfortunate side-effect of destroying type-checking performance -- at least on schema.act in acton-yang. So a proper fix will have to wait until the new solver is in place.

### DIFF
--- a/compiler/lib/test/3-types/deact.output
+++ b/compiler/lib/test/3-types/deact.output
@@ -1,43 +1,43 @@
 
-W_311: __builtin__.Number[__builtin__.int] = __builtin__.IntegralD_int()
+W_475: __builtin__.Number[__builtin__.int] = __builtin__.IntegralD_int()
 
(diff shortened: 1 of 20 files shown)


compiler/actonc/test/typeerrors/ex22.golden
```diff
@@ -5,7 +5,7 @@ Building file test/typeerrors/ex22.act using temporary scratch directory
      |
    1 |    f : [A(Plus)] => (A) -> A
      :    ^ 
-     :    `- too few argument(s) in call to f
+     :    `- keyword argument(s) 'y' is missing in call to f
```
plajjan: 10 points to Gryffindor!

PR #2567, opened 2025-12-29
Hide private names from interfaces

### LINKED ISSUE #1537: __private stuff imported?
Importing a module includes private stuff. I'm not sure that it should?

- `src/a.act`
```python
__foo = {}
```
- `src/test_a.act`
```python
from a import *

__foo = {}

actor main(env):
    print(__foo)
```

So since we do `from a import *` we get `a.__foo` available to us as `__foo` but we already have a `__foo`. This trickles down all the way to C where we hit some error:
```
zig build-lib ActonProject Debug x86_64-linux-gnu.2.27: error: error(compilation): clang failed with stderr: /home/kll/terastream/acton/test/compiler/collide/out/types/test_a.c:79:25: error: use of undeclared identifier 'test_aQ___foo'
```

It's clearly a bug somehow but I don't know which way.
- should `from a import *` actually import __private stuff? I expected it not to
- if it should import private stuff, then I guess the local thing should shadow the import? (but also see #372)

@nordlander ping! :)

plajjan:
Private names should not be part of a module's public interface. They previously leaked through .ty files and could be accessed via star imports or qualified access, and they also affected the interface hash.

Filter underscore-prefixed names out of the exported interface. This makes qualified access to private names fail as expected.

Rename __test_main to test_main so test roots remain public, update the compiler references and the golden output, and add regression fixtures for both star import and qualified access.

Fixes #1537

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -309,6 +309,9 @@ test-compiler:
 	cd compiler && stack test acton
 	cd compiler && stack test actonc --ta '-p "compiler"'
 
+test-compiler-accept:
+	cd compiler && stack test acton --test-arguments "--golden-start --golden-reset"
+
 test-cross-compile:
 	cd compiler && stack test actonc --ta '-p "cross-compilation"'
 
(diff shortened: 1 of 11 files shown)


plajjan:
@nordlander PTAL. Is this the right approach?

nordlander:
Indeed it is!

PR #2654, opened 2026-02-20
Dynamically link glibc

plajjan:
We now dynamically link glibc on Linux on x86_64 while keeping other libs static, just like we already do for aarch64.


plajjan:
Superseded by #2774. Closing.

PR #2658, opened 2026-02-20
Build GHC with Zig CC and control glibc target version

plajjan:
The key change here is that the GHC/Stack build path now goes through Zig CC. Once Zig is the compiler driver in that path, we can explicitly choose which GNU libc version we target instead of inheriting whatever happens to be on the build host.

On Linux, the Zig wrapper scripts accept ACTON_ZIG_GLIBC_VERSION and derive a -target of the form <host-arch>-linux-gnu.<major.minor> from zig env, so we do not hard-code CPU architecture while still pinning a glibc baseline.

If an explicit target is already supplied, it takes precedence. macOS behavior is unchanged.

The Makefile now defaults ACTON_ZIG_GLIBC_VERSION to 2.27 on Linux and exports it, and the dev guide documents the knob. In practice this gives us deliberate control over glibc compatibility for the compiler build products.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -231,6 +231,8 @@ jobs:
       - name: "Build Acton"
         run: |
           make -j2 -C ${GITHUB_WORKSPACE} BUILD_RELEASE=${{ env.BUILD_RELEASE }} BUILD_TIME=${BUILD_TIME}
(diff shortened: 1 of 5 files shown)


plajjan:
Superseded by #2774. Closing.

PR #2703, opened 2026-03-26
Describe root actor env parameter

mzagozen:
If the parameter of the candidate root actor (defualt main) is not of type Env, then the actor is an eligible root actor.


plajjan:
Thanks. It conflicts due to my reshaping of the guide. I included this text, somewhat modified, in #2740. Closing.

PR #2704, opened 2026-03-26
Describe different kinds of local actor attributes

mzagozen:
State variable vs public constant vs private constant.


plajjan:
Thanks. It conflicts due to my reshaping of the guide. I included this text in #2740. Closing.

PR #2705, opened 2026-03-27
Prebuild one-byte ASCII strings

plajjan:
String indexing, iteration, one-character slices, chr(), and single-byte C-string conversions all make two allocations, the B_str box and the backing C char buffer. In many string processing intensive applications this is very frequently called and thus exerts a large pressure on the memory system increasing GC runtimes.

We now add a prebuilt table of one-byte ASCII characters from 1 to 127 and use it from mk_char, one-character slices, chr(), and the ASCII fast path in to$str. Empty C-string construction now also returns the existing null_str singleton. This vastly reduces memory allocation and considerably speeds up these programs. The YANG parser in acton-yang runs about twice as fast.

This matches the existing bigint pattern of reusing prebuilt values instead of constructing them repeatedly. Sharing is correct here because strings are immutable and the table excludes NUL, preserving the runtime's C-string assumptions.

### DIFF
--- a/base/builtin/str.c
+++ b/base/builtin/str.c
@@ -33,9 +33,267 @@ static struct B_str null_struct = {&B_strG_methods,0,0,&nul};
 
 static B_str null_str = &null_struct;
 
-static struct B_str space_struct = {&B_strG_methods,1,1,(unsigned char *)" "};
-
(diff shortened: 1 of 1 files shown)


plajjan:
@sydow please have a look :)

PR #2721, opened 2026-04-10
Add Result type

plajjan:


plajjan:
@nordlander does this align with what you had in mind?

I got various build errors when I tried to place this in `__builtin__.act` so I did a separate module right now - that should be fixed obviously, but we can do that separately.

nordlander:
Looks good, but I'd much rather have abstract signatures instead of NotImplemented methods in the base class. That way no-one can be tempted to inherit them.

plajjan:
Result should be a union, not a class... so let's ditch this and return once we have unions.

PR #2726, opened 2026-04-14
added forced unwrapping

### LINKED ISSUE #1859: Add ! ? to access nested optional stuff
? to return optional with early shortcut if intermediate things are None

! to throw exception

sydow:
This PR implements forced unwrapping and should complete #1859.

We take the approach to parse the postfix operator ! as yet an alternative trailer in atomic expressions, together with ? and indexing, slicing, selection and calls. The operators ! and ? are thus interchangable and expect an expression of optional type as left operand and a sequence of trailers as right operand. It is debatable how useful it is to mix ! and ? in the same sequence, but the following is valid code:
````
actor main(env):
   d: ?dict[str,list[int]] = { "a": list(range(20)), "c": list(range(30)) }

   a1 = d ?.get("a") ? [1::2] [5] # a1 has type ?int; last !/? operation decides type.
   a2 = d !.get("a") ! [1::2] [5] # a2 has type int
   a3 = d ?.get("a") ! [1::2] [5] # a3 has type int
   a4 = d !.get("a") ? [1::2] [5] # a4 has type ?int
  
   print(a1,a2,a3,a4)
   
   print(d ?.get("b") ? [1::2])  # prints None; failure in ? operation
   print(d !.get("b") ? [1::2])  # prints None; failure in ? operation
   try:
        print(d ?.get("b") ! [1::2])  # raises ValueError; failure in ! operation
        print(d !.get("b") ! [1::2])  # Never executed; would raise ValueError
   except ValueError:
        print("ValueError")
   env.exit(0)
````
Note that for a successful sequence of operations, the last ?/! operator decides the type of the result. For a failed operation (None encountered in one !/? operation), the failing operation decides the effect: ? gives a None result and ! raises a ValueError.

### DIFF
--- a/base/builtin/exceptions.c
+++ b/base/builtin/exceptions.c
@@ -149,3 +149,9 @@ struct $RETG_class $RETG_methods = {
     .__serialize__      = $RETD___serialize__,
     .__deserialize__    = $RETD___deserialize__
 };
+
+
+$WORD $raiseValueError(B_str msg) {
+    $RAISE((B_BaseException)$NEW(B_ValueError,msg));
+    return ($WORD)0;
+}
--- a/base/builtin/exceptions.h
+++ b/base/builtin/exceptions.h
@@ -175,3 +175,6 @@ $SEQ $SEQG_new();
 $BRK $BRKG_new();
 $CNT $CNTG_new();
 $RET $RETG_new(B_value);
+
+$WORD $raiseValueError(B_str str);
+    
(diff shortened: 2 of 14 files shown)


sydow:
To reviewers: The changes to LambdaLifter.hs ended up in this PR by mistake; the changes there were done in preparation for an entirely different purpose (removal of classes generation in lambda-lifting). These changes are innocent but should be removed from this PR, but I leave it to run the tests as I must do other things now.

nordlander (approved):
I support this -- go!

(Assuming the changes to LambdaLifter.hs are moved to some other branch!)

PR #2727, opened 2026-04-15
Handle stale transitive deps after import rename

mzagozen:
Add a test showing incorrect cache invalidation for renamed imports: main.act subclasses base.Base, base.act switches its provider import from oldpkg.ttt to newpkg.ttt, and the rebuild runs without cleaning out/. The cached .ty for main could still reference oldpkg.ttt and fail with a missing type interface error even after base had been rebuilt.

@plajjan the branch also includes a fix that avoids the same problem in sorespo, but I am not 100% how well it fits in with the rest of the ideas in the compiler ... Please check, and if do you come up with a different approach, test by compiling sorespo *main* (`c693ea740a04e212db4cdf8187f6b55bf3732fb9`) and *stratoweave* (`11315b242078662715145cedf788742cd6c73183`) without clearing the local cache.

### DIFF
--- a/compiler/acton/test_incremental.hs
+++ b/compiler/acton/test_incremental.hs
@@ -2001,6 +2001,67 @@ p43_equal_act_ty_mtime_hashes_source =
     assertBool ("did not expect stale import diagnostic\n" ++ T.unpack out)
       (not (T.isInfixOf "Type interface file not found or unreadable for a" out))
 
+p44_provider_import_rename_reruns_dependent_front :: TestTree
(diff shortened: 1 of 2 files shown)


plajjan:
LGTM.

We treated many errors as hard errors initially because it made it easier for me to write this stuff, so like "invalid cache" is an error but it is of course recoverable. I just missed updating from "throw hard error" to "just forget the cache and rebuild" in a bunch of places. I should audit - might be a few more. Anyway, this is the right fix for these, so merging.

PR #2729, opened 2026-04-16
Stabilize generated type names per top-level statement

### LINKED ISSUE #2724: Replace nextint with deterministic names
Type checking currently generates a number of names, for example witnesses get a sequence ID like W7 for witness 7.

For incremental type checking or concurrent type checking, we need toa void this entirely. We want 100% deterministic names.

plajjan:
Top-level typing carried one TypeState across the whole module, and generated witness and temp names used bare nextint values. That meant an unrelated earlier top-level statement could shift later W_* and V_* names even when the typed definition itself was unchanged. In order to support concurrent or incremental type checking, we need each name to by type-checkable independently and thus it cannot share state or names with any other top-level statements.

This change makes top-level typing run each statement in a fresh TypeState, while still threading only the typed environment between statements. It also records a canonical top-level name for each statement and uses that tlname together with nextint when generating internal witness and temp names.

Recursive groups share one top-level name, chosen deterministically from their bound names. With fresh state already reset per top-level statement, this keeps the local numbering as it was while tying each generated name to the statement that produced it.

Fixes #2724


plajjan:
@nordlander PTAL

PR #2736, opened 2026-04-21
Parse module docstrings into AST

plajjan:
Module docstrings still survived parsing as synthetic top-level string statements, while declaration docstrings were already carried separately. Later passes had to recover module docs from the suite and still tried to re-extract declaration docs from declaration bodies.

Add a dedicated module docstring field to Module and populate it during parsing, alongside the existing declaration ddoc fields. Thread that through typing, printing, docs, hashing, and typed module stubs, and drop the fallback extraction paths and fake statement handling.

This gives docstrings a single representation from parse onward. Module and declaration docs now survive every pass without relying on body inspection, and the printed and typed forms stay aligned with the parsed AST.


plajjan:
@nordlander wanna look at this before merge? I think this is pretty straight forward. Moves module docstring parsing into parser instead of type checking. This was already the case for Decls but it is now uniform for all docstrings.

The TODO and the ugly match that was necessary in our top-level names generating code is now gone.

nordlander:
The description text already rings all the right bells – merge!

Issue #2733, opened 2026-04-20
Optional chaining, forced unwrapping cannot be used on bare statements


mzagozen:
### Acton Version

0.26.0.20260420.10.5.30

### Steps to Reproduce and Observed Behavior

I can't seem to use optional chaining (or forced unwraping) on bare statements. In all the tests and examples these are used on the right hand side of an expression. But this should work too?

```acton
def maybe_append(l: ?list[int]):
    l?.append(1)
```

```
acton: src/Acton/Parser.hs:374:1-30: Non-exhaustive patterns in function locate
```

### Expected Behavior

Either statements should be accepted by the compiler, or clearly stated in the docs that only expressions are supported.

plajjan:
@sydow do you mind having a look? :)

mzagozen:
Thanks @sydow! The test function I gave as an example now indeed compiles with your latest fix. But I found another one where type check succeeds, but then compile fails:

```
import logging

def maybe_log(l: ?logging.Logger):
    l?.error("boom")
```

```
install
+- install ActonProject
   +- compile lib ActonProject Debug aarch64-macos-none 2 errors
/private/tmp/boom/l.act:8:155: error: too few arguments to function call, expected 4, have 3

                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
/private/tmp/boom/l.act:14:5: error: expression result unused

    ^
error: the following command failed with 2 compilation errors:
```

sydow:
I need help from @nordlander with this. It seeems to me that the root cause is in the CPS conversion. If I place the failing acton code in test.act I get this:

````

acton> actonc newexamples/test.act --cps

== cps: test ================================
import logging

proc def maybe_log (C_cont : proc(None) -> $R, l : ?logging.Logger) -> $R:
    let:
        V_maybe_log_9: ?logging.Logger = l
        if $ISNOTNONE@[logging.Logger](V_maybe_log_9):
            C_1pre: None = $CAST@[?logging.Logger, logging.Logger](V_maybe_log_9).error("\"boom\"", None)
            C_2pre: None = C_1pre
        else:
            C_2pre: None = None
    in C_2pre
    return $R_CONTc@[None](C_cont, None)
=============================================

````

[...]

PR #2737, opened 2026-04-22
fixed bugs in CPS.hs; see comments in #2733 at github

sydow:
Fixed problems in the pre pass of CPS.hs for Let expressions and And/Or expressions.

### DIFF
--- a/compiler/lib/src/Acton/CPS.hs
+++ b/compiler/lib/src/Acton/CPS.hs
@@ -412,7 +412,11 @@ instance PreCPS Expr where
     pre env (TApp l e ts)               = TApp l <$> pre env e <*> pure ts
     pre env (Let l ss e)                = do ss' <- preSuite env ss
                                              (prefixes,e') <- withPrefixes $ pre env1 e
-                                             return $ Let l (ss' ++ prefixes) e'
+                                             -- Let suites are emitted directly by CodeGen and never
+                                             -- pass through the statement-level CPS conversion, so we
+                                             -- must lift their rewritten statements into outer prefixes.
+                                             prefix (ss' ++ prefixes)
+                                             return e'
        where env1                       = define (envOf ss) env
     pre env (Cond l e1 e e2)            = do (pre1,e1') <- withPrefixes $ pre env e1
                                              (pre2,e2') <- withPrefixes $ pre env e2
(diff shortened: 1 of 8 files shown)


sydow:
After several trivial nmistakes I managed to put together a PR which fails "only" in the four checks listed. I believe that only the acton-yang failure is important, but this needs to be looked at carefully. The test run report says the following:

````
 w3cregex_to_re_match:                                       FAIL           :  140 runs in 50.131ms @ 2792.7/s
    outcomes: 140 fail
    testing.NotEqualError: Expected equal values but they are non-equal. A: ^(a||b)$ B: ^(a|b)$
    @@ -1,1 +1,1 @@
    -^(a||b)$
    +^(a|b)$
    
   w3cregex_to_re_search_basic:                                FAIL           :  148 runs in 50.035ms @ 2957.9/s
    outcomes: 148 fail
    testing.NotEqualError: Expected equal values but they are non-equal. A: a||b B: a|b
    @@ -1,1 +1,1 @@
    -a||b
    +a|b
  ````

@mzagozen, can you look at these tests and say something about the results? And @nordlander, can you help by looking at my friend's code for And/Or expressions in the CPS pre pass?

If you want/need to run the compiler, it is branch fix-2733-again.

sydow:
I talked to my friend about the acton-yang problem, and got the following reply:

>I found the core issue: this one is in the source logic, not in CPS. In parse_regex, once one branch has been parsed, every later loop iteration with a non-empty elems appends |,  even if that iteration did not just consume a separator. That produces exactly a||b. I’m checking for any intended empty-branch behavior before I patch it.

He suggests the following patch to parse_regex

````
    312 -            if not ielems and self.try_get_char_in('|') is None:
    313 -                break
    312 +            if not ielems:
    313 +                if self.try_get_char_in('|') is None:
    314 +                    break
    315 +                elems.append('|')
    316 +                continue
    317              if ibp < bp:
    318                  bp = ibp
    316 -            if elems:
    317 -                elems.append('|')

````

leading to the modified version

````
    def parse_regex(self) -> (list[str], int):
        elems = []
        bp = BP_ATOMIC
        while True:
            ielems, ibp = self.parse_branch()
            if not ielems:
                if self.try_get_char_in('|') is None:
                    break
                elems.append('|')
                continue
            if ibp < bp:
                bp = ibp
            elems.extend(ielems)
        if len(elems) > 1:
            if BP_OR < bp:
                bp = BP_OR
        return elems, bp
````
So it would be good if @mzagozen could verify my friend's finding and if so update acton-yang. After that we make another try with this PR.

mzagozen:
@sydow yes, the fix is in and now this compiler branch passes the test.

PR #2756, opened 2026-05-04
Set DEVELOPER_DIR=/dev/null for actondb on macOS

mzagozen:
Mirror the workaround in Main.hs for avoiding Zig picking up broken Xcode SDK headers on macOS 26.4. In this case we invoke Zig outside of the Acton compiler.


mzagozen:
So this problem was dormant for a while, I just found it this morning. I've been on the latest Xcode CLI tools since #2715, but I guess my `dist/bin/actondb` didn't need rebuilding until now.

Issue #2707, opened 2026-03-30
Binaries compiled with ReleaseSafe crash on Linux x86_64 / aarch64


mzagozen:
### Acton Version

0.26.0.20260330.3.19

### Steps to Reproduce and Observed Behavior

I initially observed the problem with `--optimize ReleaseSafe` in sorespo, but the example below is sufficient to trigger a similar type of crash:

```acton
actor main(env):
    env.exit(0)
```

```
acton build --optimize ReleaseSafe safe.act
```

```
(lldb) bt
* thread #7, name = 'Worker 1', stop reason = signal SIGILL: illegal operand
  * frame #0: 0x00000000010f635a safe`$InitRootD___call__($this=<unavailable>, val=0x0000000001243a68) at rts.c:724:12
    frame #1: 0x00000000010f7f36 safe`wt_work_cb(ev=<unavailable>) at rts.c:1559:17
    frame #2: 0x00000000011afb82 safe`uv__run_check(loop=0x00007ffff7b6b400) at loop-watcher.c:67:1
    frame #3: 0x00000000011a6b30 safe`uv_run(loop=0x00007ffff7b6b400, mode=UV_RUN_DEFAULT) at core.c:462:5
    frame #4: 0x00000000010f8b1a safe`main_loop(idx=<unavailable>) at rts.c:1755:13
    frame #5: 0x000000000119a34c safe`GC_pthread_start_inner(sb=<unavailable>, arg=<unavailable>) at pthread_start.c:55:12
    frame #6: 0x000000000118e7cb safe`GC_call_with_stack_base(fn=(safe`GC_pthread_start_inner at pthread_start.c:45), arg=<unavailable>) at misc.c:2307:14
    frame #7: 0x00007ffff7c9caa4 libc.so.6`start_thread(arg=<unavailable>) at pthread_create.c:447:8
    frame #8: 0x00007ffff7d29c6c libc.so.6`__clone3 at clone3.S:78
```

### Expected Behavior

Produce a working binary that passes the safety checks of ReleaseSafe.

PR #2760, opened 2026-05-06
Use --release=fast for acton install on Linux

mzagozen:
ReleaseSafe trips UB checks at runtime on Linux (any arch), so fall back to ReleaseFast there. macOS and other hosts keep the default --release (=safe). See #2707


plajjan:
I think we should change `--release` to mean `--release=fast`... for the time being at least.

plajjan:
I did #2761 instead, changing `--release` to mean `--release=fast` for now, we obviously can't leave it so broken on Linux so before we can get around for the proper fix, this will have to do.

PR #2770, opened 2026-05-12
Add pkg update smoke tests for shipped artifacts

mzagozen:
Also test on Debian 13


mzagozen:
Only running on Debian 13 did not reproduce the segfault. The failing system also had additional NSS modules enabled in `/etc/nsswitch.conf`:

```
hosts: files myhostname dns
```

- plain `dns` is fine
- either `myhostname` or `resolve` is individually sufficient to trigger the crash

Sings point to statically linked libs with Acton compiler being responsible ...

PR #2789, opened 2026-05-18
Index active Env names

plajjan:
Env.names keeps the active name list used by the compiler. In many situations,
the order of names matters, so names is kept as an ordered list.

Many compiler paths repeatedly ask whether a name is in that list and what it
means. Since lookup scans the list from the front, the lookup costs grow with
the size of the list. That cost is present for every module and becomes
especially noticeable for large modules.

We now add hnames, a hash map from each name to its HNameInfo, to make lookups
fast. names is still kept as the ordered source of truth from which hnames is
produced. define, reserve, defineTVars, limitQuant, and Env substitution now
maintain the hnames index next to names, while order-sensitive traversals
continue to read names directly.

Route the central unqualified lookup paths through hnames so Kinds and Types
avoid repeated list scans without changing TEnv layout or attribute/MRO
ordering. On the heavy class benchmark at 1000 trees, the standalone Kinds
phase dropped from 83.96 s to 1.56 s. On a smaller 100 trees (I didn't want to
wait for type checking on 1000 trees) type checking dropped from over 300
seconds to around 100seconds. On a 200 trees module, type checking went from
over 2000 seconds to around 300 seconds. The difference increases with size,
showing the change from superlinear to being closer to linear (there are likely
many other parts that are not linear contributing time), yet we can still see
how the type checking time is still not linear (more than double; from 100s to
300s) for a twice as large module (100 -> 200 trees)

A new benchmark programming is added to benchmark the Kinds pass. It is rather 
useful as it forces the kinds check result and we can look at that independently
 from type checking. Kinds checking is simpler, so the name lookup make up a 
larger fraction of the total time, making it more extreme than the types pass.
 I think this benchmark can be useful going forward as well.

We now force the kinds checking result in the normal compiler pipeline as well.

### DIFF
--- a/compiler/lib/bench/KindsBench.hs
+++ b/compiler/lib/bench/KindsBench.hs
@@ -57,6 +57,7 @@ main = do
         printStatsMaybe "parse_stats" s0 s1
 
         env <- Env.mkEnv [typesPath] env0 parsed
+        E.evaluate (forceHTEnv (Env.hnames env))
         E.evaluate (forceHTEnv (Env.hmodules env))
         t2 <- getCurrentTime
         s2 <- getStats statsEnabled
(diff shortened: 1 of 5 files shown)


plajjan:
@nordlander I think you should review this

PR #2790, opened 2026-05-18
Add optional OCI output artifacts

plajjan:
Support optional Acton output artifacts fetched from OCI repositories during dependency resolution. Artifacts are keyed by source content hash, interface version, and target tuple.

Use valid local cache entries before registry lookups and fall back to source when no matching artifact is available.

Add artifact pack, push, and hash commands, local OCI layout support, tests, and guide documentation.

### DIFF
--- a/compiler/acton/PkgCommands.hs
+++ b/compiler/acton/PkgCommands.hs
@@ -369,6 +369,7 @@ packActonArtifactTo sourceHash outputArg = do
       Artifact.writeManifest tmp (Artifact.expectedManifest sourceHash)
       runProcessChecked Nothing "tar"
         [ "-czf", output0
+        , "--exclude=*.c", "--exclude=*.h"
         , "-C", tmp, Artifact.artifactManifestFile
         , "-C", cwd, "Build.act", "out/types"
         ]
--- a/compiler/acton/PkgCommands.hs
+++ b/compiler/acton/PkgCommands.hs
@@ -7,6 +7,7 @@ module PkgCommands
   , pkgUpgradeCommand
   , pkgUpdateCommand
   , pkgSearchCommand
+  , artifactCommand
   , zigPkgAddCommand
   , zigPkgRemoveCommand
   , PackageEntry(..)
@@ -24,6 +25,7 @@ import Prelude hiding (readFile, writeFile)
 
 import Acton.ArchiveDownload (downloadArchive)
 import Acton.HttpFetch (httpGetBytes, isHttpUrl, newProxyManager)
+import qualified Acton.Artifact as Artifact
 import qualified Acton.BuildSpec as BuildSpec
 import qualified Acton.CommandLineParser as C
 import Acton.Compile (loadBuildSpec, throwProjectError)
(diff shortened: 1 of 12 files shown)


compiler/lib/src/Acton/Artifact.hs
```diff
@@ -0,0 +1,255 @@
...
+artifactRepositoryName = "acton-out"
+
+artifactMediaType :: String
+artifactMediaType = "application/vnd.acton.out.v1+tar+gzip"
+
+artifactType :: String
+artifactType = "application/vnd.acton.out.v1"
+
+currentTargetTuple :: String
+currentTargetTuple = ""
```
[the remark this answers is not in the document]
plajjan: Fair, but our .ty / .c / .h output currently is not platform dependent. I kept the target tuple in here as a future kind of thing but left the actual value as an empty string in order to signal that right now, we effectively do not produce these output artifacts for any particular platform.

docs/acton-guide/src/output_artifacts.md
```diff
@@ -72,9 +72,15 @@ Once an artifact has been pulled into the dependency cache, later plain
...
-artifact repositories.
+requested source content hash, interface version, and target before using it.
+
+Passing `--artifact-repo` opts the build into artifact consumption: a valid
+artifact is preferred even when a cached source entry already exists, and the
+cached source is used only when no artifact can be found. This matters on a
+machine that has published the artifact itself, since producing seeds the
+caches with the package source, which would otherwise always win over the
+artifact. Without `--artifact-repo`, a cached source entry is used as is,
+without probing remote artifact repositories.
```
plajjan, line 83: It seems to me that the only use of this option is to test the artifacts. The design is that the cached artifact output is 100% deterministic, so there is no such case as "invalid cache / artifact", so why would anyone need this?

If you want to clear the cache for testing, I think it is better you run `rm -rf ~/.cache/acton/...`

compiler/acton/PkgCommands.hs
```diff
@@ -369,6 +369,7 @@ packActonArtifactTo sourceHash outputArg = do
       Artifact.writeManifest tmp (Artifact.expectedManifest sourceHash)
       runProcessChecked Nothing "tar"
         [ "-czf", output0
+        , "--exclude=*.c", "--exclude=*.h"
```
plajjan, line 372: I don't want options for this. I am fine to remove the .c and .h file entirely from the artifact since it is only the TYDB that can truly be cached. I should probably have made it only TYDB from the beginning...

compiler/lib/src/Acton/CommandLineParser.hs
```diff
@@ -91,6 +91,7 @@ data CompileOptions   = CompileOptions {
                          optimize    :: OptimizeMode,
                          only_build  :: Bool,
                          skip_build  :: Bool,
+                         skip_backpass :: Bool,
```
plajjan, line 94: This should not be an option. New selective back loader will see 0 roots for typical device model project, so no output .c files will be produced at all.

mzagozen:
@plajjan I rebased this branch many times over. It predated DBP, and at one point was also based on `dce`. I was only focused on testing the small-includes-big pattern. So don't read too much into the options, feel free to remove what doesn't fit :)

PR #2791, opened 2026-05-19
Limit Env scans to live names

plajjan:
Type checking repeatedly substitutes Env and asks Env for polarity while
improving constraints. Env.names uses list semantics and also contains
previously defined names, so scanning the whole list in these paths makes
all modules pay for accumulated environment size. That cost is particularly
visible as modules get larger.

Track how many leading names may still contain unification variables.
define, reserve, defineTVars, and setNames update that live-prefix length
when they change names. Env substitution and Env polarity then inspect only
that prefix; the closed suffix remains shared and contributes no unification
variables.

On the heavy-class type benchmark, the 100-tree type phase dropped from
about 90.9 s before this Env-scan work to 15.7 s. A 300-tree run with this
patch still takes about 155 s, so this removes one expensive Env scan but
certainly does not make the full type pass scale linearly yet.

### DIFF
--- a/compiler/lib/bench/TypesBench.hs
+++ b/compiler/lib/bench/TypesBench.hs
@@ -3,6 +3,7 @@ import qualified Acton.Kinds as Kinds
 import qualified Acton.NameInfo as NameInfo
 import qualified Acton.Parser as Parser
 import qualified Acton.Syntax as Syntax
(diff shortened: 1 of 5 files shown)


plajjan:
Implemented a better solution in #2796. Closing this.

PR #2816, opened 2026-05-27
Avoid repeated Env attribute lookup work

plajjan:
This PR makes a few Env attribute lookups do only the work they need.

`findAttr`, `closedAttr`, `abstractAttr`, and the attribute-membership helpers used to build full inherited attribute lists before keeping one result. The new code walks the same attribute environments directly and avoids those temporary lists. It also shares the active constructor enumeration used by `activeConAttr` and `allConAttrUFree`.

Simple benchmark on the heavy class fixture with `--trees 250`, using `compiler-bench --pipeline`:

- Full pipeline: 29.57s -> 23.19s, about 1.28x faster
- Types: 10.25s -> 9.65s, about 6% faster
- Back passes: 15.13s -> 9.37s, about 1.61x faster
- Back-pass allocation: 30.0 GB -> 18.0 GB


nordlander (approved):
All changes look good -- merge!

PR #2823, opened 2026-05-29
Refresh Acton compile cache via nightly snapshots

plajjan:
## Summary
- Make the zig compile cache (`~/.cache/acton`) self-refreshing instead of frozen behind a static key, which went stale and caused slow cold rebuilds (the >1h `test-macos` runs) when evicted.
- **Nightly (schedule)** builds cold and saves a date-stamped snapshot (`acton-zig-<os>-<version>-<arch>-<date>`). **PRs and pushes are restore-only** — they pull the latest snapshot via `restore-keys` and never save, so they're fast and can't accumulate dead objects.
- Split the Haskell deps cache (`~/.stack`) onto its own `stack-*` key with `restore-keys`; add a zig cache to `test-linux` (previously had none).
- Add a `workflow_dispatch` trigger that behaves exactly like the nightly run, so the cache can be populated on demand instead of waiting for the next 00:04 UTC schedule.

## Why
The repo sits at GitHub's cache size ceiling, so the static-keyed zig cache got LRU-evicted during busy daytime PR activity → cold macOS rebuilds (60–100 min vs ~20–30 min warm). A static key also means `actions/cache` only writes on a miss, freezing the cache forever once populated. Date-stamped key + nightly producer keeps it both warm and fresh.

## Rollout / test plan
- [ ] Merge this PR.
- [ ] Trigger a manual run from `main` (Actions → "Build, Test & Release" → Run workflow) to produce the first snapshot immediately. Note: a manual run from main is a full nightly — it also runs `pre-release-tip` and `update-apt-tip-repo`, same as the scheduled nightly.
- [ ] Confirm the "Save Acton compile cache" step runs on the `cache: true` jobs.
- [ ] Open a trivial PR and confirm "Restore Acton compile cache" reports a hit and `test-macos` is warm/fast.

## Notes
- One-time transient: until the first nightly/manual run lands a snapshot, PRs/pushes build the zig cache cold.
- Stack cache prefix renamed `test-*` → `stack-*` so its `restore-keys` can't match the old `test-*-acton` zig caches; old caches age out in 7d.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -137,9 +137,11 @@ jobs:
       - name: "Install build prerequisites"
         run: brew install haskell-stack
       - name: "Build Acton"
+        id: build_acton
         run: |
           make -j2 -C ${{ github.workspace }} BUILD_RELEASE=${{ env.BUILD_RELEASE }} BUILD_TIME=${BUILD_TIME}
       - name: "Build a release"
+        id: build_release
         run: make -C ${{ github.workspace }} release BUILD_TIME=${BUILD_TIME}
       - name: "Upload artifact"
         uses: actions/upload-artifact@v7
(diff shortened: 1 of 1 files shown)


plajjan:
Good catch — addressed in fc04eb363. The save step now requires `steps.build_acton.outcome == 'success' && steps.build_release.outcome == 'success'` (in both `test-macos` and `test-linux`), so a failed build can no longer poison the date-stamped key with a partial cache. `!cancelled()` is retained intentionally so a flaky *test* still doesn't deny PRs a warm cache.

PR #2824, opened 2026-05-29
New unbox

sydow:
First version with total unbox of bounded ints and floats


base/src/testing.act
```diff
@@ -474,22 +474,22 @@ class UnitTest(Test):
         self.module = module
 
     def run_test(self, report_result: action(?bool, ?Exception, ?str) -> None, env: Env, log_handler: logging.Handler, tags: set[str]):
-        output = None
+        output: ?str = None
         success = None
         exception = None
         try:
-            output = self.fn()
```
plajjan, line 481: This is wrong, I don't think you really meant to make this change. It is what is causing the test problem. I will fix.

nordlander (commented):
Looks good to me! (The complexity of CodeGen had grown large long before this PR...)

plajjan:
FYI; I will take over this branch, rebase it etc - I noticed there are a few small mistakes, more technical in nature that I think come from rebasing issues or similar. I will clean up that as well as updating goldens etc to make sure the tests passes. Thanks for the fantastic work :) there will be downstream dependencies to be updated in all libraries that do C bindings basically - I will fix those too in time...

PR #2828, opened 2026-05-30
Lower Linux glibc floor to 2.28

plajjan:
Target glibc 2.28 instead of 2.31 so released Linux binaries run on OracleLinux 8 / RHEL 8 (and Debian 10, Ubuntu 18.04+). 2.28 was Acton's historical floor; the 2.31 bump was only to match the Debian 11 release host and was never a hard promise.

The floor is set through Zig's -target <arch>-linux-gnu.2.28, so it is independent of the build host's glibc. The existing `make ldd` check fails the build if acton, lsp-server-acton, or actondb require a GLIBC symbol newer than the target, so the ceiling stays enforced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,11 @@
   instead of rejecting the default stop position. [#2801]
 
 ### Packages & Distribution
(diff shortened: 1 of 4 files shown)


plajjan:
Superseded by #2959. Closing.

PR #2833, opened 2026-06-01
Avoid repeated substitution state reads

plajjan:
Substituting a type used to read the current unifier state again for each nested type node. The state is unchanged during that walk, so this repeated monadic state access without changing the result.

Read the substitution map once for Type and TCon substitution, then recurse through small pure helpers. Substitution chains are still followed, TOpt flattening is preserved, and all Type constructors remain explicit.

On latest main, the 500-tree class-heavy compiler-bench --types run had type checking at 10.39s, 10.36s, and 10.36s before this change, and 8.23s, 8.31s, 8.28s, and 8.23s after this change. Type allocation dropped from about 128.0 GB to about 101.1 GB.

### DIFF
--- a/compiler/lib/src/Acton/TypeEnv.hs
+++ b/compiler/lib/src/Acton/TypeEnv.hs
@@ -656,8 +656,34 @@ instance USubst TVar where
                                             TVar _ v' -> return v'
                                             _         -> return v
 
+usubstTConWith                      :: IntMap Type -> TCon -> TCon
+usubstTConWith s (TC n ts)          = TC n (map (usubstTypeWith s) ts)
+
+usubstTypeWith                      :: IntMap Type -> Type -> Type
+usubstTypeWith s (TUni l u)         = case IntMap.lookup (uvid u) s of
+                                           Just t  -> usubstTypeWith s t
+                                           Nothing -> TUni l u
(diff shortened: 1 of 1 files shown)


plajjan:
@nordlander can you take a look at the diff please - is this OK? aesthetics change a bit, not sure how we can get perf improvement with the same style

nordlander:
Looks good to me. But I'm really surprised that this change can make any measurable improvement. My understanding is that ghc is heavily armed to perform very similar rewrites as part of its optimization of monadic code. Apparently there's still something in this realm it doesn't nail...

nordlander:
But of course!

I've been so fixated with `usubstitution` and the fact that it doesn't modify the `TypeM` state that I missed the obvious: we have no right to expect ghc to spot that `usubstitution` is the _only_ monadic operation that occurs during a `usubst` traversal -- especially considering that `usubst` is recursive! So narrowing the meat part of `usubst` to just depend on the current substitution is indeed the right thing to do, otherwise every monadic step of `usubst` will spend a fresh tuple pairing its result with the current state (in total ignorance of the fact that this state has remained constant).

But this also means that we can make the same pattern apply to all `USubst` instances, and possibly gain some further speedups along the way. I'm pushing an update to this effect.

PR #2840, opened 2026-06-02
Simplify mapModules callback

plajjan:
The linear mapModules implementation no longer threads an updated environment through each callback. Keeping Env0 in the callback type implies lookup behavior that the traversal does not provide.

Pass only the containing module name plus the current entry. Existing callers either ignore the environment or close over their own stage environment, so the API now matches the behavior they need.

### DIFF
--- a/compiler/lib/src/Acton/Converter.hs
+++ b/compiler/lib/src/Acton/Converter.hs
@@ -240,15 +240,15 @@ fixupSelf s                             = s
 
 convEnvProtos env                       = mapModules conv env
   where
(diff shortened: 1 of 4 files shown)


plajjan:
@nordlander this you had in mind?

nordlander:
Indeed!— Johan2 juni 2026 kl. 15:45 skrev Kristian Larsson ***@***.***>:﻿plajjan left a comment (actonlang/acton#2840)
@nordlander this you had in mind?

PR #2850, opened 2026-06-03
Run cross-compilation in a dedicated CI job

plajjan:
Cross-compilation is the reason the test-X jobs disk-died on cache restore. The cross-compilation tests build hello.act for many targets (aarch64/x86_64 macos, windows-gnu, linux-gnu, linux-musl), which pulls large per-target objects into ~/.cache/acton. Because those tests ran as part of make test in every test-macos / test-linux job, they bloated the zig cache those jobs save and restore — and restoring that oversized cache exhausted disk on the linux container runners, killing the runner mid-extraction (the silent hang/failure during "Restore Acton compile cache").

This moves cross-compilation out of the default make test suite and into a standalone test-cross-compile job. crossCompileTests is gated behind an ACTON_TEST_CROSS env var so it is excluded from the normal suite (keeping the test-X zig caches small) but still runnable; the make test-cross-compile target sets the var. The new job installs acton from the build-debs .deb (so it needs no stack or zig build cache of its own) and builds hello.act for all targets. It runs on a host runner rather than in a container, so it can free ~15-30 GB of preinstalled toolchains we never use, giving cross-compilation plenty of headroom.

Net effect: the test-X jobs no longer cross-compile, so their ~/.cache/acton shrinks substantially and the disk-exhaustion deaths go away; cross-compilation coverage is preserved in a fast, isolated job. The release and pre-release-tip jobs gain test-cross-compile in their needs so a cross-compile regression still blocks a release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -626,6 +626,62 @@ jobs:
           path: |
             ${{ github.workspace }}/core*
 
+  # Cross-compilation is expensive on disk: building for many targets
+  # (macos/windows/musl/linux-gnu) pulls large per-target objects into the zig
+  # cache. Rather than do that inside every test-X job (which bloated their
+  # ~/.cache/acton and exhausted container disk on restore), exercise it once
+  # here, on the host runner (not a container, so we can free host disk), using
+  # the installed .deb and an isolated cache.
+  test-cross-compile:
+    needs: build-debs
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [amd64, arm64]
+    runs-on: ${{ matrix.arch == 'amd64' && 'ubuntu-latest' || 'ubuntu-24.04-arm' }}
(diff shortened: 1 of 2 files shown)


plajjan:
Good catch on the missing checkout (P1) — fixed in c1ffe4b7b. The job builds the test/compiler/hello project from the repo, so it now runs actions/checkout before download-artifact (checkout cleans the workspace, so the .deb download has to come after).

Also reworked how cross-compilation is excluded from the normal run, per review discussion: instead of an ACTON_TEST_CROSS env var + test.hs change, `make test` now just filters it out with a tasty pattern (-p "! /cross-compilation/") and `make test-cross-compile` selects it (-p "/cross-compilation/"). Verified against the built binary: 306 tests total, 300 with the negation (the 6 cross tests excluded), 6 with the positive filter.

PR #2852, opened 2026-06-03
Free host disk in test-linux containers

plajjan:
The test-linux jobs run inside a container (image: debian/ubuntu), so a normal host-level disk cleanup can't reach the host's large preinstalled toolchains (dotnet, android SDK, ghc, hostedtoolcache, boost) that occupy most of the runner disk. The container's steps only see the container filesystem. But the container's root overlay lives on the same host disk as those toolchains, so if we bind-mount them in, we can delete them from within the container and reclaim real space.

This adds a volumes: block to the test-linux container mapping each big host path to /host-disk/*, and a "Free host disk space" step that runs first (before the compile-cache restore that was exhausting disk and killing the runner mid-extraction) and rm -rf's them. Removals are defensive — a path missing on a given runner is ignored, and bind-mounting a nonexistent host path just yields an empty dir — so it's safe across the amd64 and arm64 images. df -h is printed before and after so we can see how much was reclaimed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -254,6 +254,19 @@ jobs:
     runs-on: ubuntu-24.04${{ matrix.arch == 'arm64v8' && '-arm' || '' }}
     container:
       image: ${{ matrix.os }}:${{ matrix.version }}
+      # Bind-mount the host's large preinstalled toolchains so the step below
+      # can delete them from inside the container. The container's root overlay
+      # lives on the same host disk as these, so freeing them gives the cache
+      # restore + build headroom (these jobs were dying of ENOSPC mid-restore).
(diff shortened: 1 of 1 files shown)


plajjan:
Both addressed in 21826d372:

1. Gating: added test-cross-compile to the needs of all publishing jobs — update-apt-repo, update-apt-tip-repo, update-homebrew, and also build-container-image (it pushes to ghcr) — so a cross-compile regression can no longer advance the apt/Homebrew/container channels. Previously only release and pre-release-tip were gated on it.

2. Windows --no-threads: the two *-windows-gnu builds now pass --no-threads, matching AGENTS.md (Windows has no RTS thread support) so we exercise the build mode we actually ship. Worth noting the original crossCompileTests in test.hs also omitted this, so this CI job is now stricter than the old in-suite test was.

PR #2873, opened 2026-06-04
Run one macOS source test on PRs

plajjan:
Pull request runs currently enqueue the same three macOS source-test variants as the full main and scheduled matrix. Those jobs use scarce runners, so otherwise useful PR feedback can sit behind macOS queue time even when the Linux, package, and downstream app jobs have already finished.

This keeps the primary macOS 15 arm64 source test on pull requests and leaves the full macOS matrix on main pushes, tag pushes, scheduled runs, and manual runs, so broader macOS coverage still happens outside the PR feedback loop.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,10 +90,18 @@ jobs:
     outputs:
       matrix: ${{ steps.setmatrix.outputs.matrix }}
     steps:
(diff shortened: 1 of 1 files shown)


plajjan:
Superseded by #2992. Closing.

PR #2886, opened 2026-06-05
Fix processing comments in XML

### LINKED ISSUE #2849: XML comment followed by more characters
### Acton Version

0.27.0.20260603.9.8.17

If comment in an XML node is followed by more characters (even whitespace), then the node *text* is incorrectly parsed:
```xml
<data><!-- comment --> </data>
```

```
   xml_comment:                        ERR            :  776 runs in 50.280ms @ 15433.6/s
    outcomes: 776 err
    xml.XmlParseError: Unexpected nodetype 3, content is  
```

### Expected Behavior

nodetype 3 is `XML_COMMENT_NODE`. The function `collect_text_cdata_nodes` in *xml.ext.c* incorrectly stops processing the element contents when it encounters a comment node - it only collects `XML_TEXT_NODE` and `XML_CDATA_SECTION_NODE`. It should skip comments entirely.

mzagozen:
When collecting element text (text and CDATA), skip over comments that appear before the current node closing tag or another child node opening tag. Unsupported nodes (entity references, processing instructions) are rejected.

Closes #2849

### DIFF
--- a/base/src/xml.ext.c
+++ b/base/src/xml.ext.c
@@ -68,56 +68,70 @@ static unsigned char* copy_with_xml_escape(unsigned char *dst, B_str src, int es
     return dst;
 }
 
-// Helper function to collect text from consecutive TEXT and CDATA nodes
-// Returns the combined string and updates the node pointer to the first non-text node
-// Note: cur_ptr is passed by reference (pointer to pointer) so we can update the caller's pointer
-//       to skip past all consumed text/CDATA nodes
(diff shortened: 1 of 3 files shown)


mzagozen:
We still don't have a way of representing comments in the `std.xml.Node` object (also see https://github.com/actonlang/acton/pull/2148#issuecomment-2629356862), but this fixes the immediate problem of parsing an XML document with inlined comments.

PR #2897, opened 2026-06-08
Hash AST fragments via serialization instead of pretty printing AST

plajjan:
Per-name source and implementation hashes currently render top-level fragments through the pretty-printer before hashing them. That makes hashing pay for text rendering even though the compiler already has the parsed and typed ASTs.

This strips source locations from each fragment, encodes the normalized AST payloads directly, and uses persist for the hash payloads. Location-only changes stay out of the cache keys, and the hash path avoids the pretty-print and String-to-ByteString conversion work.

On the generated class-heavy front-pass benchmark with 3000 trees, with both sides using persist for hash payloads, the hash phase dropped from 24.77s to 15.07s over three-run averages. Total front-pass time dropped from 53.61s to 44.02s, and front-pass allocation dropped from 488.7 GB to 445.4 GB.

### DIFF
--- a/compiler/acton/test/parse/simple.all.golden
+++ b/compiler/acton/test/parse/simple.all.golden
@@ -66,7 +66,7 @@ pure def main () -> None:
     return None
 ===============================================
 
-/* Acton impl hash: 0ca2a92cafc149a9706464cc8c1405785ee688eb49a8ebe57c2f63bc80d954c1 */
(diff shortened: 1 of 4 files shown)


plajjan:
@nordlander it wasn't 20% like I guessed, it was mooore! and this is just the improvement going from pretty printing to directly serializing the AST (using new persist library) to bytes and hashing that.. after stripping srcLoc. My idea to directly hash the AST should reduce this even further!

nordlander:
Great news!

PR #2903, opened 2026-06-09
Direct calls

sydow:
This PR implements an optimization to certain calls of methods in protocol instances, as illustrated by the following trivial example: The Acton function definition

````
def app(lst: list[int],x):
    lst.append(x)
````

has upto now been compiled to the C code

````
B_NoneType appendQ_app (B_list lst, int64_t x) {
    B_Sequence W_app_7 = B_SequenceD_listG_witness;
    W_app_7->$class->append(W_app_7, lst, toB_int(x));
    return B_None;
````

where we, for readability, omitted some type casts. After this PR the result is instead

````
B_NoneType appendQ_app (B_list lst, int64_t x) {
    B_SequenceD_listD_append(B_SequenceD_listG_witness, lst, toB_int(x));
    return B_None;
}
````

There are two changes:

- the indirect call to the implementation of append for lists via the witness method table has been replaced by a direct call to the method, which here has a known name.
- the name W_app_7 is never introduced.

In this PR, the simplification/optimization applies only to instances of builtin protocols (here Sequence) to builtin types (here lists). These all have static witnesses defined and the name convention for naming the specific method to be called directly is known. (Note: For protocols with restrictions on a type parameter, such as Mapping and Set, which require the type of keys/elements to be Hashable, not all combinations of protocol and type have presently static witnesses implemented.)


plajjan:
@sydow at this point, wouldn't it be a good idea to look at the generated machine instructions for functions like this so we can see what is actually doing? All the C-compiler optimizations are otherwise opaque to us, right?

sydow:
@plajjan, of course it would be absolutely wonderful if someone could look at this, and more generally, at what optimizations the C compilers do to our code (and how we could influence that). For the present PR, I think that the simplification of the code and the removal of some witnesses is in itself worthwhile, even if it should not give much speedup.

sydow:
This is replaced by PR #2935.

PR #2916, opened 2026-06-11
Selective tydb reads for imported modules

plajjan:
This makes the compiler read imported modules' .tydb interfaces selectively instead of decoding them in full. Until now, compiling a module forced the complete interface of every transitive import: doImp decoded the whole .tydb, initTypeEnv folded every imported constructor and extension witness into the type-check environment, the solver's candidate queries scanned every imported constructor, each back pass rewrote every loaded module's interface, and cache checks decoded whole per-name hash sections. A small module importing a large one therefore paid for the dependency's size on every rebuild.

The series introduces ModuleInfo as the only environment representation of dependency modules: a per-module view that answers questions through lookup functions rather than exposing a materialized TEnv. A locally compiled module backs it with its in-memory interface; an imported module backs it with a long-lived read-only LMDB handle (InterfaceDB) doing keyed reads that are memoized per handle. To support this, the .tydb format gains narrow per-module query indexes written next to the existing keyed name entries: public names, constructors, actors, attribute owners, descendants, and extensions keyed by protocol and by type. The version is bumped to [0,25].

The conversion is staged for review, so the commits are best read in order, oldest first. First the plumbing lands (the indexes and read API, then ModuleInfo alongside the old modules/hmodules storage), then consumers are rerouted one group per commit while the two representations are kept in sync and behavior is preserved: qualified-name lookup, imports, the solver-facing queries, the type-check environment, tools, and the back passes (which now convert imported entries on demand instead of rewriting whole module trees). Only then does the lazy flip happen: doImp and cached-module reuse install .tydb-backed shells, dependency hashes resolve per name, and cache checks read a header summary with a stored name count instead of every hash row. The final step deletes the NModule tree and its hashed mirror, leaving the ModuleInfo map as the only module storage.

A deliberate design constraint throughout: no change to inference behavior, and Solver.hs is untouched. The attribute indexes record attributes where they are declared, and readers complete inherited owners through the descendants of each declaring constructor, preserving the ancestry-aware semantics of the old hasAttr scans. Witness queries merge keyed extension-index reads over all transitive imports, so cross-module extensions keep working and query cost follows the number of imported modules, never their size.

The benchmark this targets is a project with a generated class-heavy module (test/compiler/concurrent_typecheck_class_heavy/generate_heavy.py) and a small module importing it. Rebuilding only the small module takes 0.09s whether the dependency has 3,000, 10,000 or 30,000 classes; on main the same rebuild takes 0.55s, 1.6s and 5.8s respectively. Full builds also got faster, mostly from the dependency-index writer deduplicating user lists with sets instead of quadratic nub: the 30,000-class project drops from 345s to 96s.

New traced regression tests pin the behavior down with ACTON_TYDB_TRACE_READS: a small importer must perform keyed lookups only (exact name reads, attribute-index reads, keyed extension reads) with no whole-section decodes, an unused import must not read names at all, and an unrelated added name in the dependency must keep the importer fresh. A module_lookup compiler test project covers imported classes, extension witnesses and actors end to end. The lib, incremental and compiler test suites all pass.

Known follow-up work, intentionally out of scope here: when DBP has to regenerate a large dependency's code (its selection changed or the module itself changed), prepareDeferredBackJob still does one full typed-module read before pruning; storing per-name statement indexes so only the selected statements are read is the natural next step. Test discovery in acton test still decodes whole name-hash sections, initEnv still reads __builtin__ in full once per process, and InterfaceDB handles rely on a GC finalizer backstop, which is fine for CLI compiles but worth revisiting for the long-lived LSP process.

This PR is arranged as a, rather long, series of commit. Each make sense on its own, introducing one conceptual change. It is expressly written to be simple to review.

### DIFF
--- a/compiler/acton/Main.hs
+++ b/compiler/acton/Main.hs
@@ -2353,9 +2353,9 @@ writeRootC env gopts opts paths tasks binTask = do
         rootsHeader <- case tyPath of
                          Just ty -> do (_sourceMeta, _, _, _implH, _imps, _depModules, _nameHashes, roots, _tests, _) <- InterfaceFiles.readHeader ty; return roots
                          Nothing -> return []
-        let rootsEnv = case Acton.Env.lookupMod m env of
+        let rootsEnv = case Acton.Env.lookupModuleInfo m env of
                          Nothing -> []
-                         Just te -> [ n' | (n', i) <- te, rootEligible i ]
+                         Just mi -> [ n' | (n', i) <- Acton.Env.modulePublicTEnv mi, rootEligible i ]
             shouldGen = n `elem` rootsHeader || n `elem` rootsEnv
         if shouldGen
           then do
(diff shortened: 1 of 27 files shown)


plajjan:
@nordlander btw, this is going to conflict with the changes in the other branches. I think #2924 should take priority, so please review that one first, and we will have to rebase and resolve the conflicts in this branch.

plajjan:
Closing in preference of #2930

PR #2919, opened 2026-06-12
Index name environment lookups

plajjan:
The type environment keeps all known names in association lists (activeNames/closedNames). Several lookups scanned those lists linearly, and since every top-level statement adds a name, each such lookup costs O(names defined so far): summed over a module this is quadratic, and it is why type checking a 288k-statement generated module slowed down as it progressed.

The worst offender was findDefLoc, which inferCall consults for every call expression with an unqualified callee just to build the error-message DeclInfo - i.e. a full scan of the name list per call expression. findSigLoc (signature/definition compatibility errors), stateScope (consulted per def), the kinds pass's tvKind/tvars (per type-variable occurrence in every signature), Simp's alias restoration (per printed qualified name), and free-variable filtering (diffQ, per statement) all had the same shape.

This commit mirrors the name lists with purpose-built indexes that are maintained at exactly the same write sites as the pre-existing hnames hash index (setActiveNames/addActiveNames/addClosedNames/setX/initEnv), so the mirror invariant is local to a handful of functions:

- sigLocs/defLocs: HashMap Name SrcLoc of NSig/NDef entries; first- match/shadowing semantics reproduced by insert-overwrites-older ordering (foldr over the prepended TEnv), verified against degenerate duplicate orderings.
- activeStateNames/closedStateNames: the NSVar names, prepended in the same order the old comprehension produced.
- activeTypeVars/closedTypeVars: the NTVar entries; lookupTypeVarKind is name-keyed first-match, identical to the old filter (==) because TVar equality is name-only.
- Simp QName first consults the hnames index for the common bare-name alias before falling back to the old linear scan, so output can only differ when several distinct aliases target the same qualified name.
- diffQ switches to Set membership beyond 8 bound names (Eq/Ord Name agree, so semantics are unchanged).

Substitution and scope truncation (usubstWith, limitQuant) go through setActiveNames and rebuild the active-side indexes from the closed base, so no stale entries can survive a scope pop; closedNames may not contain unification variables, so closed-side indexes are never invalidated.

### DIFF
--- a/compiler/lib/src/Acton/Env.hs
+++ b/compiler/lib/src/Acton/Env.hs
@@ -53,6 +53,14 @@ data EnvF x                 = EnvF {
                                 closedNames:: TEnv,
                                 hnames     :: HTEnv,
                                 closedHNames:: HTEnv,
+                                sigLocs    :: LocEnv,
+                                closedSigLocs:: LocEnv,
+                                defLocs    :: LocEnv,
+                                closedDefLocs:: LocEnv,
+                                activeStateNames:: [Name],
+                                closedStateNames:: [Name],
+                                activeTypeVars:: [(Name, Kind, CCon)],
+                                closedTypeVars:: [(Name, Kind, CCon)],
                                 imports    :: [ModName],
                                 improots   :: [Name],
                                 modules    :: TEnv,
@@ -65,6 +73,7 @@ data EnvF x                 = EnvF {
                               } deriving (Show)
 
 type Env0                   = EnvF ()
+type LocEnv                 = M.HashMap Name SrcLoc
 
 -- activeNames may contain live unification variables; closedNames must not.
 -- hnames is the full active-over-closed lookup index, while closedHNames is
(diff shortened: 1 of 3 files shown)


plajjan:
@nordlander please take a look at this. This is the first of a number of changes that I'm pulling out of my experimental branch to speed up type checking even further. I collated a few smaller hacks into what I think makes sense as a conceptual change. Perf for a relatively small test file goes from 92s -> 70s from this, so I think the gain is even bigger for larger files

nordlander:
The code looks really good and its purpose entirely obvious and motivated. One observation, though: neither state variables nor type variables exist on the Acton top level so their corresponding _closed_ data structures will always be empty. I understand this wasn’t obvious from the way the code was written before, but it’s a consequence of the Acton semantics.

I also have some ideas on how to get rid of the current ”simp” stag of right before printing, meaning its current need to scan for aliases will eventually disappear too. But this commit won’t disturb any of that work so go ahead and merge.

PR #2924, opened 2026-06-13
Index name, solver candidate and witness lookups in the type checker

plajjan:
Type checking very large generated modules slowed down dramatically as the checker progressed through the file. A 288k statement module generated from the juniper_crpd YANG models took around two and a half hours, with the cost of each statement growing the further into the module the checker got. The cause was a handful of linear scans over data structures that grow with every statement processed, which makes each scan quadratic when summed over a whole module. This branch removes three such scans, as three independent commits, each with a detailed commit message describing the scan it eliminates and why the replacement preserves behaviour.

The first commit mirrors the name environment association lists with purpose built indexes. Several lookups scanned activeNames and closedNames linearly, and since every top level statement adds a name, each lookup cost time proportional to the number of names already defined. The worst offender ran for every call expression, just to recover a source location for a potential error message. The new indexes cover signature and definition locations, state variable names and type variables, and are maintained at exactly the same points as the existing hnames hash index, so the invariant that they mirror the name lists stays local to a few functions. State variables and type variables only ever exist inside actor, class and def bodies and never among the closed top level names, so those two keep no closed side structure at all, which the commit notes is a consequence of Acton semantics rather than an omission.

The second commit makes the previously debug only type id graph the source of the solver's candidate sets. When the solver ranks alternatives for an unresolved goal it needs candidate types, for example all classes or protocols carrying a given attribute, or all descendants of an upper bound. These were produced by scanning every type in scope and walking the ancestry of each, so candidate generation grew with every class and extension the module defines. The graph already recorded the necessary relationships but was only used for debug output. Candidate enumeration now becomes a set lookup, with the ascending traversal of integer id sets reproducing the previous definition order exactly. The commit also fixes a latent double indexing of the builtin module that would otherwise feed duplicated candidates to the solver and change which solutions it commits, so inference results now match the previous compiler exactly. For the common case where candidate lists are tiny, the set operations fall back to plain list operations, because hashing the long qualified names of generated types is more expensive than a structural comparison that stops at the first difference.

The third commit adds a second witness map keyed by the implementing type. Witness lookup by type and the duplicate check performed on every witness insertion both scanned the witness bucket of a protocol, and popular protocols such as Iterable and Indexed gain one witness per implementing type, so the bucket grows with every extension in the module. Keyed by the implementing type instead, both become proportional to the few protocols a single type implements. Buckets are kept in insertion order and closed witnesses are enumerated before active ones, which reproduces exactly the order the previously reversing consumers produced, so the order in which the solver considers witnesses is unchanged.

Candidate and witness enumeration order is preserved throughout, since that order determines which alternative the backtracking solver commits to first. On the full juniper_crpd module the type reconstruction phase now finishes in well under twenty minutes where it previously ran for hours.

### DIFF
--- a/compiler/lib/src/Acton/Env.hs
+++ b/compiler/lib/src/Acton/Env.hs
@@ -574,16 +574,6 @@ actorSelf env               = case lookupName selfKW env of
                                 Just (HNVar (TCon _ tc)) | isActor env (tcname tc) -> True
                                 _ -> False
 
-actorMethod env n0          = walk [] (activeNames env) (closedNames env)
-  where
-    walk ns ((n,NDef{}):te) rest
-                            = walk (n:ns) te rest
-    walk ns ((n,NVar (TCon _ tc)):te) rest
-      | n == selfKW         = isActor env (tcname tc) && n0 `elem` ns
-    walk ns (_ : te) rest   = walk ns te rest
-    walk ns [] []           = False
-    walk ns [] rest         = walk ns rest []
-
 isDef                       :: EnvF x -> QName -> Bool
 isDef env n                 = case tryQName n env of
                                 Just HNDef{} -> True
--- a/compiler/lib/src/Acton/Env.hs
+++ b/compiler/lib/src/Acton/Env.hs
@@ -53,6 +53,12 @@ data EnvF x                 = EnvF {
                                 closedNames:: TEnv,
                                 hnames     :: HTEnv,
                                 closedHNames:: HTEnv,
+                                sigLocs    :: LocEnv,
+                                closedSigLocs:: LocEnv,
+                                defLocs    :: LocEnv,
+                                closedDefLocs:: LocEnv,
+                                activeStateNames:: [Name],
+                                activeTypeVars:: [(Name, Kind, CCon)],
                                 imports    :: [ModName],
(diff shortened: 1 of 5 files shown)


plajjan:
@nordlander here we go!!!! the first commit on this branch is the same you said was OK yesterday, but now with your feedback, i.e. removed those closed variants that can never happen, so I guess you can mostly skip past that one. Please read the other two commits.

These three are a combination of the good stuff from my prototype branch (out of 29 commits) and together are what makes the big difference in type checking speedup.

nordlander:
Second commit looks perfectly ok (although ”the previously debug only type id graph” is quite some understatement - it was written precisely for the present purpose, just not activated!).

nordlander:
Commit 3 all good as well. One reflection here is that this new approach seems open up for a whole set of further conceptual simplifications (not really any need for NExt anymore, maybe not even the NClass/NProto/NAct representations, nor the current mixing of types and terms int the same kind of environment). But this can be deferred to a future refactoring, once these complexity improvement steps are taken.

plajjan:
@nordlander so you are OK that we merge this? or any reservations? anything we should address before merge?

nordlander:
I’m good to go, my mumblings are more about what will come next.

PR #2935, opened 2026-06-16
Direct calls 2

sydow:
This is a new version of a compiler improvement that replaces calls to protocol instance methods via a witness variable and the instance's method table with a direct call to the method, when this is statically known. This version only does this rewrite for builtin protocols and instances of these for builtin types.

The previous version (PR #2903) introduced a new constructor in the abstract syntax for expressions. Such a side channel for transferring info between passes should be avoided and this version does not change abstract syntax. Instead, in the boxing pass we rewrite the indirect call ```Call (Dot w n) p``` (with some SrcLoc info omitted), where n is a method of some protocol Prot and w a witness that some builtin type typ implements Prot, to ```Call ProtD_typD_n (PosArg w_s p)```  where ```w_s```  is an existing builtin static witness, the name of which is known by convention, and ```ProtD_typD_n``` is the conventional name for the direct method. A problem is that this method only exists at the C level and is not accessible to QuickType. CodeGen recognizes such conventional names and is prepared to bypass qType and compute relevant types.


sydow:
Reviewer approval via email

PR #2943, opened 2026-06-17
Box unboxed-rep values reaching `value` targets at more use-sites

### LINKED ISSUE #2942: Missing box when assigning fixed-size int to value
When a fixed-size int (or I guess an expression whose final representation is an atom) is assigned to a `value` type (attribute), it doesn't get boxed. Accessing the `value` type attribute later attempts to dereference the unboxed value, resulting in a crash.

We do this for example when constructing gdata.Leaf - parse a numeric string into a fixed-size atom, then assign that to the `val: value` attribute. It is valid, right?

Small repro case:

```
PORT: u64 = u64(2001)

class Cell(object):
    v: ?value
    def __init__(self):
        self.v = None

def fill(c: Cell) -> None:
    c.v = PORT # <- no box (toB_u64)

actor main(env):
    c = Cell()
    fill(c)
    # isinstance reads c.v->$class; with the raw int it dereferences 0x7d1 -> crash
    print("isinstance(c.v, atom) =", isinstance(c.v, atom))
    env.exit(0)
```

mzagozen:
This closes #2942. The initial fix was only limited to the crash I found in one of the applications:

```diff
 fixassign env t e
   | Just t' <- unboxedRepType t    = forceUnbox env t' e
+  | Box{} <- e                     = e
+  | Just rt <- exprUnboxedRep env e = Box rt e
   | otherwise                      = e
```

But then my robot continued the search and came up with more edge cases

> Unboxable primitives (u64/int/float) are now carried unboxed; each use-site that needs a boxed `value` must insert the `Box`. That insertion existed for fixarg/fixreturn/boxValueExpr but was missing elsewhere, so a raw int reached a boxed slot and the first virtual access dereferenced it, causing a runtime crash.
> 
> - fixassign: box a raw unboxed-rep RHS into a value/?value target. #2942 
> - Cond: box then/else branches when the join type is not unboxable.
> - IsInstance: box the operand ($ISINSTANCE derefs $class).
> - BinOp And/Or: box both operands ($AND/$OR call __bool__ on them).
> - exprUnboxedRep: treat a processed And/Or as already boxed, else a fallback re-boxes it (toB_u64(B_u64), a compile error).

### DIFF
--- a/compiler/lib/src/Acton/Boxing.hs
+++ b/compiler/lib/src/Acton/Boxing.hs
@@ -242,6 +242,11 @@ literalUnboxedRep _              = Nothing
 exprUnboxedRep env e@Int{}       = literalUnboxedRep e
 exprUnboxedRep env e@Float{}     = literalUnboxedRep e
 exprUnboxedRep env e@DotI{}      = Nothing
+-- After boxing, an And/Or expression has boxed operands and yields a boxed
+-- value (its result IS one of the operands), so it must not be re-boxed by a
+-- use-site fallback, even though its static type is unboxable.
+exprUnboxedRep env (BinOp _ _ op _)
+  | op `elem` [And, Or]           = Nothing
 exprUnboxedRep env c@(Call _ f _ KwdNil)
   | callReturnsMsg env f          = Nothing
   | rawClassConstructor env c f    = unboxedRepType (typeOf env c)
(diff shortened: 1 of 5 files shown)


mzagozen:
@sydow can you please take a look? I also included regression programs that crash on *main*.

sydow:
Good morning Marko,

did you get a review? I thought I submitted one but now I see no signs of it on the PR page at GitHub. I hate git…

Anyhow, the short form is: go ahead and merge!

Björn

PR #2946, opened 2026-06-18
Drop project prefix from snapshot dir paths

mzagozen:
Snapshot files lived under snapshots/{expected,output}/<proj>.<mod>/ — the last user-visible surface still carrying the redundant prefix after the test-output de-prefixing work, and the paths users open and diff.

Bake the bare module name into the generated test harness (Test.module) so the runtime resolves snapshots/expected/<mod>/ directly. The runner keeps its project-qualified identity (.tydb layout, cache keys, result correlation) and de-prefixes only the snapshot path sites. Rename the committed expected snapshots and the snapshot-cache fixture to match.


mzagozen:
OK, this is ready, but it will require a synchronized update across the project repos with snapshot tests. I can merge and do the needful elsewhere between 2 and 4 PM today.

PR #2954, opened 2026-06-19
Repair environ for zig-cc-linked binaries

mzagozen:
acton and lsp-server-acton are linked as -no-pie executables by zig cc (to target an older glibc). Under that link lld emits a copy relocation for glibc's `environ` but does not redirect the `__environ` alias to it, so __libc_start_main populates __environ (getenv/lookupEnv keep working) while the executable's `environ` copy -- which GHC's RTS reads for getEnvironment -- stays NULL. getEnvironment then returns [], so e.g. http-client's proxyEnvironment never sees HTTP(S)_PROXY and dependency fetches ignore the proxy.

Link a small C shim (cbits/environ_fix.c) into both executables: it defines `environ` in the executable (so no copy relocation is generated) and points it at glibc's live __environ in a startup constructor. Linux only.

### DIFF
--- a/compiler/acton/cbits/environ_fix.c
+++ b/compiler/acton/cbits/environ_fix.c
@@ -1,0 +1,30 @@
+/*
+ * environ_fix.c -- repair getEnvironment() on Linux binaries linked by `zig cc`.
+ *
+ * The compiler is linked as a -no-pie executable by zig cc (to target an older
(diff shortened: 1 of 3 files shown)


plajjan:
Included in #2956 - closing.

Holy cow this was tricky, thanks for the hard work @mzagozen !

PR #2968, opened 2026-06-23
Unbox bool

sydow:
This PR makes bool an unboxed type like float and the bounded integer types, i.e. boolean values are unboxed unless they are used as arguments to polymorphic functions or occur as a value in a slot with a broader declared type (e.g. ?bool or value).  Method __bool__ in the C code has result type bool, as has also various isXXX functions in types str, bytes and bytearray and all methods in protocols Eq and Ord.


plajjan:
The rebased & cleaned up branch in #2976 has been merged, which supersedes this PR, thus I'm closing this. We now have unboxed bools!

Issue #2975, opened 2026-06-25
unbox-bool: forward subclass reference mixes bool method-slot ABI


plajjan:
Bool unboxing can emit different C ABIs for the same inherited bool-returning method slot when a subclass is referenced before its class body.

```acton
class Base(object):
    def __init__(self):
        pass

    def flag(self) -> bool:
        raise NotImplementedError()

def make_child() -> Base:
    return Child()

class Child(Base):
    def flag(self) -> bool:
        return True

def through_base(obj: Base) -> bool:
    return obj.flag()

actor main(env: Env):
    if through_base(Child()):
        env.exit(0)
    env.exit(1)
```

Run:

```sh
acton build --color never
out/bin/main

panic: load of value 24, which is not valid for type 'bool'
src/main.act:16:12: return obj.flag()
src/main.act:19:9:  if through_base(Child()):
```

Generated C mismatch:

```c
/* main.h */
bool (*flag) (unbox_bool_method_reproQ_mainQ_Base);
B_bool (*flag) (unbox_bool_method_reproQ_mainQ_Child);

/* main.c */
bool N_1tmp = ((bool (*) ($WORD))((unbox_bool_method_reproQ_mainQ_Base)(obj))->$class->flag)(obj);

unbox_bool_method_reproQ_mainQ_ChildG_methods.flag =
    (B_bool (*) (unbox_bool_method_reproQ_mainQ_Child))unbox_bool_method_reproQ_mainQ_ChildD_flag;
```

Expected: the inherited slot ABI must agree between base and child. Removing only `make_child()` makes this compile to consistent raw `bool` slots.

PR #2976, opened 2026-06-25
Unbox bool

plajjan:


plajjan:
I have squashed and cleaned up @sydow unbox-bool branch. Then on top of that I placed one fix, which addresses #2975.

@sydow please read #2975 and then the last fix on this branch and let me know if you agree with the fix or not.

sydow:
The actual bug in #2975 is due to an oversight that missed one case in function matchTypes in Boxing.hs. Just adding 
````
matchTypes t (TUnboxed _ t')        = matchTypes t t'
````
as line 210 in Boxing.hs makes matchTypes work as intended, which also makes rtypeOf work as intended and thus makes the newly introduced functions methodRType, methodVarType and inheritMethodOwner in CodeGen.hs unnecessary together with the use of these functions in various places.

This commit also includes a second, separate cleanup, which seems legitimate and useful. This is what is summarized as "keep ignored wildcard bindings from forcing pure projections into raw expressions, and preserve side-effecting ignored expressions." This may prevent representation changes only done to assign to an ignored variable. Maybe one should keep this as a commit by itself and then make another, separate commit of the one-line change I suggest above for the first part.

Finally, the comments by chatgpt on the PR page seems to be nonsense. Does it not recognize comments in C code, or?

PR #2981, opened 2026-06-28
Mark loop-carried locals volatile

plajjan:
For-loops are compiled with setjmp/longjmp StopIteration handling, so a local declared before the loop, assigned inside it and read afterwards must be emitted as volatile to survive the longjmp. CodeGen already had this analysis but it was gated on isPUSH, and the Boxing pass wraps the loop's $PUSH() condition in a Box node, so the check never matched and volatile was never emitted. In an optimized build the local was rolled back to its pre-loop value, which made argparse drop explicitly provided sub-command options under --release=fast while debug builds worked fine.

The fix strips the boxing layer before the isPUSH check so the analysis runs again and such locals get the volatile qualifier.

This includes a codegen test that runs the compiler pipeline and asserts the generated C marks the loop-carried local volatile, and an integration test that builds and runs a release binary exercising the pattern.


plajjan:
@nordlander PTAL

plajjan:
@nordlander I'm merging because this blocks other work and my fix unlocks me to continue. Please still check though so we handle this correctly :)

nordlander:
Looks good to me.

PR #2993, opened 2026-06-30
Synthesize __get_attr__ attribute getter

plajjan:
Add a compiler-synthesized reflective getter

    pure def __get_attr__(self, name: str) -> ?value

on every class. The synthesized body is empty (return None) through type checking and hashing, so it references no instance attributes and contributes nothing to the dependency graph. CodeGen ignores that body and emits the real field-reflecting if-chain from the class property set: one `name == "attr"` branch per instance attribute (keyed by the Acton attribute name), boxing scalars and pointer-casting object attributes to value, falling through to None.

This is a language-level reflection primitive. Because the getter pins no attributes, a reachable serializer that reflects over attributes no longer forces every attribute and its subtree to be kept, which lets deferred-back-pass selection prune unused subtrees of large generated data models.

### DIFF
--- a/compiler/acton/test/typeerrors/reserved_get_attr.act
+++ b/compiler/acton/test/typeerrors/reserved_get_attr.act
@@ -1,0 +1,6 @@
+class Foo(object):
+    x: int
+    def __init__(self):
+        self.x = 1
+    def __get_attr__(self, name: str) -> ?value:
+        return self.x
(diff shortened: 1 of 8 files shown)


plajjan:
@nordlander PTAL

plajjan:
Johan said OK in other medium. Merging.

PR #2994, opened 2026-06-30
Register bigint as a serializable type

mzagozen:
bigint had no class id, so it serialized under INT_ID and could not be read back: the int deserializer cannot decode its multi-word blob. Give it a BIGINT_ID so it takes the standard object path, where $step_serialize handles dedup and the header row and __serialize__ writes only the value row.


plajjan:
@mzagozen is this used in acton-yang? I figured we would only use `int` there.. so maybe that is worth looking into and changing separately!? :)

mzagozen:
@plajjan yes, we use `yang.type.Decimal` with *bigint* attributes to represent ranges.

PR #2997, opened 2026-07-01
Implement serialize/deserialize builtins

### LINKED ISSUE #1614: Expose (de)serialization function
We have the ability to serialize everything internally. We should expose functions to serialize and deserialize objects so that we can send them across sockets or write to files. Very handy!

mzagozen:
Add `serialize: (value) -> bytes` and `deserialize: (bytes) -> value` for writing and reading object graphs in memory. Shared and cyclic references are preserved. Closes #1614

serialize() runs without a globmap, so attempts to serialize an actor raise a ValueError.

This is quite a naive implementation that does lots of tiny allocations which could be further optimized in future. The existing $serialize_file and $deserialize_file are untouched - probably need to rewrite them to use capabilities. But nevertheless it is a good starting point for verifying whether we can serialize / deserialize the compiled YANG schema (DNode).

### DIFF
--- a/base/src/__builtin__.act
+++ b/base/src/__builtin__.act
@@ -1027,6 +1027,19 @@ def reversed[A](seq : Sequence[A]) -> Iterator[A]:
 def round [A(Real)] (x : A, n : ?int) -> A:
     return x.__round__(n)
 
+def serialize(obj: value) -> bytes:
+    """Serialize an object graph into a compact binary representation.
+
+    The result can be turned back into an equivalent graph with deserialize().
+    Shared and cyclic references within the graph are preserved.
+    """
+    NotImplemented
+
+def deserialize(data: bytes) -> value:
+    """Reconstruct an object graph previously produced by serialize().
+    """
+    NotImplemented
+
 def sorted [A(Ord)] (iter: Iterable[A]) -> list[A]:
     NotImplemented
 
--- a/base/builtin/serialize.c
+++ b/base/builtin/serialize.c
@@ -88,6 +88,9 @@ void $step_serialize($WORD self, $Serial$state state) {
     if (self) {
         int class_id = $GET_CLASSID((($Serializable)self)->$class);
         if (class_id > ITEM_ID) { // not one of the Acton builtin datatypes, which have hand-crafted serializations
+            if (!state->globmap && issubtype(class_id, ACTOR_ID))
+                // This also catches Msg or Cont because they reference the target actor transitively
+                $RAISE((B_BaseException)$NEW(B_ValueError, to$str("serialize: cannot serialize actors")));
             if (state->globmap) {
                 long key = (long)state->globmap(self);
                 if (key < 0) {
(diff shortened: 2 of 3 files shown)


base/src/__builtin__.act
```diff
@@ -1027,6 +1027,19 @@ def reversed[A](seq : Sequence[A]) -> Iterator[A]:
...
+    The result can be turned back into an equivalent graph with deserialize().
+    Shared and cyclic references within the graph are preserved.
+    """
+    NotImplemented
+
+def deserialize[T](data: bytes) -> T:
+    """Reconstruct an object graph previously produced by serialize().
+    """
+    NotImplemented
+
```
plajjan:
```suggestion
def serialize(obj: ?value) -> bytes:
    """Serialize an object graph into a compact binary representation.

    The result can be turned back into an equivalent graph with deserialize().
    Shared and cyclic references within the graph are preserved.
    """
    NotImplemented

def deserialize(data: bytes) -> ?value:
    """Reconstruct an object graph previously produced by serialize().
    """
    NotImplemented

```

I think this should work on `?value` rather than a `T`. Makes it a tad more sound and honest IMHO.

mzagozen: How so? Then we can't serialize a bare tuple, but an embedded tuple works :)

plajjan: serialize() would work just the same as now.

deserialize() would give back ?value (we could also just do `value` if that is easier), so you would then `isinstance` check it - I think that is just a tad better

plajjan: oh, a bare tuple... hmm

mzagozen: Right, so if we make it `serialize(?value)`, then `serialize(("a", "tuple"))` is a compile error, as well as `serialize(Actor())`, but both may still be passed in some wrapper to avoid that. My question was, how does the non-generic version make it more honest?

plajjan: Because saying this returns T - a general type variable - is a complete lie. A general type T will fit anything and since you implement deserialize in C, the constraints placed on T are just ignored. You ask for deserialize to give you a `str` and you will from a type perspective get a str but it can be something else entirely at run time.

This amounts to disabling type checking and allowing type / run time mismatch, which is basically the basic design behind statically compiled languages.

With a `value` we are honestly saying "deserialize can return anything" (since value is a union of anything) - thus, to use it, you must isinstance check it - then we do the check (a run time check against the actual type of the value)). Thus, this is safe. For our simple case of DRoot, it is just a single isinstance check at the top and then the rest follows the type of DRoot.

Now deserialize is of course still unsafe in the sense that you can modify and write whatever bytes stream you want which will then become objects in memory, but if we assume no one tampers with or tries to maliciously construct the serialized bytes stream, then at least by using `value` no user can mistakenly do unsound things.

mzagozen: Of course 🤦, I will refactor.

PR #3004, opened 2026-07-02
Fix panic when using a tuple as a condition

plajjan:
Using a tuple in a truth-test position, like `if t:`, typechecked and compiled but panicked at runtime with a misaligned-address error: the raw C bool returned by the tuple's __bool__ slot was dereferenced as a boxed B_bool pointer.

The seam was between the Boxing pass and CodeGen. Boxing.rtypeOfFun had no case for tuple-typed receivers, so the callee type of the normalized t.__bool__() call fell back to the plain boxed () -> bool and no Box marker was inserted around the call result. CodeGen.dotCast, however, resolves tuple attributes via the value class and emits the unboxed slot ABI (bool return), so the condition ended up doing ((B_bool)raw_true)->val. The fix resolves tuple receivers via the value class in rtypeOfFun too, mirroring CodeGen.dotCast, so the two passes agree on the ABI and conditions consume the raw bool directly, the same as for list/str/bytes receivers.

Since every truth-test context (if, while, not, ternary, assert, comprehension guards) normalizes through the same __bool__ call, they were all affected for tuples and are all fixed by this change. The other tuple value attributes, __str__ and __repr__, return boxed types so their behavior is unchanged.

Adds a regression test covering truthy non-empty tuples and falsy empty tuples across if, not, while and ternary forms.

### DIFF
--- a/compiler/lib/src/Acton/Boxing.hs
+++ b/compiler/lib/src/Acton/Boxing.hs
@@ -505,6 +505,7 @@ rtypeOfFun env f@(Dot _ (Var _ x) n)
 rtypeOfFun env f@(Dot _ e n)        = case typeOf env e of
                                         TCon _ tc -> rtypeOf env tc n
                                         TVar _ tv -> rtypeOf env (findTVBound env tv) n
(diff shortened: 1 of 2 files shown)


plajjan:
FYI @sydow @nordlander - this looks straight forward enough so I'll merge but perhaps you just want to see and read

PR #3009, opened 2026-07-02
Eq, Ord, Hashable for tuples

plajjan:
Generate Eq extension for tuples of fields that support Eq. Then do the same for Ord & Hashable.


plajjan:
@nordlander PTAL

nordlander:
Looks good overal, but there were two omissions that I've now fixed. I've also removed some documentation text that talked about a workaround that's no longer required. Merging.

PR #3015, opened 2026-07-06
Optimize module global initialization and reduce stack usage

mzagozen:
- Chunk module init to bound stack frame: previously the module global init function was unbounded, which required a bigger startup stack size (main thread) than defaults (8MB on macOS for instance).
- Assign module globals directly to their global: avoid unnecessary local declarations consuming a slot (in Zig debug builds)


mzagozen:
@nordlander / @plajjan PTAL. This chunking addresses two problems when compiling unpruned device data models:
1. Runtime error when attempting to start a program that requires stack size > default
2. Non-linear compilation times for release build on x86 (LTO enabled) for huge init functions

nordlander:
This looks good to me. Never imagined a rewrite like this would be needed, though... 😅

PR #3024, opened 2026-07-13
Fix panic on pure lambda returning tuple

mzagozen:
idtup lacked a catch-all for non-tuple parameter types in the eta-reduction rule, crashing on any pure lambda returnign a tuple.


mzagozen:
@nordlander PTAL

nordlander:
Yes!

PR #3025, opened 2026-07-14
Add ldexp function to math module

ciaran2:


plajjan:
@ciaran2 isn't this more idiomatic as a from_X staticmethod on float directly!?

PR #3029, opened 2026-07-19
Remove the macOS 26 DEVELOPER_DIR override

mzagozen:
macOS 26.4 Command Line Tools broke Zig (ziglang/zig#31658) for a while. We worked around it by forcing DEVELOPER_DIR=/dev/null so Zig fell back to its bundled headers. That issue seems to be resolved on my machine at least with the latest updates.


mzagozen:
@plajjan @nordlander can you test this on your machines? It seems to build projects (sorespo) on mine.

PR #3034, opened 2026-07-21
Add tests of await semantics

plajjan:
Add 12 tests that pin down the observable semantics of await ahead of the planned runtime rework of message/future handling. They cover awaiting a pending result, already-frozen values and exceptions, exception propagation to one and to several waiters, escaped futures awaited by a foreign actor, chained awaits, fan-in, fan-out, interleaving with timers, mailbox ordering around a park, and divergent outcomes from one producer.

Each test carries a docstring describing exactly which RTS code paths it exercises (dispatch cases, waiter registration and flush ordering, freeze and wake loops) and why its outcome is deterministic. The tests are written to fail on missed wakes, duplicate wakes and ordering regressions, using a separate never-parked watchdog actor since a timer in a parked actor can never fire.

### DIFF
--- a/test/core_lang_auto/await_already_done.act
+++ b/test/core_lang_auto/await_already_done.act
@@ -1,0 +1,53 @@
+# The $RWAIT frozen-VALUE fast arm, deterministically, for both awaiter
+# kinds — plus the zero-waiter completion.
+#
+# RTS coverage:
+#  - Zero-waiter freeze: produce() is issued in main's __init__ turn and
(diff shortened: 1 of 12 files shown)


plajjan:
@nordlander I would like for you to read these tests at some point. I think this should exhaustively test the code paths we have through RTS around message handling, so once we start working on Msg split and then further reshaping to avoid cross-actor heap manipulation, this should be a solid foundation. I am particularly interested if you think there are missed cases not covered by these tests.

PR #3037, opened 2026-07-23
Fix witness field loss in imported classes

mzagozen:
A generic type with a constraint on the protocol, like Keyed[K(Eq)], stores a hidden protocol witness as an instance field. When such a class was imported and subclassed in another module, convEnvProtos rebuilt its attributes without the witness field, so the subclass struct came out one slot short and the base class methods read and wrote fields at the wrong offsets, corrupting memory.

Include the witness fields when converting imported classes, like the protocol and extension conversions already do.

Add a regression test that subclasses a generic Eq-constrained class from another module, both concretized and still generic, and round-trips an instance through serialize()/deserialize().


mzagozen:
@nordlander PTAL. I initially discovered this in https://github.com/stratoweave/acton-yang/pull/472, or rather when attempting to use the refactored lists in sorespo.

nordlander:
Yes, this is what it must look like. Good catch!

PR #3041, opened 2026-07-28
Fix comprehensions in branch conditions

plajjan:
The Normalizer rewrites each comprehension into a call to a generated function (compfun), deferring the function itself via shared NormM state. normSuite drains that state after normalizing each statement and emits the pending defs in front of it. But a comprehension in a branch condition is deferred while the statement is still being normalized, and the branch body's nested normSuite drained it first: the def ended up inside the branch body while the call sits in the condition, out of scope. The back pass then failed with NameNotFound (Internal NormPass "compfun" 0) when the Deactorizer looked up the call target. Any comprehension in an if, elif or while condition triggered this, in actors and plain functions alike.

Shield comprehensions pending on entry to a suite from that suite's drain, so they are materialized by the enclosing suite in front of the statement whose condition references them.

### DIFF
--- a/compiler/lib/src/Acton/Normalizer.hs
+++ b/compiler/lib/src/Acton/Normalizer.hs
@@ -102,10 +102,19 @@ normEnv env0                        = setX env0 NormX{ marksX = [], rtypeX = Not
 
 -- Normalize terms ---------------------------------------------------------------------------------------
 
-normSuite env []                    = return []
-normSuite env (s : ss)              = do s' <- norm' env s
+-- Comprehensions deferred while normalizing an enclosing statement's expressions
+-- (e.g. a branch condition) must be materialized before that statement, not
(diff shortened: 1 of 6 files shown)


plajjan:
@nordlander PTAL. PR is built with 2 commits, first introduces a test showing how the compfun is referenced before it is defined. The second commit fixes the compiler and shows the updated snapshot that rearranges so the compfun definition is before it is used.

plajjan:
@nordlander I mean, take a look when you have time. This is not critical now, just async style notification, the current vacation-continuation can run to completion  ;)

nordlander:
[Heavy rainfall behaves quite like *yield await* on vacation plans…!]

This might be Björn’s code originally, but I recall making a very similar mistake in the CPS pre-pass a long time ago. So I can easily verify that the bug is real and the fix is correct!

plajjan:
heh, ok, good. Merging..

PR #3043, opened 2026-07-28
Keep boxed class-method params boxed

plajjan:
A method that overrides a method declared generic in its base class
receives its arguments boxed: calls can arrive through the base class's
method table, where the argument type is not known, so a Deriv(Base[int])
override of Base[T].f takes a B_int, never a raw machine integer.

The Boxing pass assumed that every scalar-typed local holds a raw value.
For these parameters that assumption is wrong. Code generation quietly
compensated at call boundaries, at the cost of box/unbox round-trips, but
the primitive operator paths cancelled the compensation away entirely, so
a body like "return x > 3" compared the pointer instead of the value:
Deriv().f(1) returned True on main.

Boxing now tracks these parameters and leaves their reads boxed; where a
raw value is needed, the unboxing survives to code generation and is
emitted as ->val. Assignments cannot follow the read convention: x += 1
would update the caller's box in place, and a plain rebinding would store
a raw value that later reads unbox again. So a boxed parameter that the
body assigns is renamed aside and copied into an ordinary raw local at
function entry, after which reads and stores follow the standard
raw-local rules and the caller's box is never written.

The first commit adds a small fixture and commits the golden files
produced by the unfixed compiler, so the second commit's diff shows
exactly what the fix changes in the generated code. The churn in the
deact and lines goldens is the removal of the now-unnecessary box/unbox
round-trips at call boundaries; behavior there was already correct.

### DIFF
--- a/compiler/lib/src/Acton/Boxing.hs
+++ b/compiler/lib/src/Acton/Boxing.hs
@@ -12,6 +12,7 @@ import Acton.Subst
 import Pretty
 import Utils
 import Debug.Trace
+import Data.Maybe (isJust)
 import qualified Data.HashSet as HashSet
 import Control.Monad.State.Strict
 import Control.Monad.Except
@@ -30,6 +31,7 @@ data BoxX                          = BoxX { inClassX :: Maybe (TCon, TEnv)
                                           , inActionX :: Bool
                                           , returnTypeX :: Maybe Type
                                           , staticWitnessesX :: M.HashMap Name StaticWitness
+                                          , boxedParamsX :: HashSet.HashSet Name
                                           }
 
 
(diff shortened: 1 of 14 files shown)


plajjan:
@nordlander do you want to take a look at this? it's a correctness fix but also a general speedup since we avoid boxing+unboxing for variables in quite a lot of common scenarios. I think the updated testing snapshots show pretty clearly what is going on

nordlander:
I'd argue that the Boxing pass _is_ correct in its preservation of generic method parameters in an instantiated subclass, and that it also correctly provides boxed arguments when such methods are invoked. To me, the core problem rather seems to be that the Boxing pass fails to use this additional information when method parameters are reassigned, or referenced in an otherwise unboxable binary operation. However, fixing the Boxing pass the way I believe it was intended to work would indeed risk producing code with repeated boxing/unboxing steps, at least if one assumes that an instantiated subclass method will typically do more than simply pass along its parameters to other generic functions.

So the proposed fix is a good move, I think, although primarily for a different reason: it opens up the ability to treat even instantiated method parameters in a consistently unboxed fashion, which ought to be is a big win under most circumstances. It should be noted, though, that the use of multiple disjoint parameter environments in Boxing remains a vulnerabililty despite this fix; the manifest problem has merely been circumvented. One could even consider the added _boxedParam_ environment to be an amplification of this vulnerability, but we'll worry about that at some later time!

[BTW, my idea of avoiding parameter renaming failed because of some recent changes to Types and CodeGen that I haven't caught up on. Ah well...]

PR #3045, opened 2026-07-29
Register no witness for finalized protocols

plajjan:
An extension covers its protocol's whole inheritance ancestry: extending Pt with Ord also makes the extension a witness for Eq, because Ord inherits Eq. The first extension to cover a protocol owns it: the type checker requires it to implement the protocol's methods, and every later extension covering the same protocol is forbidden from implementing them again (the checker calls these methods finalized). So for each protocol and type there is exactly one implementation, held by whichever extension came first.

Witness registration did not follow that rule. An extension registered a witness for every protocol in its ancestry, including the finalized ones it holds no implementation for. Such an entry is an empty shell: resolution that lands on it finds methods the extension was not allowed to write. Within one module the shells were harmless, because registration deduplicates against the local table and the owning extension's earlier entry wins. Across modules there was no deduplication at all, since the local table and the imported interfaces are enumerated separately. Declaring extension points.Pt (Hashable) in a module that imports points, where points already extends Pt with Ord, registered a second Eq witness for Pt even though the extension implements nothing of Eq. The constraint solver requires exactly one candidate per witness lookup, so every == involving Pt stopped compiling, inside the extending module and in every module importing both. In practice, extending an imported class with any protocol that inherits something was unusable.

Nothing about a registration itself distinguishes a shell from a genuine implementation declared in an unrelated module; the distinction exists only at the moment the extension is checked, when we know whether an earlier witness covers the protocol. So the extension record now lists its finalized protocols (bumping the .tydb interface version), and those protocols are skipped everywhere witnesses are registered: the local type table, the extension's own self-witnesses used while checking its body, and the entries importers read from its interface. Every remaining registration is backed by an implementation. A protocol with one implementation resolves to it from any module, in any import order. And nothing is silently legalized: if two independent modules, neither importing the other, both implement the same protocol for the same type, then neither was finalized, both register, and every use remains ambiguous and rejected in either import order, whether they cover the protocol directly or through their ancestry.

The first commit pins the three outcomes as build-failure projects under the unchanged compiler: two rival arrangements that must keep failing, and the single-implementation arrangement that failed for no good reason. The second commit's diff flips only the third into a running test and leaves the rival projects untouched.


plajjan:
@nordlander I don't normally hope for rain in your location... :P  anyhow, please have a look at this, when you have time. Is this the right direction / solution? I didn't see any other way than to store more stuff.

nordlander:
The goal here is to avoid installing witness duplicates, something the existing implementation already attempts to do albeit with limited success after the environmet representation was split into distinct structures for each imported module. But letting each NExt carry info about which of its ancestor protocols that already have implementations in the global witness database must be considered an overkill -- this info is just as readily available when the witness database is extended with NExt offsprings. In essence it amouns to also check for duplicates against the imported witnesses as well as those of the current module, a rather modest fix compared to the current proposal. I've created PR #3061 for that purpose, which carries over the new tests introduced here. Closing in favor of #3061, that is.

PR #3049, opened 2026-07-31
Do witness resolution & forwarding in Converter

plajjan:
This builds on #3043 and #3045 (perhaps easier to review them individually), so it's just the last 2 commits in this PR that is the real change in this PR.

We currently have some code running in CodeGen to fill in unpopulated witness slots, like when a protocol inherits from some other protocol but a particular extension doesn't concretely specify all methods. Our classic example is in builtins where both the Ord and Hashable protocol inherit from Eq. Some instance implement Ord, which define `__eq__` to cover the Eq part, and then the later Hashable protocol should have its `__eq__` resolved to the available method. The first extension to implement the methods for a protocol "owns" it and can be used by later extensions.

Doing this in CodeGen is really rather late. For the work on selective back passes, we want to collect this information about witnesses and we can't do this from CodeGen, so we repeated the effort. This PR does a more proper move of this witness resolution functionality into Converter so it is available in the AST just after the type checking pass. The change itself is structured into 2 commits, the first which adds a snapshot test so we can see in the second commit the effect it has on the AST.

### DIFF
--- a/base/builtin/bigint.c
+++ b/base/builtin/bigint.c
@@ -648,14 +648,6 @@ bool B_OrdD_bigintD___ge__ (B_OrdD_bigint wit, B_bigint a, B_bigint b) {
 
 // B_HashableD_bigint ///////////////////////////////////////////////////////////////////////////////////////////////////////
 
-bool B_HashableD_bigintD___eq__(B_HashableD_bigint wit, B_bigint a, B_bigint b) {
-    return zz_equal(&a->val,&b->val);
-}
-
-bool B_HashableD_bigintD___ne__(B_HashableD_bigint wit, B_bigint a, B_bigint b) {
-    return !zz_equal(&a->val,&b->val);
-}
-
 B_NoneType B_HashableD_bigintD_hash(B_HashableD_bigint wit, B_bigint a, B_hasher h) {
     long sz = a->val.size;
     unsigned long data;
--- a/base/builtin/complex.c
+++ b/base/builtin/complex.c
@@ -142,14 +142,6 @@ bool B_EqD_complexD___ne__ (B_EqD_complex wit, B_complex a, B_complex b) {
 
 // B_HashableD_complex  ////////////////////////////////////////////////////////////////////////////////////////
 
-bool B_HashableD_complexD___eq__(B_HashableD_complex wit, B_complex a, B_complex b) {
-    return creal(a->val) == creal(b->val) && cimag(a->val) == cimag(b->val);
-}
-
-bool B_HashableD_complexD___ne__(B_HashableD_complex wit, B_complex a, B_complex b) {
-    return !B_HashableD_complexD___eq__(wit,a,b);
-}
-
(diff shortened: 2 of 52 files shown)


plajjan:
@nordlander I felt I had a bit of an epiphany during #3045, an insight upon which this branch is then based. simple and elegant 😃  the branch is actually a net-negative in terms of lines of code, removing more code than it adds (if we do not count the new test cases). It is also quite possible that I'm missing something central on how this needs to work, rendering the PR useless 😜

nordlander:
This change is really good, I like it a lot! Very appropriate to let the type-checker itself sort out what witnesses to use in the forwarding methods.

I'm just curious about the potential clashes that would yield the "Conflicting inherited signatures" error. Has an example of where this happens occurred? My intuition is that the mro algorithm would have trapped any such inconsistency already, but I could be wrong.

plajjan:
@nordlander I am not aware of any "real" occurrences in real life - only this new test case introduced in this PR test/core_lang_auto/witness_forward_diamond_clash__bf.act which shows the problem and MRO does not seem to throw any error there. Rather than resolving the wrong thing, we explicitly throw an error. Makes sense?

nordlander:
Yes, makes sense. Good to have an example, I’ll check it up. But we should be good to merge.

plajjan:
@nordlander "should be" sounds like you still might have some last doubts ;) If you think we should merge, please press the merge button :)

nordlander:
Ok, I did some more digging on the MRO topic and decided to make a few additions/clarifications to the clash test while we're at it. Here's what I found:
- The MRO algorithm does indeed detect inconsistent ancestries (when a protocol needs to appear both before and after some other protocol in the resolved order) and inconsistent instantiations (when a parametrized ancestor is reached via multiple paths that apply different type arguments). I've added tests that illustrate this.
- The algorithm hasn't so far cared about detecting inherited signature clashes, though, for a reason I bought into which is entirely valid in untyped Python: the resolved ancestry order uniquely selects a single method implementation whenever there's a clash, so it's "safe" to just ignore the alternatives. The fact that this puts an unreasonable burden on the programmer (methods may take on a completely new signature when overridden) should of course have reminded me that Python's notion of "safe" doesn't automatically carry over to a typed setting, but alas, it didn't.
- Still, the example in `test/core_lang_auto/witness_forward_diamond_clash__bf.act` isn't really among the dangerous ones, which can be verified by simply turning off the clash test (and actually invoking `f` in the test). Or at least this _would_ be verified if it weren't for an entirely separate bug I discovered: the environments computed by `parentTEnv` are in reversed order compared to all other TEnvs. That bug is now fixed, making the previous statement true and hopefully avoiding future confusion.
- Nevertheless, a really dangerous example could be constructed by adding a method `g` to `Beta`, with a default implementation that calls `f` (assuming a `str` signature). If that method is ever invoked on a `Thing` all hell would break loose. And I see no way we can prevent this in a systematic manner while still allowing the "benign" clashes described above. So my conclusion is: the proposed clash test is an essential extension of the sanity checks done by the MRO algorithm, but it should be generalized to cover _all_ inherited signatures, not just those determined final. This is now implemented.

Merging.

PR #3055, opened 2026-08-10
Fix forced unwrap chained into tuple selection

plajjan:
Forced unwrapping followed by a tuple field selection, like x!.key on a value of type ?(key: bytes, value: bytes), crashed the compiler with a non-exhaustive pattern error in QuickType during the back passes. The optional chain elaboration rewrites the chain into a conditional whose else branch calls $raiseValueError, and that branch was instantiated at the type of the unwrapped receiver (the tuple) while the then branch carries the type of the selection result (the field type). QuickType recomputes a conditional's type as the least upper bound of its branch types. Two class types always share a common ancestor, so class receivers never exposed the mismatch, but a tuple type and its field type have no upper bound at all and the computation had no answer. Since the raise branch never produces a value, its claimed type is free to choose: it is now instantiated at the type of the whole chain, making both branches agree for any receiver.

The splitter that divides an optional chain at its ?/! mark also lacked cases for some trailers the parser accepts inside a chain: positional tuple selection (x!.0, and x?.0 alike) crashed in split, as did the row-rest selectors x!.~key and x!.~0. The missing cases are added, so positional selection now compiles and runs, and the rest selectors produce their regular "not yet supported" error instead of crashing the compiler.

A new test, test/core_lang_auto/optchain8.act, covers ! and ? chained into named and positional selections on optional tuples, plain unwrap, and the ValueError raised when unwrapping None.

### DIFF
--- a/compiler/lib/src/Acton/Types.hs
+++ b/compiler/lib/src/Acton/Types.hs
@@ -2213,7 +2213,7 @@ instance Infer Expr where
                                             let env1 = define [(x,NVar te)] env
                                             (cs2,t,e2') <- infer env1 e2
                                             y <- newTmp
-                                            (cs3,t',e3) <- alt t te b
+                                            (cs3,t',e3) <- alt t b
                                             return (cs1++cs2++cs3,
                                                      t', eLet [sAssign (pVar y (tOpt te)) e1']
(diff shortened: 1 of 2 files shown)


nordlander:
All good, merging!

PR #3060, opened 2026-08-14
Use ThinLTO for release builds

mzagozen:
ThinLTO keeps cross-module optimization, but partitions the optimizer and code generator per module, so the final link scales across cores rather than merging the whole program into a single LLVM module and processing it in one thread.

On sorespo this reduces the time spent linking a release build from ~150s to ~50s and also peak memory usage by about 40% on x86.


plajjan:
LGTM

Issue #3063, opened 2026-08-17
Handle non-generic class extensions


nordlander:
### Acton Version

0.29.1.20260817.9.18.51

### Steps to Reproduce and Observed Behavior

This fails to find the `int(Eq)` extension when comparing the `item` attributes:
```
class Apa[X](object):
    item: X
    def __init__(self, item: X):
        self.item = item

extension Apa[int] (Eq):
    def __eq__(a, b) -> bool:
        return a.item == b.item
```

### Expected Behavior

Should compile just fine

PR #3065, opened 2026-08-17
Ignore type params when checking prior existence

### LINKED ISSUE #3062: Witness dedup to ignore type variable names
This does not compile:
```
class Apa[X](object):
    item: X
    def __init__(self, item: X):
        self.item = item

extension Apa[A(Eq)] (Eq):
    def __eq__(a, b) -> bool:
        return a.item == b.item

extension Apa[B(Eq)] (Hashable):
    def hash(self, h):
        pass

def eq(x: Apa[int], y: Apa[int]) -> bool:
    return x == y
```

change name of type variable B to A in second Apa and it works:

```
class Apa[X](object):
    item: X
    def __init__(self, item: X):
        self.item = item

extension Apa[A(Eq)] (Eq):
    def __eq__(a, b) -> bool:
        return a.item == b.item

extension Apa[A(Eq)] (Hashable):
    def hash(self, h):
        pass

def eq(x: Apa[int], y: Apa[int]) -> bool:
    return x == y
```

I think dedup should not compare so equal. schematic?

nordlander:
Resolve #3062

### DIFF
--- a/compiler/lib/src/Acton/Env.hs
+++ b/compiler/lib/src/Acton/Env.hs
@@ -280,7 +280,6 @@ extWitnesses mkQ skip exts  = fst (foldl' add ([], Map.empty) wits)
           | otherwise       = (w : ws, Map.insertWith (++) k [w] seen)
           where k           = (tcname (proto w), witnessTypeQName w)
                 bucket      = Map.findWithDefault [] k seen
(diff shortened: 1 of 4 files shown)


nordlander:
See #3063 for an example of the limitations of this fix.

PR #3070, opened 2026-08-18
Build tuples in a single allocation

mzagozen:
Do not re-allocate memory for the components field in B_tupleD___init__. Instead make the components field (tail of struct) a flexible array member. This halves the memory usage for constructing tuples.


plajjan:
LGTM :)

I have been trying to think whether there is a potential issue about the box of boxed objects being of variable size. Should be fine as-is. It is conceivable we could put a box on the stack and perhaps that would be harder, but I don't think that is very appealing anyway.

@nordlander you saw this, right?

nordlander:
Yup, this is ok with me. It might even have been my original intention!😉

PR #3071, opened 2026-08-19
Scan __init__ try blocks in runtime order

plajjan:
scanSelfAssigns decides which self attributes a constructor is guaranteed to have assigned before self can escape. For a try statement it scanned the finally block first and counted its assignments as already made while scanning the try body, even though finally runs last. A constructor that reads an attribute inside try which is only assigned in finally therefore compiled, and the read saw an uninitialized field at runtime.

The scan now follows execution order: the try body, then each handler and the else block as alternative continuations, then finally. The example is now rejected with "Attribute 'first' is not initialized", and becomes the first snapshot test for that check.

### DIFF
--- a/compiler/acton/test/typeerrors/init_try_finally_order.act
+++ b/compiler/acton/test/typeerrors/init_try_finally_order.act
@@ -1,0 +1,8 @@
+class Holder(object):
+    first: int
+    second: int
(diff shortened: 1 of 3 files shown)


plajjan:
Closing for local review before opening; branch kept.

PR #3073, opened 2026-08-24
Optimize type checking by caching imports and attribute lookups

mzagozen:
A recent refactor in https://github.com/stratoweave/acton-yang/pull/472 resulted in a significant increase in times to type check a module with adata user code (seconds to minutes). These proposed caching changes bring the times back down to seconds. The branch `typecheck-env-caching-repro` includes an anonymized repro project.


nordlander:
I'm all for this -- merging!

PR #3074, opened 2026-08-25
Fix set lookup for hashes with the high bit set

### LINKED ISSUE #2552: set.c incorrectly uses has as long instead of uint64_t
This year [Advent of Code](https://adventofcode.com/2025/day/8) again brings interesting puzzles and obscure edge cases in Acton 🎅 

Membership checks using certain hashes on an emptied set cause `B_set_lookkey` to enter an infinite loop. I suspect an overflow error because the function treats the hash as a `long`, while our hashing function produces a `uint64_t`. I’m unsure how to fix it: we need to change the types of values derived from `hash`, like `perturb`, `mask`, `i` in set.c and set.h functions, but what about `finger`, `numelements`, and `nxt`? @sydow, please help.

To reproduce the condition, check for existence of 173418 after emptying the set in [set_test.act](https://github.com/actonlang/acton/blob/main/test/builtins_auto/set_test.act):

```diff
diff --git a/test/builtins_auto/set_test.act b/test/builtins_auto/set_test.act
index b33e8985..8fe22d86 100644
--- a/test/builtins_auto/set_test.act
+++ b/test/builtins_auto/set_test.act
@@ -13,6 +13,9 @@ def test_set_pop():
         pass
     else:
         raise ValueError("pop() did not raise an exception when called on an empty set")
+    if 173418 not in s:
+        # hash = 14252248035062160740
+        print("This one in particular")

 def test_set_update_int():
     s1 = {1, 2, 3}
```

thc1006:
`B_set_lookkey`, `B_set_add_entry`, and `B_set_insert_clean` declare the probe `perturb` as a signed `long`. The hash function produces a `uint64_t`, so a hash with the high bit set becomes a negative `long`, and `perturb >>= PERTURB_SHIFT` is then an arithmetic shift that keeps the sign bit and never decays to zero. The probe `i = (i * 5 + 1 + perturb) & mask` never gets its full permutation and can loop forever.

Reproducer from #2552: looking up `173418` (whose hash has the high bit set) on a set emptied by `pop()` spins in `B_set_lookkey` and never returns; even `SIGTERM` does not stop the tight loop.

## Fix

Make `perturb` a `uint64_t` in the three probe loops so the shift is logical and the sequence decays to zero. That matches the hash function's own type; `dict.c` already guards its probe the same way, with an unsigned perturb (`unsigned long`). Nothing else changes: `numelements`, `fill`, `mask`, and `finger` stay `long` (they never go negative, per @sydow's note on the issue), and the dummy sentinel (`hash == -1`) is untouched, so tombstone handling is identical.

## Tests

`set_test.act` gains regression tests for the reported case and other table histories that drive the probe through tombstones and resizes:

- the reported lookup on a set emptied by `pop()`
- delete then reinsert the same value
- a tombstone-heavy table (add many, discard all)
- a resized table emptied again
- many mixed hashes, checking each membership against what was kept

I checked each against the runtime both ways: with the current `set.c` the test binary has to be `SIGKILL`ed, and with the fix it exits cleanly.

Closes #2552

### DIFF
--- a/base/builtin/set.c
+++ b/base/builtin/set.c
@@ -26,7 +26,9 @@ static $WORD _dummy;
 
 static void B_set_insert_clean(B_setentry *table, long mask, $WORD *key, long hash) {
     B_setentry *entry;
-    long perturb = hash;
+    // perturb is unsigned so >>= is logical: a signed high-bit hash would
+    // sign-extend, never reach 0, and let the probe loop forever (#2552).
+    uint64_t perturb = hash;
     long i = hash & mask;
     long j;
 
@@ -90,7 +92,7 @@ static int B_set_table_resize(B_set so, int minsize) {
 
 static B_setentry *B_set_lookkey(B_set set, B_Hashable hashwit, $WORD key, long hash) {
     B_setentry *entry;
-    long perturb;
+    uint64_t perturb;  // unsigned (#2552)
     long mask = set->mask;
     long i = hash & mask;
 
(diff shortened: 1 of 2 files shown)


plajjan:
Thank you very much for this. It is conceptually correct but I felt I wanted to go a few steps further, which I did in #3075. This is thus superseded.

PR #3075, opened 2026-08-30
Fix set handling of high-bit hashes

plajjan:
Set probing stores hashes produced as uint64_t in signed long values. Hashes with the high bit set can therefore sign-extend during perturbation and make lookups loop forever, while long also has different widths across supported targets.

This uses uint64_t consistently for set hashes, counters, masks, and probing. Tombstones are represented by their key state instead of reserving a hash value, leaving the full hash domain available. Set resizing now raises MemoryError when its allocation cannot be represented or completed, and the regression coverage exercises high-bit hashes across deletion, reinsertion, resizing, and serialization.

### DIFF
--- a/base/builtin/set.c
+++ b/base/builtin/set.c
@@ -13,10 +13,8 @@
  */
 
 
-#define DISCARD_NOTFOUND 0
-#define DISCARD_FOUND 1
 #define PERTURB_SHIFT 5
-#define MIN_SIZE 8
+#define MIN_SIZE UINT64_C(8)
 
 static $WORD _dummy;
 #define dummy (&_dummy)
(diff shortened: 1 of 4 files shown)


plajjan:
@sydow I would like for you to read this. I think it is correct - I can certainly say that it addresses an issue I found with stratoweave where transaction hang completely which I can trace back to a particular (random) hash which has high bits set. This fix removes that issue. I'm not sure if changing everything to uint64_t is overdoing it, like set size of 32 bit feels quite enough for example... but anyway, I will merge this now because it improves things. Please read @sydow and shape as you feel is appropriate!

sydow:
This looks fine. One more step towards consistent use of the intXX_t famiy of integer types instead of int, long, etc. We (or I) should continue to do this throughout builtin.

Also a good thing to remove traces of signed hash values, which is conceptually unnatural but goes back to our first choice to follow  Python and use the identity function as hash function  for int's. 

The hash perturbation mechanism is identical for sets and dicts but duplicated in (not identical) code in dict.c and set.c. An improvement would be to move these types and algorithms to hasher.c/h (or possibly separate files perturb.c/h). I could volunteer to do this.

plajjan:
@sydow ah yes, I did notice that dict probably has the same issue as fixed here. Please feel free to fix more things in this code :)

PR #3083, opened 2026-09-03
Effectively treat pure and mut as interchangeable (for now)

nordlander:
Also allow pure callouts in action callbacks (for now).

The first wart will be removed once we have good immutable data structures and a method of safely inferring/allowing local mutations inside otherwise pure functions. The second one goes away the day effects can be redefined as sets of effect tokens, with `pure` standing for the empty set.

Relaxing the effect hierarchy like we do now opens up for less contrieved application code, which currently is more important than fully "correct" effect inference.


nordlander:
Let's just do the first half, closing this.

PR #3098, opened 2026-09-08
Fix type arguments in protocol inheritance

plajjan:
The compiler can crash when a protocol inherits another protocol with a fixed type argument. This fixes the lookup so each generated base class gets the type arguments it expects, and adds a small example that reproduces the crash.


plajjan:
@nordlander PTAL - found issue while prototyping the mutable vs immutable protocol split. First commit in this PR contains minimal reproduction scenario and second commit is a potential fix

nordlander:
Yes, this looks right. But the Converter code is way too complex as it is. We need to switch to the C++ multiple inheritance object layout before we get completely lost!

PR #3103, opened 2026-09-10
Implement Hashable for set

plajjan:
Allow sets to be dictionary keys and elements of other sets. Hash the element count and the sum of cached element hashes so equal sets hash equally regardless of insertion order or table history.

set should ideally not be mutable, but until we have proper mutable / immutable split, this will have to do.

### DIFF
--- a/base/builtin/set.c
+++ b/base/builtin/set.c
@@ -554,6 +554,21 @@ bool B_OrdD_SetD_setD___le__ (B_OrdD_SetD_set wit, B_set set, B_set other) {
     return  B_OrdD_SetD_setD___ge__(wit, other, set);
 }
 
+// B_Hashable
+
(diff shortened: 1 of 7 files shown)


plajjan:
@nordlander @sydow not sure if you have input on the particular technique for hashing this... but this seems to work :)

sydow:
This seems to be a good hashing strategy. But I expect that we will not have the full mutable / immutable split for quite some time. So if you expect to need to use "morally immutable" sets as dict keys and/or set members, I wonder whether we should not have a temporary solution with some version of frozen sets, which would be trivial/simple to implement. 

One might even consider putting in a bit more work and rename protocol Set to ImmutableSet, move all methods except isdisjoint to a subprotocol Set (perhaps taking the opportunity to rename isdisjoint to aredisjoint).

PR #3120, opened 2026-09-12
Allow list equality for Eq-only elements

plajjan:
Allow list equality and inequality when elements implement only Eq. Ordering comparisons still require Ord.

Add a separate Eq extension and let Ord equality forward to it, sharing one runtime equality implementation. Add focused builtin coverage in a preceding commit; existing list tests used ordered elements and missed this case.


plajjan:
@nordlander is this correct? or is the compiler supposed to infer that Eq on list can be implemented for anything that implements Eq simply because Ord relates to Eq.

Regardless, I actually prefer to write the extensions explicitly, like I think it looks nicer in code when we first implement Eq and then Ord and Hashable rather than letting the first protocol that inherits Eq also implement `__eq__` - but I guess we should do this more uniformly if we agree that this style is nicer ;)

nordlander:
This is correct. Had the extensions not been marked NotImplemented the compiler would also have checked that the Ord and Hashable extensions don’t attempt to override `__eq__` and `__neq__`. But the NotImplemented programmer must, as always, ensure such properties hold by construction in the C code.

PR #3122, opened 2026-09-13
Add generic freezing for sets

plajjan:
Converting a set with iset() copies its storage. Add a global freeze() function and a Freeze protocol so sets can become immutable in place by changing their class pointer.

Callers must stop using the original set and its aliases. The compiler does not yet enforce this ownership requirement.


nordlander:
I maintain that the logic here is upside down. A function that takes a set and returns an iset is easy to define and explain, and so is the usage pattern that lets the compiler avoid the copying. Putting the short-cut in the hands of the programmer and then asking the compiler to validate its correct usage is quite out of character with the rest of the language, like offering an explicit (but checked) `tailrecurse` primitive in order to facilitate tail-recursion. If the compiler is adept enough to spot the valid contexts of such a short-cut, it is of course just as able to insert the short-cut itself. 

I’d rather view this function as an interim solution until the compiler becomes sufficiently equipped.

PR #3135, opened 2026-09-16
Fix iset string representation

mzagozen:
Print "iset({...})" instead of the unparseable "i{...}".


mzagozen:
In #3109 @sydow initially noted how `i{...}` doesn't parse, but then I think @plajjan's robot just rolled with it and even added tests :D

PR #3137, opened 2026-09-17
Nonthrowing iterator

sydow:
This PR changes the result type of method __next__ in Iterator[A] from A to maybe[A], were class maybe is defined in __builtin__.act. Iteration termination is handled by ordinary control flow, not with exceptions.

All Iterators in builtin are rewritten; Iterators in user code will have to be rewritten.


sydow:
The bug ithat caused perf-vs-main to fail is fixed but I haven't pushed it yet (and I can swear that this was not revealed when I (several times) did 'make test' on my local machine???)

Many/most of the app failures are due to one old iterator termination treatment in acton-yang/src/type.act, in function isdisjoint for ranges (line 264). I could fix it, but I would like to understand when this is supposed to return true; to me it looks a bit strange.

sydow:
There is one more error reported from several of the apps; it must be a record-holder in lack of info about where the error occurs:

[error]: Constraint violation
     ╭──▶ @0:0-0:0
     │
   0 │ <no line>
     • 
     • ╰╸ __builtin__.value must be a subclass of __builtin__.str
─────╯

In which file? And where?

mzagozen:
> Many/most of the app failures are due to one old iterator termination treatment in acton-yang/src/type.act, in function isdisjoint for ranges (line 264). I could fix it, but I would like to understand when this is supposed to return true; to me it looks a bit strange.

@sydow I can take a look at that function. It is not used right now, but I think it should be to validate the range restrictions in YANG.

sydow:
Thank you Marko. But note that we must synchronise changes; I don’t know how… I expect to talk about this with Johan and Kristian this afternoon.

mzagozen:
So it turned out that the `is_disjoint` function is not necessary because we then validate the ranges (disjoint, non-inverted) when combining ranges in `RangesBuilder` anyway. The function is removed in https://github.com/stratoweave/acton-yang/pull/509

mzagozen:
@sydow I updated acton-yang last week so that it is forward compatible with this. If you are happy with the changes here, I will rewrite the iterators in the remaining repositories and then merge this.

sydow:
@mzagozen Thanks! I also updated acton-yang, including the iterators so that on my machine it ran with all tests passing. I have pushed it to the acton-yang repo as branch nonthrowing-iterator. I have done the same with stratoweave and was just going to write to Kristian to ask for help in doing a synchronized update (I have only superficially checked the other apps bu several of them fail because acton-yang fails. 

Please check what I did in acton-yang and if necessarychange it to your liking.

sydow:
More detailed explanation of how this PR affect programmers.

The __next__ method in class __Iterator__[A] now has type ()->maybe[A], where class maybe is defined in __builtin__.act. The generated C code __for__ next does *not* allocate memory on the heap, but returns a boolean (a value is returned or loop termination); in case of a value, this is passed as an out parameter in C. 

For loops over an iterator will generate proper code; to handle calls to __next__, code should have the shape 

```
iter = expression returning an iterator
x = iter.__next__()        # or x = next(iter)
if isinstance(x,just):
   use the value as x.val
```

mzagozen:
I aligned the rest of the repos in the ecosystem, the last commit here is changing the branch for sorespo from the default *stable* to *main* (#3151). 

I didn't measure the performance impact of this change in detail, but just seeing `make test-yang-compile` in acton-yang show a 2-2.5x reduction in time to compile (YANG -> DNode) the full set of device models, for example 35s to 15s for Cisco IOS XR 25.3.1, is very promising. Excellent @sydow!

PR #3141, opened 2026-09-18
Fix scalar value hashing

plajjan:
Hash i8, i16, i32 and float by value so separately allocated equal values work as set elements and dictionary keys. Normalize positive and negative floating zero to the same hash input.

Hash bool and narrow unsigned values without reading object padding.


plajjan:
@sydow I've merged this now because I'm pretty certain it is correct but please have a look, just we're all aligned :)

sydow:
It certainly looks correct. It is really embarrassing that such obvious errors have not been caught in the test suite!

